PR #1, opened 2019-12-05
Improve Context interface

ionut-arm:
This commit improves the Context structure interface by making all
parameters going into the ESYS calls visible to clients of Context.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -50,8 +50,10 @@ use std::ptr::{null, null_mut};
 pub use tss2_esys::*;
 pub use utils::{AsymSchemeUnion, Signature, TpmaSession};
 
-pub type SessionTrio = (ESYS_TR, ESYS_TR, ESYS_TR);
-pub const NO_SESSIONS: SessionTrio = (ESYS_TR_NONE, ESYS_TR_NONE, ESYS_TR_NONE);
+pub type Sessions = (ESYS_TR, ESYS_TR);
+pub const NO_SESSIONS: Sessions = (ESYS_TR_NONE, ESYS_TR_NONE);
+pub const NO_NON_AUTH_SESSIONS: (ESYS_TR, ESYS_TR, ESYS_TR) =
+    (ESYS_TR_NONE, ESYS_TR_NONE, ESYS_TR_NONE);
 
 // Possible TCTI to use with the ESYS API.
 // TODO: add to each variant a structure for its configuration. Currently using the default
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -132,23 +142,47 @@ impl Context {
...
+        non_auth_sessions: (ESYS_TR, ESYS_TR, ESYS_TR),
+        tpm_key: ESYS_TR,
+        bind: ESYS_TR,
+        nonce: &[u8],
+        session_type: TPM2_SE,
+        symmetric: TPMT_SYM_DEF,
+        auth_hash: TPMI_ALG_HASH,
+    ) -> Result<ESYS_TR> {
+        if nonce.len() > 64 {
+            return Err(Tss2ResponseCode::new(TPM2_RC_SIZE));
```
hug-dev: Should we define our own errors for this layer and abstraction layers above as the `Tss2ResponseCode` should IMO only come from the TPM commands sent from the TPM?

hug-dev: (As in, only used when returning error codes from ESYS functions)

src/lib.rs
```diff
@@ -167,47 +201,85 @@ impl Context {
...
                 .try_into()
                 .unwrap(),
             sensitive: TPMS_SENSITIVE_CREATE {
                 userAuth: TPM2B_DIGEST {
                     size: auth_value.len().try_into().unwrap(),
-                    buffer,
+                    buffer: auth_value_buffer,
+                },
+                data: TPM2B_SENSITIVE_DATA {
+                    size: initial_data.len().try_into().unwrap(),
```
hug-dev: I guess this can never panic as `initial_data.len()` will never be bigger than 256?

src/lib.rs
```diff
@@ -492,15 +601,15 @@ impl Context {
...
         let mut buffer = null_mut();
         let ret = unsafe {
             Esys_GetRandom(
                 self.mut_context(),
+                self.session,
                 sessions.0,
                 sessions.1,
-                sessions.2,
-                num_bytes,
+                num_bytes.try_into().unwrap(),
```
hug-dev: Maybe a check would be nice that `num_bytes` is less than the maximum unsigned value on 16 bits to avoid possible panicking or just deal with the `try_into()` error as you wish.

ionut-arm: I was going to raise an issue to look into all uses of `unwrap` and that sort of thing - just getting things to work for now

src/lib.rs
ionut-arm: yes, should be alright

src/lib.rs
ionut-arm: I thought about that, but you'd need to occupy a block of response codes from the TSS2 space - again, I think I'll raise an issue to look into it and how to document it.

hug-dev: Yes we should raise an issue about that and discuss it there!

src/lib.rs
hug-dev: Cool, as long as it is noted somewhere and we do not forget to fix it.

ionut-arm:
> If so, maybe it is not strictly needed to add more abstraction to this layer (like using struct for parameters) but maybe directly write higher level layer?

I had to suppress some clippy warnings about too many parameters, I think for those methods I'll factor out the params, but the rest I'll leave as is unless there's a compelling point for it.

PR #4, opened 2019-12-09
Add transient object context

ionut-arm:
This commit adds the TransientObjectContext structure which allows
clients to utilise keys in a transient way (i.e. only dealing with
saved contexts, not with handles).

Some resources are also moved around in the crate.

### DIFF
--- a/src/abstraction/transitory.rs
+++ b/src/abstraction/transitory.rs
@@ -1,0 +1,234 @@
+// Copyright (c) 2019, Arm Limited, All Rights Reserved
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//          http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::constants::*;
+use crate::response_code::{Result, Tss2ResponseCode};
+use crate::tss2_esys::*;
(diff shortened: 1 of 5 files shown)


src/abstraction/transitory.rs
```diff
@@ -0,0 +1,218 @@
...
+        let key_handle = self
+            .context
+            .load(NO_SESSIONS, self.root_key_handle, key_priv, key_pub)?;
+
+        let key_context = self.context.context_save(key_handle)?;
+        self.context.flush_context(key_handle)?;
+        Ok((key_context, key_auth))
+    }
+
+    pub fn load_external_rsa_public_key(&mut self, public_key: &[u8]) -> Result<TpmsContext> {
```
hug-dev: Is there a way to set an `auth_value` for an imported public key? Or is it fine not to have one?

ionut-arm: Unfortunately, the only way to set an auth value is to provide a sensitive part to the object, which means it can no longer be loaded into the Owner hierarchy - only the NULL hierarchy allows uploading objects with both public and private sections. 

Since we want to keep the keys persistent, we have to stick to the OH, but that's probably ok - we have mutual authentication for the commands on the key, it's just that someone who gets to know the key handle in the short timeframe when it's in the TPM can use it to verify signatures.

src/abstraction/transitory.rs
```diff
@@ -0,0 +1,218 @@
...
+            vec![]
+        };
+        if !owner_hierarchy_auth.is_empty() {
+            context.set_handle_auth(ESYS_TR_RH_OWNER, owner_hierarchy_auth)?;
+        }
+
+        let root_key_handle = context.create_primary_key(
+            NO_SESSIONS,
+            ESYS_TR_RH_OWNER,
+            &get_rsa_public(true, true, false, root_key_size.try_into().unwrap()),
```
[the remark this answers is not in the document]
ionut-arm: Shall raise an issue to look into improving method signatures all around.

src/abstraction/transitory.rs
hug-dev: Ok! Thanks for the explanation.

PR #5, opened 2019-12-09
Add a CI job for tests with TPM simulation server

hug-dev:
Adds a Docker container that contains the stack and tools needed to run
the TPM simulation server and build our crate.

### DIFF
--- a/tests/all.sh
+++ b/tests/all.sh
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+
+# ------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


tests/all.sh
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env bash
```
ionut-arm: [`set -e`](https://ss64.com/bash/set.html)?

hug-dev: Will add!

PR #11, opened 2019-12-11
Fix various issues

ionut-arm:
This commit fixes the following issues:
* `type X = Y` is typically considered an anti-pattern in Rust -
removing our usage of it
* remove authenatication values from reading public parts of keys
in transient context
* store all 3 sessions required by most commands straight on the
context for brevity

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ Cargo.lock
 *.swp
 *DS_Store
 *.patch
+*NVChip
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -4,3 +4,4 @@ Cargo.lock
 *.swp
 *DS_Store
 *.patch
+*NVChip
```
hug-dev: I got this too! I think it is created somehow when running the Docker image

ionut-arm: No, that's the persistent part of the `tpm_server`, so if you have problems related to DA locking, you have to remove it. Might add a line in `all.sh` to remove this file when finished.

PR #15, opened 2019-12-13
Add wrapper crate-specific errors

### LINKED ISSUE #3: Implement error handling for our wrapping layer
Currently methods in the crate that need to report an error do so by using TSS2 response codes. We should probably have our own means of flagging errors to reduce the confusion.

ionut-arm:
This commit adds a new type of error specific to the wrapper crate.
All errors that are detected within the logic of this crate are
returned as this new type.

This PR fixes #3

### DIFF
--- a/src/response_code.rs
+++ b/src/response_code.rs
@@ -15,7 +15,7 @@
 use crate::tss2_esys::TSS2_RC;
 use bitfield::bitfield;
 
-pub type Result<T> = std::result::Result<T, Tss2ResponseCode>;
+pub type Result<T> = std::result::Result<T, Error>;
 
 bitfield! {
     pub struct ResponseCode(TSS2_RC);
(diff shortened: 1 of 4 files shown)


src/response_code.rs
```diff
@@ -49,10 +49,11 @@ pub enum Tss2ResponseCode {
     Success,
     FormatZero(FormatZeroResponseCode),
     FormatOne(FormatOneResponseCode),
+    WrapperError(Tss2ResponseCodeKind),
```
hug-dev: As this enum list all of the TSS 2.0 specs defined response codes, would it maybe be better to contain a `Tss2ResponseCode` variant inside a bigger enumeration?
It could even be in another file, named `error.rs` which contains an `Error` enum:
```rust
pub enum Error {
    LocalError(LocalErrorKind),
    Tss2Error(Tss2ResponseCode),
};
```
so that our functions (or only the most high-level ones) could return that?

I am just afraid that we are mixing something that is defined by a spec and our own error code in the same module.

ionut-arm: Done!

PR #18, opened 2019-12-18
Improve usage of unsafe blocks

### LINKED ISSUE #2: Investigate use of `unsafe` and panicking

ionut-arm:
This commit improves our handling of `unsafe` blocks. It moves the
`unsafe` label to methods that could lead to undefined behaviour
instead of hiding the lack of safety inside the method.

All remaining `unsafe` blocks should be safe to a sane level - we control the inputs to the function calls within and can guarantee, based on trust (usually in the TSS stack below) that the calls are safe. 
Thread safety is offered by all `Context` struct methods requiring mutable references to `self`. 

Fixes #2

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -12,7 +12,24 @@
 // WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+
 #![allow(dead_code)]
+///! # Notes on code safety:
+///! * the `unsafe` keyword is used to denote methods that could panic, crash or cause undefined
+///! behaviour. Whenever this is the case, the properties that need to be checked against
+///! parameters before passing them in will be stated in the documentation of the method.
+///! * `unsafe` blocks within this crate need to be documented through code comments if they
+///! are not covered by the points of trust described here.
+///! * the TSS2.0 library that this crate links to is trusted to return consistent values and to
+///! not crash or lead to undefined behaviour when presented with valid arguments.
+///! * the `Mbox` crate is trusted to perform operations safely on the pointers provided to it, if
+///! the pointers are trusted to be valid.
+///! * methods not marked `unsafe` are trusted to behave safely, potentially returning appropriate
+///! erorr messages when encountering any problems.
+///! * whenever `unwrap`, `expect`, `panic` or derivatives of these are used, they need to be
+///! thoroughly documented and justified - preferably `unwrap` and `expect` should *never* fail
+///! during normal operation.
+///! * these rules can be broken in test-only code and in tests.
 
 #[allow(
     non_snake_case,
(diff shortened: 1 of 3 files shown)


src/utils.rs
```diff
@@ -425,15 +422,13 @@ pub struct Signature {
     pub signature: Vec<u8>,
 }
 
-impl TryFrom<TPMT_SIGNATURE> for Signature {
-    type Error = Error;
-
-    fn try_from(tss_signature: TPMT_SIGNATURE) -> Result<Self> {
+impl Signature {
+    pub unsafe fn try_from(tss_signature: TPMT_SIGNATURE) -> Result<Self> {
```
hug-dev: I think that is a very good decision to put the `unsafe` on the method here as we can not trust that the `tss_signature` argument is always well-composed. The caller of this method needs to make the decision as only him knows where the parameter comes from.

hug-dev: I was thinking: "what if the method is not `public`?"
But even so, it would maybe be wrong to put it safe anyway.

src/lib.rs
```diff
@@ -659,14 +659,10 @@ impl Drop for Context {
...
-        unsafe {
-            tss2_esys::Tss2_TctiLdr_Finalize(
-                &mut tcti_context.into_raw() as *mut *mut TSS2_TCTI_CONTEXT
-            )
-        };
+        unsafe { tss2_esys::Tss2_TctiLdr_Finalize(&mut tcti_context.into_raw()) };
 
         // Close the context.
-        unsafe { tss2_esys::Esys_Finalize(&mut esys_context.into_raw() as *mut *mut ESYS_CONTEXT) };
+        unsafe { tss2_esys::Esys_Finalize(&mut esys_context.into_raw()) };
```
hug-dev: 💌

src/utils.rs
hug-dev: Maybe a comment would be nice to explain those things? Or that could be part of the documentation of the method itself.

src/lib.rs
```diff
@@ -412,7 +412,7 @@ impl Context {
 
         if ret.is_success() {
             let signature = unsafe { MBox::from_raw(signature) };
-            Ok((*signature).try_into()?)
+            Ok(unsafe { Signature::try_from(*signature)? })
```
hug-dev: Maybe a comment to explain why this is safe would make us better `unsafe` citizens 👷

hug-dev: Same everywhere we use the `unsafe` block actually in my opinion. The same way we kind of have to justify when we panic explicitely.

src/utils.rs
ionut-arm: Yeah, I decided to just leave it to documentation, a comment isn't as visible for users

src/lib.rs
ionut-arm: Indeed, will add

PR #19, opened 2019-12-19
Improve TransientObjectContext interface

ionut-arm:
This commit slightly improves the transient context by abstracting the
last type present in the interface that was defined in the FFI layer.

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -587,3 +587,101 @@ pub fn get_rsa_public(restricted: bool, decrypt: bool, sign: bool, key_bits: u16
         .build()
         .unwrap() // should not fail as we control the params
 }
+
+#[derive(Clone, Copy)]
+pub enum Hierarchy {
+    Null,
+    Owner,
+    Platform,
+    Endorsement,
+}
+
+impl Hierarchy {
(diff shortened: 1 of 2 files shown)


src/utils.rs
```diff
@@ -587,3 +587,101 @@ pub fn get_rsa_public(restricted: bool, decrypt: bool, sign: bool, key_bits: u16
...
+
+impl TryFrom<TPMT_TK_VERIFIED> for TpmtTkVerified {
+    type Error = Error;
+
+    fn try_from(tss_verif: TPMT_TK_VERIFIED) -> Result<Self> {
+        if tss_verif.tag != TPM2_ST_VERIFIED {
+            return Err(Error::local_error(WrapperErrorKind::InconsistentParams));
+        }
+
+        let len = tss_verif.digest.size.try_into().unwrap();
```
hug-dev: `unwrap` 😨 Maybe a comment would be nice to be consistent with our recent changes?

ionut-arm: Good spot, I'd forgotten

ionut-arm: Actually, that one's `u16` -> `usize` which can happen losslessly, so I'll just add a comment to the `unwrap` on line 682.

PR #20, opened 2019-12-23
Add documentation

ionut-arm:
This commit adds documentation to the features exposed in the create.


ionut-arm:
Should we add the constraints in the documentation for each method? (e.g. "`auth_value` can be at most 64 bytes long") 
@hug-dev

src/lib.rs
```diff
@@ -605,6 +715,11 @@ impl Context {
         }
     }
 
+    // TODO: Should we really keep `num_bytes` as `u16`?
```
hug-dev: Maybe we should put `num_bytes` as `u16`, for this level of abstraction, as it is how the API defines it.
We can still define a higher-level function that takes an `usize` if needed.

hug-dev: meaning that the `TODO` here is justified, not that you have to do the changes in this PR 😄

hug-dev:
> 
> 
> Should we add the constraints in the documentation for each method? (e.g. "`auth_value` can be at most 64 bytes long")
> @hug-dev

I think it would be better to do so!

That would be a good idea in the future to create an `AuthValue` new-type that checks that and use that type here so that, ideally, with any argument the function behaves correctly!

src/abstraction/transient.rs
```diff
@@ -171,6 +223,13 @@ impl TransientObjectContext {
...
     }
 
+    /// Sign a digest with an existing key.
+    ///
+    /// Returns the signature.
+    ///
+    /// # Errors
+    /// * errors are returned if any method calls return an error: `Context::context_load`,
+    /// `Context::sign`, `Context::flush_context`, `TransientObjectContext::set_session_attrs`
     pub fn sign(
```
ionut-arm: This method along with others around here make use of constraints that apply inside the methods it calls (e.g. in this case, the length of `key_auth` is checked in `set_handle_auth` so there's no need to check it here) - should we document this specifying the same constraints?

hug-dev: I think this is fine not to document it in the upper layer as well as you already say that "errors are returned if any method calls return an error".

On that note, maybe nice to add `set_handle_auth` to the list then?

ionut-arm:
I'll make adjustments to documentation as I go, merging this as is for now

PR #25, opened 2020-01-14
Add integration tests

ionut-arm:
This commit adds integration tests for the `Context` and
`TransientObjectContext` structures.

### DIFF
--- a/tests/context_tests.rs
+++ b/tests/context_tests.rs
@@ -1,0 +1,1135 @@
+// Copyright (c) 2020, Arm Limited, All Rights Reserved
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//          http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+const HASH: [u8; 64] = [
+    0x69, 0x3E, 0xDB, 0x1B, 0x22, 0x79, 0x03, 0xF4, 0xC0, 0xBF, 0xD6, 0x91, 0x76, 0x37, 0x84, 0x69,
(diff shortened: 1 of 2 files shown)


tests/context_tests.rs
```diff
@@ -0,0 +1,1147 @@
...
+            ESYS_TR_RH_OWNER,
+            &utils::get_rsa_public(true, true, false, 2048),
+            &key_auth,
+            &[],
+            &[],
+            &[],
+        )
+        .unwrap();
+
+    dbg!(prim_key_handle);
```
hug-dev: Do we want to keep this?

hug-dev: Same for the others `print`, `println` and `dbg` in this file

tests/context_tests.rs
```diff
@@ -0,0 +1,1147 @@
...
+    println!();
+    dbg!(
+        context
+            .verify_signature(key_handle, &HASH[..32], &signature.try_into().unwrap())
+            .unwrap()
+            .tag
+    );
+}
+
+mod test_start_sess {
```
hug-dev: Did you create a new module to launch those tests separately from the one above?

hug-dev: Could all tests be under:
```
#[cfg(test)]
mod tests {
    #[test]
    fn test() {
        ...
    }
}
```
(same way as we do in Parsec)

tests/context_tests.rs
```diff
@@ -0,0 +1,1147 @@
...
+                prim_key_handle,
+                &[],
+                TPM2_SE_HMAC,
+                utils::TpmtSymDefBuilder::aes_256_cfb(),
+                TPM2_ALG_SHA256,
+            )
+            .unwrap();
+    }
+
+    #[ignore]
```
hug-dev: Maybe it would be nice to explain with a comment why this test is ignored?

tests/context_tests.rs
ionut-arm: I did make a module per function that's tests, just to group them a bit, don't know if having all of them in one module makes any difference.
Also, the `cfg` isn't needed since this is a test file so it's by default only checked for tests

tests/context_tests.rs
ionut-arm: shall remove

tests/context_tests.rs
ionut-arm: Done!

tests/context_tests.rs
hug-dev: Fair enough!

tests/context_tests.rs
hug-dev: Super

PR #26, opened 2020-01-14
Add TransientObjectContext integration tests

hug-dev:
This commit adds a few integration tests for the transient abstraction
module.
Also adds a stronger check to limit the key sizes to 1024 and 2048
bytes.

### DIFF
--- a/tests/transient_object_context.rs
+++ b/tests/transient_object_context.rs
@@ -1,0 +1,189 @@
+// Copyright (c) 2020, Arm Limited, All Rights Reserved
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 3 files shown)


tests/transient_object_context.rs
```diff
@@ -0,0 +1,189 @@
...
+    let (key1, auth1) = ctx.create_rsa_signing_key(2048, 16).unwrap();
+    let (key2, _) = ctx.create_rsa_signing_key(2048, 16).unwrap();
+
+    // Sign with the first key
+    let signature = ctx.sign(key1.clone(), &auth1, &HASH).unwrap();
+
+    // Import and verify with the second key
+    let pub_key = ctx.read_public_key(key2.clone()).unwrap();
+    let pub_key = ctx.load_external_rsa_public_key(&pub_key).unwrap();
+    if let Tss2Error(error) = ctx.verify_signature(pub_key, &HASH, signature).unwrap_err() {
```
ionut-arm: Nice, I should add something like this to the `Context` tests too

PR #31, opened 2020-01-17
Run checks on ARM server

ionut-arm:
This commit adds a CI step to run the test scripts on an aarch64 server
provisioned for our use.


ionut-arm:
Don't work on arm :(

PR #32, opened 2020-01-17
Improve docs

ionut-arm:
This commit improves some of our documentation and allows the crate docs
to be built for docs.rs.


Cargo.toml
```diff
@@ -22,3 +24,9 @@ env_logger = "0.7.1"
 [build-dependencies]
 bindgen = "0.52.0"
 pkg-config = "0.3.17"
+
+[features]
+docs = []
```
hug-dev: Maybe we will also have to add some of thse dependencies as optional if they do not build on the docs.rs server.

ionut-arm: ... hopefully they do

hug-dev:
Clever fix to allow to have nice docs on docs.rs 😃 In the future we might want to load the library dynamically.

ionut-arm:
> Clever fix to allow to have nice docs on docs.rs 😃 In the future we might want to load the library dynamically.

Agreed, it would've been too much work not justified for now, though

PR #33, opened 2020-01-24
Remove 'unimplemented' use and fix lints

ionut-arm:
This commit removes all uses of the `unimplemented` macro and makes some
code improvements hinted by the linter.

### DIFF
--- a/src/response_code.rs
+++ b/src/response_code.rs
@@ -51,6 +51,7 @@ bitfield! {
 }
 
 /// Rust native representation of the TSS2 response codes as defined in the spec.
+#[allow(clippy::module_name_repetitions)]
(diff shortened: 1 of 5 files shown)


src/response_code.rs
```diff
@@ -51,6 +51,7 @@ bitfield! {
 }
 
 /// Rust native representation of the TSS2 response codes as defined in the spec.
+#[allow(clippy::module_name_repetitions)]
```
hug-dev: Could this be put at the crate root with `#![allow(clippy::module_name_repetitions)]` to avoi putting it in all modules?

ionut-arm: There might appear valid cases of it at some point, I found these two to be false-positives, but who knows

PR #39, opened 2020-03-06
Trim Context constructor; Improve ctx creation

### LINKED ISSUE #35: Create builders for context structures
At the moment context structures are built directly with a constructor method which makes some assumptions about some of the parameters inside, e.g. algorithms used in auth sessions.

This could be improved using some builders that parameterize the creation but also allow the use of default values. 

Existing constructor methods should be deprecated, but this would imply a breaking change - thoughts @hug-dev ?

ionut-arm:
This commit improves the creation of Contexts. The main constructor for
the basic Context structure was trimmed by removing "automatic" session
creation - any clients will now have to start their own session after
constructing the context.
A new builder was also added for the transient context, allowing finer
control over settings used within the context. `TransientObjectContext`
was also renamed to `TransientKeyContext`.

Closes #35


ionut-arm:
Closes #35

src/utils/primitives.rs
```diff
@@ -0,0 +1,162 @@
...
+            CipherMode::ECB => TPM2_ALG_ECB,
+        }
+    }
+}
+
+impl From<Cipher> for TPMT_SYM_DEF {
+    fn from(cipher: Cipher) -> Self {
+        let key_bits = match cipher {
+            Cipher::XOR { hash } => hash,
+            _ => cipher.key_bits().unwrap().into(), // should not fail since XOR is covered above
```
hug-dev: Even if a new `Cipher` with no `key_bits` is added in the future?

src/utils/primitives.rs
```diff
@@ -0,0 +1,162 @@
...
+            Cipher::XOR { .. } => TPM2_ALG_NULL,
+            _ => cipher.mode().unwrap().into(), // should not fail since XOR is covered above
+        };
+
+        TpmtSymDefBuilder::new()
+            .with_algorithm(cipher.algorithm_id())
+            .with_key_bits(key_bits)
+            .with_mode(mode)
+            .build_object()
+            .unwrap() // all params are strictly controlled, should not fail
```
hug-dev: What if we add new parameters to the builder 😀

ionut-arm: As noted above - this already covers everything there is to cover as parameters for these structures, as defined in the spec.

src/utils/primitives.rs
ionut-arm: 1) The only reason it's an option is because somehow they decided it's sane to include XOR in the Cipher category. Any other cipher will have to have *some* number of key bits.
2) The changes here already include everything defined in a finished spec. Hence, no new `Cipher` will need to be added.

PR #41, opened 2020-03-09
Remove explicit handles in StartSessionAuth

ionut-arm:
This commit adds a few fixes. The main improvement is due to a bugfix
being published upstream in a release. This bugfix allows auth sessions
to be started with the help of encryption sessions (so that parameters
are encrypted within the command). Thus, StartAuthSession no longer
needs to receive separate sessions handles and can now simply use the
ones stored on the Context. This does, however, mean that the most
recent version of the TSS libraries is needed.
Another fix is for memory allignment for 32bit machines for
TPMS_CONTEXT. When generating the bindings on 32 bit architectures, a
padding field that did not exist on 64 bit machines is added. Any
attempt to use the One True Constructor for said structure then fails if
the new field is not considered.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -16,7 +16,7 @@ use std::env;
 use std::path::PathBuf;
 
 // Minimum version of the TSS 2.0 libraries that this crate can use.
-const MINIMUM_VERSION: &str = "2.3.0";
+const MINIMUM_VERSION: &str = "2.3.3";
 
 fn main() {
     if cfg!(not(feature = "docs")) {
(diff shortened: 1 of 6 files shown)


build.rs
```diff
@@ -16,7 +16,7 @@ use std::env;
 use std::path::PathBuf;
 
 // Minimum version of the TSS 2.0 libraries that this crate can use.
-const MINIMUM_VERSION: &str = "2.3.0";
+const MINIMUM_VERSION: &str = "2.3.3";
```
hug-dev: 😎

hug-dev: `pkg-config` rocks!

ionut-arm: beware if you do any development work on your machine, you might need to update the library 😬

PR #42, opened 2020-03-09
Add TestParms method

### LINKED ISSUE #38: Implement TestParms function
`TPM2_TestParms` can be used to determine if the parameters for an object are supported by the underlying hardware. This would of great use to anyone trying to use the best/most secure available primitives without knowledge of the hardware capabilities a priori.

ionut-arm:
This commit adds the TestParms method to the ones supported by Context.
Users can now verify if their public parameters are supported by the
TPM before commiting to using them.

Closes #38

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -136,7 +136,7 @@ use std::convert::{TryFrom, TryInto};
 use std::ffi::CString;
 use std::ptr::{null, null_mut};
 use tss2_esys::*;
-use utils::{Signature, TpmaSession, TpmsContext};
+use utils::{PublicParmsUnion, Signature, TpmaSession, TpmsContext};
 
 #[macro_use]
 macro_rules! wrap_buffer {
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -796,6 +796,35 @@ impl Context {
...
+                self.sessions.1,
+                self.sessions.2,
+                &public_parms,
+            )
+        };
+
+        let ret = Error::from_tss_rc(ret);
+        if ret.is_success() {
+            Ok(())
+        } else {
```
hug-dev: Do we care about differenciating between the parameters not being compatible and a true error in the TSS stack? If so maybe a `Result<bool>` and a different method name (`are_parms_supported`) would be better but if not nevermind.

hug-dev: From the specs (the "Commands + Code" document):

> This command is used to check to see if specific combinations of algorithm parameters are supported.The  TPM  will  unmarshal  the  provided  TPMT_PUBLIC_PARMS.  If  the  parameters  unmarshal  correctly, then the TPM will return TPM_RC_SUCCESS, indicating that the parameters are valid for the TPM. The TPM will return the appropriate unmarshaling error if a parameter is not valid.

It might be too complicated for now to make the difference between the two. I thought a specific error would be returned if the parameters were not supported. Ignore this comment!

src/lib.rs
```diff
@@ -136,7 +136,7 @@ use std::convert::{TryFrom, TryInto};
 use std::ffi::CString;
 use std::ptr::{null, null_mut};
 use tss2_esys::*;
-use utils::{Signature, TpmaSession, TpmsContext};
+use utils::{PublicParmsUnion, Signature, TpmaSession, TpmsContext};
```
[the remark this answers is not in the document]
ionut-arm: Yes, they use this as the short version

src/lib.rs
ionut-arm: Yeah, I was gonna look up if there was any specific response code that matches that description (although, even then, there's no guarantee that the unmarshalling was wrong because the params are not supported, or because there's something wrong with how the structure was assembled.....)

src/utils/mod.rs
```diff
@@ -435,12 +436,47 @@ impl PublicIdUnion {
...
     RsaDetail(TPMS_RSA_PARMS),
     EccDetail(TPMS_ECC_PARMS),
     AsymDetail(TPMS_ASYM_PARMS),
 }
 
+impl PublicParmsUnion {
+    /// Get the object type corresponding to the value's variant.
+    pub fn object_type(&self) -> TPMI_ALG_PUBLIC {
+        match self {
+            PublicParmsUnion::AsymDetail(..) => TPM2_ALG_NULL,
```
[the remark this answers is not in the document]
ionut-arm: There's no definition of the `AsymDetail` variant being "selected"/represented through a `TPMI_ALG_PUBLIC` like the others are. So presumably that means `NULL`. :) 
The spec is full of these "The variants are all the same! Oh, apart from this one which behaves completely differently. Deal with it" moments

PR #45, opened 2020-03-11
Fix building for docs.rs

ionut-arm:
This commit fixes the issue underlying the docs.rs build failure for
v3.0.0 and adds documentation for the new `primitives` module.
It also adds a CI test for the documentation and building with a legacy
and the nightly compilers.


tests/all.sh
```diff
@@ -55,3 +50,28 @@ RUST_BACKTRACE=1 RUST_LOG=info cargo test -- --test-threads=1 --nocapture
...
+
+############################
+# Install legacy toolchain #
+############################
+rustup toolchain install 1.38.0
+
+####################
+# Verify doc build #
+####################
+cargo +nightly doc --features docs --verbose
```
hug-dev: Would that command fail if any of the dependency fail to build on nightly? Does docs.rs build with `--no-deps`?

tests/all.sh
```diff
@@ -55,3 +50,28 @@ RUST_BACKTRACE=1 RUST_LOG=info cargo test -- --test-threads=1 --nocapture
...
+
+########################
+# Verify nightly build #
+########################
+cargo +nightly build
```
hug-dev: Should we add a requirement on this crate that it has to build on nightly?

tests/all.sh
ionut-arm: ... I started making this change and then got carried away trying to do something else, forgot about actually adding that. Will do

tests/all.sh
ionut-arm: Maybe a note on the doc landing page that the crate is expected to build successfully under the nightly compiler and every Rust compiler since 1.38.0

PR #49, opened 2020-03-20
Allow the Context values to be unwrapped

ionut-arm:
This commit allows the Context objects to be unwrapped into a
ESYS_CONTEXT and one TSS2_TCTI_CONTEXT object. This would allow
developers to use said contexts with any of the methods found in the FFI
layer.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,7 +18,7 @@
 
 Contributions to the `tss-esapi` crate need to follow the process below.
 
-* Contributions are done through GitHub pull-requests.
+* Contributions are done through GitHub pull-requests using a [fork-based workflow](https://github.com/firstcontributions/first-contributions).
 * Contributors need to apply `rustfmt` and `clippy` to their Rust code.
 * New code needs to be tested with unit tests and/or integration tests when applicable and the global test script needs to pass.
 * The code is accepted under the [Developer Certificate of Origin](DCO.txt), so you must add following fields to your commit description:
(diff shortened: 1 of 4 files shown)


ionut-arm:
> 
> 
> Is the problem that we do not have enough abstractions for all of the ESYS APIs? This provides a way for developpers to opt-out of our abstractions and use the FFI structures directly, is that corect?

Correct - and they could, presumably, do that directly anyway, but it offers an option to use the `Context` structure initially, for setup, after which you can unwrap and use the raw contexts for whatever else is needed (e.g. creating a primary key and 2 AES keys, and then unwrapping the context and using the `ESYS_CONTEXT` to encrypt data)

hug-dev:
I see. Maybe to make it clearer it would be nice to have an independant function that does the unwrapping? I think using `From` implementation to do this might be a bit confusing? The documentation could also maybe be `unsafe` and clearly state that users of this functions will have to do the necessary clean up steps. It would also be nice to say that this method is only temporary and will be removed once the abstraction is done?
Also, do we care about returning the other fields of the `Context` structure?

ionut-arm:
> Maybe to make it clearer it would be nice to have an independant function that does the unwrapping?

Probably better. Closing this as I have a better idea, will reopen when I get around to implementing it.

Quick sketch of the idea:
I'll restructure the `Context`, taking out the two FFI context structures (`ESYS_CONTEXT` and `TSS2_TCTI_CONTEXT`) and the handle set, and put them into a `RawContext` structure. This raw context will become a private field in `Context` and will allow it to work as it does now, but when "unwrapping" you'll get the `RawContext` which you can then use to get `*mut ESYS_CONTEXT` and keep track of your open handles (manually). Context closing will be implemented on the `Drop` of `RawContext`, so that will not fall onto the user anymore, but cleaning up the outstanding handles will still be the responsability of the user.

PR #53, opened 2020-03-23
allow to get ESYS_CONTEXT from Context

### LINKED ISSUE #50: Allow Context to be unwrapped to ESYS_CONTEXT
Usage of the `Context` structure is fairly limited by the number of methods implemented on it. This restriction could be lifted if the `Context` could be unwrapped to expose the `ESYS_CONTEXT` beneath. 

This should be a one-way call, as once this is done, we can no longer keep track of generated handles. 

My suggestion would be to bundle up the Esys and TCTI contexts together with the handle set in a "shell" structure that is only used to access them. This would be incorporated in the `Context` to let it work as usual. Additionally, closing the contexts could be delegated to a custom implementation of `Drop` for the new structure.

genofire:
shout be fixed #50

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -874,7 +874,7 @@ impl Context {
     }
 
     /// Returns a mutable reference to the native ESYS context handle.
-    fn mut_context(&mut self) -> *mut ESYS_CONTEXT {
+    pub fn mut_context(&mut self) -> *mut ESYS_CONTEXT {
         self.esys_context.as_mut().unwrap().as_mut_ptr() // will only fail if called from Drop after .take()
     }
 }
--- a/src/response_code.rs
+++ b/src/response_code.rs
@@ -461,7 +461,7 @@ pub enum Error {
 }
 
 impl Error {
-    pub(crate) fn from_tss_rc(rc: TSS2_RC) -> Self {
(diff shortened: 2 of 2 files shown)


ionut-arm:
I was thinking of doing this, but then the Drop method would end up not properly cleaning up the handles being created.

Maybe we should add a method to add new handles to the set stored by the `Context`?

ionut-arm:
I think we'll hold off with this for now, the main reason being that I'd want to avoid leaking the internal state of the `Context`. Extracting the inner context would be safer and easier to reason about (within the context). 

If you want, you can modify this PR to change the internal structure of the `Context` as described in #50 , but that will be a fairly large change!

genofire:
I am not sure, maybe we use another function to create a safe context? If you do not like. Do not feel harm if you close this PR.

hug-dev:
What is the status of this? I think our general view is that we want to "leak" or use as little as `bindgen` structures as possible and create abstractions for those. If there were any TSS operations needing the raw `ESYS_CONTEXT` that you wanted to use, do not hesitate making PRs for those, or raising them as issues 😃

genofire:
so i close this PR now?

hug-dev:
I think so, if you don't mind!

PR #54, opened 2020-03-23
Added more TPM Resource API:s

Superhepper:
Added more of the ESYS TPM Resource API:s. Some of the API:s already existed with
some extended functionality or a different name. I changed this to call new functions
that are only thin wrappers around the original C API:s. The reason for the changes is
to keep the naming consistent. So that the TPM Resource functions are prefixed with 'tr'.

### DIFF
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -647,44 +647,72 @@ impl TryFrom<Signature> for TPMT_SIGNATURE {
 }
 
 /// Rust native wrapper for session attributes objects.
-///
-/// If no mask has been specified then it will be
-/// defaulted to be have the same value as the flags.
 #[derive(Copy, Clone, Debug, Default)]
 pub struct TpmaSession {
+    flags: TPMA_SESSION,
+    mask: TPMA_SESSION,
+}
+
+impl TpmaSession {
+    // Clones the TpmaSession but adds a new mask.
+    pub fn clone_with_new_mask(self, new_mask: TPMA_SESSION) -> TpmaSession {
+        TpmaSession {
+            flags: self.flags,
+            mask: new_mask,
+        }
+    }
+
+    /// Function to retrieve the masks
+    /// that have been set.
+    pub fn mask(self) -> TPMA_SESSION {
+        self.mask
+    }
+
+    /// Function to retrive the flags that
+    /// gave been set.
+    pub fn flags(self) -> TPMA_SESSION {
+        self.flags
+    }
+}
+
+/// A builder for TpmaSession.
+///
+/// If no mask have been added then the mask
+/// will be set till all flags.
+#[derive(Copy, Clone, Debug, Default)]
+pub struct TpmaSessionBuilder {
     flags: TPMA_SESSION,
     mask: Option<TPMA_SESSION>,
 }
 
-impl TpmaSession {
-    /// Create a new session attributes object.
-    pub fn new() -> TpmaSession {
-        TpmaSession {
+impl TpmaSessionBuilder {
+    pub fn new() -> TpmaSessionBuilder {
+        TpmaSessionBuilder {
             flags: 0,
             mask: None,
         }
     }
 
-    /// Set flag.
+    /// Function used to add flags.
     pub fn with_flag(mut self, flag: TPMA_SESSION) -> Self {
         self.flags |= flag;
         self
     }
 
-    /// Set a mask
+    /// Function used to add masks.
     pub fn with_mask(mut self, mask: TPMA_SESSION) -> Self {
         self.mask = Some(self.mask.unwrap_or(0) | mask);
         self
     }
 
-    /// Get mask.
-    pub fn mask(self) -> TPMA_SESSION {
-        self.mask.unwrap_or(self.flags)
-    }
-
-    /// Get all set flags.
-    pub fn flags(self) -> TPMA_SESSION {
-        self.flags
+    /// Function used to build a TpmaSession
+    /// from the parameters that have been
+    /// added.
+    pub fn build(self) -> TpmaSession {
+        TpmaSession {
+            flags: self.flags,
+            mask: self.mask.unwrap_or(self.flags),
+        }
     }
 }
 
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -647,29 +647,44 @@ impl TryFrom<Signature> for TPMT_SIGNATURE {
 }
 
 /// Rust native wrapper for session attributes objects.
+///
+/// If no mask has been specified then it will be
+/// defaulted to be have the same value as the flags.
 #[derive(Copy, Clone, Debug, Default)]
-pub struct TpmaSession(TPMA_SESSION);
+pub struct TpmaSession {
+    flags: TPMA_SESSION,
+    mask: Option<TPMA_SESSION>,
+}
 
 impl TpmaSession {
     /// Create a new session attributes object.
     pub fn new() -> TpmaSession {
-        TpmaSession(0)
+        TpmaSession {
+            flags: 0,
+            mask: None,
+        }
     }
 
     /// Set flag.
     pub fn with_flag(mut self, flag: TPMA_SESSION) -> Self {
-        self.0 |= flag;
+        self.flags |= flag;
         self
     }
 
-    /// Get mask for all set flags.
+    /// Set a mask
+    pub fn with_mask(mut self, mask: TPMA_SESSION) -> Self {
+        self.mask = Some(self.mask.unwrap_or(0) | mask);
+        self
+    }
+
+    /// Get mask.
     pub fn mask(self) -> TPMA_SESSION {
-        self.0
+        self.mask.unwrap_or(self.flags)
     }
 
     /// Get all set flags.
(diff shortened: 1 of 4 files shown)


ionut-arm:
I'm guessing the reason for adding the methods for setting session attributes using the raw FFI objects (`TPMA_SESSION`) is because you wanted to have different values for `mask` and `flags`? 

Also, out of curiosity - what's the use case for having `tr_set_session_attr` and `set_session_attr` in parallel? Do you need both?

Superhepper:
1. Yes, that is correct. And I also thought it would be a good idea to keep the API as close to the C api as possbile. That way it would be easy to add more utility functions later that make usage easier but still have 'core' intact. I am not sure it is a good idea but that was the thought.

2. I hope you ment `tr_sess_set_attributes`. The reason for having both is that I was afraid that I would break existing code (I am not sure that is something that should be considered). And `set_session_attr` is nice to have around. If one uses the TpmSession struct then it is nice to be able to simply pass that object tot he function.

3. I was also thinking that it would be nice in the future to add `get_session_attr` that could return a TpmSession.

4. `tr_set_auth` and `set_handle_auth` is basicly just a backward compatible rename.

ionut-arm:
I think breaking compatibility at this point shouldn't be a problem (although we'll need to think of a better way of labelling versions). 

Having two methods that essentially do the same thing (just with different parameters) sounds confusing to me, so I'd stick with just having one that is more usable. However! at the same time, one of the main goals of the `Context` structure (and actually of this library as a whole) is to present a fully Rust-native interface to the user - i.e. no types generated by `bindgen`. That's because a lot of those types are not optimal in the Rust world. Fir example `TPM2B_DATA` or other such structures which contain just a fixed-sized buffer and the size of the data in that buffer; that kind of data structure is actually harmful in Rust, because it can introduce an inconsistency, between how much data you put in the buffer and what you put in the length field; this was just an example as to why we're trying to clean up the interface of these types. Plus, the TPM spec is really difficult to cope with in terms of edge-cases and "same-but-different" situations. The `tss_esys` and `constants` modules are public, so anyone wishing to use them is free to do so!

My suggestions would be: 
* rename the methods to be closer to the TSS API
* when creating new methods/modifying old ones, try to reduce the number of FFI types in the interface as much as possible
These suggestions apply to all PRs, although maybe the first point should be condensed in a PR of its own. That would help with verifying that all methods are named consistently.

For the second point, and in this case with `TPMA_SESSION` as an example, I would try and do the following:
* keep the masks and flags parameters separate, that is probably the sane option
* instead of using `TPMA_SESSION`, please try and implement something similar to `ObjectAttributes` in `utils/mod.rs`; by that I mean a [bitfield ](https://crates.io/crates/bitfield) that encloses a `TPMA_SESSION` object and exposes setters and getters for all the different flag values. 
e.g.
```
bitfield! {
    pub struct SessionAttributes(TPMA_SESSION);
    impl Debug;
    // Object attribute flags
    pub continue_session, set_continue_session: 0;
    pub audit_exclusive, set_audit_exclusive: 1;
    pub audit_reset, set_audit_reset: 2;
    pub decrypt, set_decrypt: 5;
    pub encrypt, set_encrypt: 6;
    pub session_audit, set_session_audit: 7;
}
```
It's [safer and more Rusty than using simple `u8` values](https://henrikwarne.com/2020/03/22/secure-by-design/).

Superhepper:
Oh. I was not aware that one of the main goals was to present a fully rust native interface. But it makes perfect sense! When I have the time I will completely change the code I have written.

Superhepper:
Should I close this for now and com back with a new PR when I have done some more work to incorporate your comments?

ionut-arm:
Not really - you can either amend your commit (or push more commits on top), no need to close-reopen!

Superhepper:
Removed duplicated functions and kept the ones with names close to the TPM2 Resource API names.

Modified the TpmaSession struct. To work with the case when you want to use a mask. If no mask is added then it works as previously.

tr_sess_get_attributes now returns a TpmaSession object.


I think with these changes I have covered your concerns. I do not think this PR would be best place to look over all the names.

src/utils/mod.rs
```diff
@@ -647,29 +647,44 @@ impl TryFrom<Signature> for TPMT_SIGNATURE {
...
 
     /// Set flag.
     pub fn with_flag(mut self, flag: TPMA_SESSION) -> Self {
-        self.0 |= flag;
+        self.flags |= flag;
+        self
+    }
+
+    /// Set a mask
+    pub fn with_mask(mut self, mask: TPMA_SESSION) -> Self {
```
ionut-arm: Do you think it'd be more useful to replace the existing mask completely and maybe have a separate method like `add_mask_flags` to add new ones? Don't know if there's any use case for removing flags.

Superhepper: Hrrm... I think you are some what correct. Though the `with_mask` cannot clear the mask in  `TpmaSession` that would make behave differently then `with_flag` which would be confusing.

A `set_mask` would be logical for a function that clears the mask and sets it.

Superhepper: Or rather I do not know. The `TpmaSession` struct is currently more of a builder. It would make more sense to split it up in a `TpmaSession` and a `TpmaSessionBuilder`. 

The `TpmaSession` should have the possibility to create a new `TpmaSession` out of an old one but with a specific mask.

Something like `copy_with_mask` or `mask_specific_copy` or something like that.


rust```
fn foo(tpma_session: &TpmaSession) {
context.tr_sess_set_attributes(session, tpma_session.copy_with_mask(MY_FIRST_MASK));
.
.
.
context.tr_sess_set_attributes(session, tpma_session.copy_with_mask(MY_SECOND_MASK));
}
```

I think that would make the most sense.

Comment (ionut-arm) [reply src/utils/mod.rs]: I think the name was mostly drawn from the builder pattern, although in this case the structure is its own builder. Maybe it would make more sense to implement `From<TPMA_SESSION> for TpmaSession` and change these ones to just do the correct thing implied by their name (either overwrite or just set some particular flags) - i.e. not return `Self` as well

Comment (ionut-arm) [reply src/utils/mod.rs]: Agree, would be ok with removing the "builder" methods and replacing them with setters and a `clone_with_mask(TPMA_SESSION)` method (clone and not copy because [copy is supposed to be implicit](https://doc.rust-lang.org/std/clone/trait.Clone.html))

Comment (Superhepper) [reply src/utils/mod.rs]: I do not really see a reason to remove the builder. I was thinking to have it like this:

```rust
#[derive(Copy, Clone, Debug, Default)]
pub struct TpmaSession {
    flags: TPMA_SESSION,
    mask: TPMA_SESSION,
}

impl TpmaSession {
    // Clones the TpmaSession but adds a new mask.
    pub fn clone_with_mask(self, mask: TPM_SESSION) -> TmaSession {
        TpmaSession {
            flags: self.flags,
            mask: mask,
        }
    }

    /// Get mask.
    pub fn mask(self) -> TPMA_SESSION {
        self.mask
    }

    /// Get all set flags.
    pub fn flags(self) -> TPMA_SESSION {
        self.flags
    }
}

/// A builder for TpmaSession.
///
/// If no mask have been added then the mask
/// will be set till all flags.
#[derive(Copy, Clone, Debug, Default)]
struct TpmaSessionBuilder {
    flags: TPMA_SESSION,
    mask: Option<TPMA_SESSION>,
}

impl TpmaSessionBuilder {
    pub fn new() -> TpmaSession {
        TpmaSession { flags: 0, mask: 0 }
    }

    /// Set flag.
    pub fn with_flag(mut self, flag: TPMA_SESSION) -> Self {
        self.flax |= flag;
        self
    }

    /// Set a mask
    pub fn with_mask(mut self, mask: TPMA_SESSION) -> Self {
        self.mask = (self.mask.unwrap_or(0) | mask);
        self
    }

    pub fn build(self) -> TpmaSession {
        TpmaSession {
            flags: self.flags,
            mask: self.mask.unwrap_or(self.flags),
        }
    }
}
```

Superhepper:
Added the code for TpmaSession and TpmaSessionBuilder.

src/utils/mod.rs
```diff
@@ -648,28 +648,71 @@ impl TryFrom<Signature> for TPMT_SIGNATURE {
 
 /// Rust native wrapper for session attributes objects.
 #[derive(Copy, Clone, Debug, Default)]
-pub struct TpmaSession(TPMA_SESSION);
+pub struct TpmaSession {
```
hug-dev: Just a small question: this structure seems to represent session attributes, and is name `TpmaSession`. The variable name holding this structure value is often "attributes" or "session_attrs". Could we rename it `TpmaSessionAttributes`? Is it to match with the name it is given in the TSS?

ionut-arm: I guess we could, the name was initially based on the TSS structure, but since we're doing something more abstract now, we can drop the TSS naming convention (it might even be confusing if we use the same name but offer different functionality)

Superhepper: I agree. But most of the names in the code is using the TSS names. But it would make sense to rename it to something like SessionAttributes instead.

hug-dev: I don't mind you doing it in this PR or another one 😃 Or just raising an issue.

PR #56, opened 2020-03-23
Improve RSA parameter creation

ionut-arm:
This commit improves the way RSA parameter structures are created,
making it easier to create "types" of keys with only the required
parameters.
These changes were also propagated through to the transient key context,
allowing it to control the parameters and making it easier to implement
finer control over the keys in the future.


ionut-arm:
> 
> 
> I think it looks nice. But one thing I am thinking about is why key sizes is limited t0 1024 or 2048 another valid size would be 4096 and I think that size should be included as well.

Done

src/utils/mod.rs
```diff
@@ -886,3 +913,66 @@ impl TryFrom<TpmtTkVerified> for TPMT_TK_VERIFIED {
...
+    object_attributes.set_fixed_parent(true);
+    object_attributes.set_sensitive_data_origin(true);
+    object_attributes.set_user_with_auth(true);
+    object_attributes.set_decrypt(true);
+    object_attributes.set_sign_encrypt(false);
+    object_attributes.set_restricted(true);
+
+    Tpm2BPublicBuilder::new()
+        .with_type(TPM2_ALG_RSA)
+        .with_name_alg(TPM2_ALG_SHA256)
```
hug-dev: Should we indicate in the function documentation that those parameters are hardcoded or is it fine because it's a constant throughout the whole crate?

src/utils/mod.rs
```diff
@@ -886,3 +913,66 @@ impl TryFrom<TpmtTkVerified> for TPMT_TK_VERIFIED {
         })
     }
 }
+
+/// Create the TPM2B_PUBLIC structure for a restricted decryption key.
```
hug-dev: By "decryption key", do you mean a public key to verify signatures?

hug-dev: Or does it do both?

src/utils/mod.rs
```diff
@@ -177,48 +159,55 @@ impl Default for Tpm2BPublicBuilder {
...
+    /// Size of key, in bits
+    pub key_bits: TPMI_RSA_KEY_BITS,
+    /// Public exponent of the key. A value of 0 defaults to 2 ^ 16 + 1
+    pub exponent: u32,
+    /// Flag indicating whether the key shall be used for signing
+    pub for_signing: bool,
+    /// Flag indicating whether the key shall be used for decryption
+    pub for_decryption: bool,
+    /// Flag indicating whether the key is restricted
+    pub restricted: bool,
```
hug-dev: Could those fields be private so that we can control that only builders with valid parameters are created?

src/abstraction/transient.rs
```diff
@@ -363,22 +372,23 @@ impl TransientKeyContextBuilder {
...
     /// * if the root key authentication size is given greater than 32 or if the root key size is
-    /// not 1024 or 2048, a `WrongParamSize` wrapper error is returned
+    /// not 1024, 2048 or 4096, a `WrongParamSize` wrapper error is returned
     pub unsafe fn build(self) -> Result<TransientKeyContext> {
         if self.root_key_auth_size > 32 {
             return Err(Error::local_error(ErrorKind::WrongParamSize));
         }
-        if self.root_key_size != 1024 && self.root_key_size != 2048 {
+        if self.root_key_size != 1024 && self.root_key_size != 2048 && self.root_key_size != 4096 {
             error!("The reference implementation only supports key sizes of 1,024 and 2,048 bits.");
```
hug-dev: I think we first set it this way because of the message in the `error` below, that the reference implementation is only supporting key sizes of 1024 and 2048 bits and that we were targeting that reference implementation. Is that fine to add 4096 here?

src/utils/mod.rs
ionut-arm: From the spec:
>A hierarchy is rooted in a secret seed key, kept in the TPM. To create a hierarchy of keys, the seed key (Primary Seed) is used to generate a key that uses a specific set of algorithms. If this key is a **restricted decryption key**, then it is a Parent  Key. If it  is not a Derivation Parent, then  it is  a Storage Parent under which other objects may be created or attached.A Storage Parent provides protection for the sensitive area in another object (a child) when that object is stored  outside  of  the  TPM.  Protection  is  provided  by  symmetric  encryption  and  HMAC-based  integrity protection of the sensitive area. There are two different cases for sensitivearea encryption: storage and duplication.

src/utils/mod.rs
ionut-arm: The thing about builders is that you can't check that they're valid at point X in time. The only time where it makes sense to check is when `build` is called. Otherwise you have to do the same checks every time when a new parameter is added and there might be things that are impossible to do because you'd need to change two parameters simultaneously to make the builder valid.

src/abstraction/transient.rs
ionut-arm: "The reference implementation" is only really a simulator, ideally we'd be targeting physical implementations. Tbh implementations could support any size, not only powers of two, and now that we have `TestParms` we can just relegate that check to users...

src/utils/mod.rs
ionut-arm: It should be in the docs, sorry!

src/abstraction/transient.rs
hug-dev: We might have had that discussion before but in my mind "implementation" in that context means the ESYS API implementation, ie the library that we link to in this crate.

hug-dev: As in, even before the commands each the physical TPM, the calls might fail in the TSS because this size is not supported. At least that's what I understood from this sentence.

ionut-arm: ah, true, I forgot about that; will check again just to make sure

ionut-arm: Interestingly, this time the specs took me in a different direction.

So `TPMI_RSA_KEY_BITS` is supposed to be the maximum size for a specific RSA key, and implementations should allow any value that is one of `RSA_KEY_SIZES_BITS` which are the predefined allowed values for RSA key sizes. This constant is defined [here](https://trustedcomputinggroup.org/wp-content/uploads/TCG_TPM2_r1p59_Part4_SuppRoutines_code_pub.pdf) on page 634 and allows RSA keys to be of 1024, 2048, 3072 or 4096 bits. So, if anything, we're actually missing one value. I think the TSS might default to a maximum of 2048 but might allow for up to 4096 if compiled with certain params?

hug-dev: oh good to know! I guess it is fine to support that then.

ionut-arm: I've updated this to cover 3072 bits as well and changed the way the check is done

PR #57, opened 2020-03-24
Thin wrapper around Esys_PCR_Read

Superhepper:
Made a thin wrapper around Esys_PCR_Read.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -768,13 +768,13 @@ impl Context {
     }
 
     pub fn pcr_read(
-        &mut self, 
+        &mut self,
         pcr_selection: &TPML_PCR_SELECTION,
     ) -> Result<(UINT32, TPML_PCR_SELECTION, TPML_DIGEST)> {
-        let mut pcr_update_counter : u32 = 0;
+        let mut pcr_update_counter: u32 = 0;
         let mut pcr_selection_out = null_mut();
         let mut pcr_values = null_mut();
-        let ret = unsafe { 
+        let ret = unsafe {
             Esys_PCR_Read(
                 self.mut_context(),
                 self.sessions.0,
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -767,6 +767,38 @@ impl Context {
         }
     }
 
+    pub fn pcr_read(
+        &mut self,
+        pcr_selection: &TPML_PCR_SELECTION,
+    ) -> Result<(UINT32, TPML_PCR_SELECTION, TPML_DIGEST)> {
```
ionut-arm: It's ok for now since these params look fairly awkward to Rustify, but in the future the plan'll be to make these Rust-native (as per my comment on the other PR!). Just as a heads-up

src/lib.rs
```diff
@@ -767,6 +767,38 @@ impl Context {
...
+                &mut pcr_update_counter,
+                &mut pcr_selection_out,
+                &mut pcr_values,
+            )
+        };
+        let ret = Error::from_tss_rc(ret);
+
+        if ret.is_success() {
+            let pcr_selection_out =
+                unsafe { MBox::<TPML_PCR_SELECTION>::from_raw(pcr_selection_out) };
```
ionut-arm: Interesting, does the compiler not do the type inference automatically here? It worked for me to just do `MBox::from_raw(...)`

src/lib.rs
```diff
@@ -767,6 +767,38 @@ impl Context {
...
+            let pcr_values = unsafe { MBox::<TPML_DIGEST>::from_raw(pcr_values) };
+            Ok((pcr_update_counter, *pcr_selection_out, *pcr_values))
+        } else {
+            error!("Error in creating derived key: {}.", ret);
```
ionut-arm: Not critical to change and I'll actually raise an issue to look at all these error messages we log (since some are wrong) - but this should probably be something else 😄

tests/context_tests.rs
```diff
@@ -291,6 +291,26 @@ mod test_start_sess {
     }
 }
 
+mod test_pcr_read {
+    use super::*;
+
+    #[test]
+    fn test_pcr_read_command() {
```
ionut-arm: 👍

Superhepper:
The compiler could probably deduct the type. I just wanted to be clear.

That error message is bad copy paste bug!

As you commented out in my other PR we should Rustify the return types. I need to work on that with this one. And the input as well.

ionut-arm:
No, I think it's ok to keep it as is right now and raise a new issue to do it later, that's what happened to a lot of the other ones - unless you're 100% set on Rustifying it now?

Superhepper:
Uhm I think I will do it now.

I will close my other PR because the change I made there were to big. And to much not in line with the goals of the project.

PR #63, opened 2020-04-08
Pcr selection improvement

Superhepper:
I have created a utilities (PcrSelectionsBuilder) to make it easier to create TPML_PCR_SELECTION.

I am not really pleased with how it turned out(but it works) and I would really like some input on how I can improve things.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,5 +49,4 @@ This project uses the following third party crates:
 * mbox (MIT)
 * bitfield (MIT and Apache-2.0)
 * pkg-config (MIT and Apache-2.0)
-* num-traits (MIT/Apache-2.0)
-* num-derive (MIT/Apache-2.0)
+* enumfags2 (MIT OR Apache-2.0)
--- a/README.md
+++ b/README.md
@@ -49,4 +49,4 @@ This project uses the following third party crates:
 * mbox (MIT)
 * bitfield (MIT and Apache-2.0)
 * pkg-config (MIT and Apache-2.0)
-* enumfags2 (MIT OR Apache-2.0)
+* enumflags2 (MIT and Apache-2.0)
--- a/README.md
+++ b/README.md
@@ -49,3 +49,5 @@ This project uses the following third party crates:
 * mbox (MIT)
 * bitfield (MIT and Apache-2.0)
 * pkg-config (MIT and Apache-2.0)
+* num-traits (MIT/Apache-2.0)
+* num-derive (MIT/Apache-2.0)
--- a/README.md
+++ b/README.md
@@ -50,3 +50,5 @@ This project uses the following third party crates:
 * bitfield (MIT and Apache-2.0)
 * pkg-config (MIT and Apache-2.0)
 * enumflags2 (MIT and Apache-2.0)
+* num-derive (MIT and Apache-2.0)
+* num-traits (MIT and Apache-2.0)
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -15,10 +15,9 @@ use crate::response_code::{Error, Result, WrapperErrorKind};
 use crate::tss2_esys::*;
 use algorithm_specifiers::{Cipher, HashingAlgorithm};
 use bitfield::bitfield;
-use num_derive::FromPrimitive;
-use num_traits::FromPrimitive;
+use enumflags2::BitFlags;
 use serde::{Deserialize, Serialize};
-use std::collections::{HashMap, HashSet};
+use std::collections::HashMap;
 use std::convert::{TryFrom, TryInto};
 
 /// Helper for building `TPM2B_PUBLIC` values out of its subcomponents.
@@ -1014,45 +1013,37 @@ pub fn create_unrestricted_signing_rsa_public(
         .build() // should not fail as we control the params
 }
 
-#[derive(Hash, Eq, PartialEq, Copy, Clone, Debug, FromPrimitive)]
+// Enum with the bit flag for each PCR slot.
+#[derive(BitFlags, Hash, Debug, PartialEq, Clone, Copy)]
+#[repr(u32)]
 pub enum PcrSlot {
-    Slot0 = 0,
-    Slot1,
-    Slot2,
-    Slot3,
-    Slot4,
-    Slot5,
-    Slot6,
-    Slot7,
-    Slot8,
-    Slot9,
-    Slot10,
-    Slot11,
-    Slot12,
-    Slot13,
-    Slot14,
-    Slot15,
-    Slot16,
-    Slot17,
-    Slot18,
-    Slot19,
-    Slot20,
-    Slot21,
-    Slot22,
-    Slot23,
+    Slot0 = 0x0000_0001,
+    Slot1 = 0x0000_0002,
+    Slot2 = 0x0000_0004,
+    Slot3 = 0x0000_0008,
+    Slot4 = 0x0000_0010,
+    Slot5 = 0x0000_0020,
+    Slot6 = 0x0000_0040,
+    Slot7 = 0x0000_0080,
+    Slot8 = 0x0000_0100,
+    Slot9 = 0x0000_0200,
+    Slot10 = 0x0000_0400,
+    Slot11 = 0x0000_0800,
+    Slot12 = 0x0000_1000,
+    Slot13 = 0x0000_2000,
+    Slot14 = 0x0000_4000,
+    Slot15 = 0x0000_8000,
+    Slot16 = 0x0001_0000,
+    Slot17 = 0x0002_0000,
+    Slot18 = 0x0004_0000,
+    Slot19 = 0x0008_0000,
+    Slot20 = 0x0010_0000,
+    Slot21 = 0x0020_0000,
+    Slot22 = 0x0040_0000,
+    Slot23 = 0x0080_0000,
 }
 
 /// A struct representing pcr selections
-///
-///
-/// TPML_PCR_SELECTION is a structure that can contain up to
-/// 16 TPMS_PCR_SELECTION
-///
-/// The TPMS_PCR_SELECTION contains:
-/// - A hashing algorithm
-/// - A size of the select i.e. how many bytes of the variale
-/// bit mask to use.
-/// - A variable bitmask where each bit represents a pcr slot.
 ///
 /// The minimum number of octets allowed in a TPMS_PCR_SELECT.sizeOfSelect
 /// is not determined by the number of PCR implemented but by the
@@ -1062,20 +1053,17 @@ pub enum PcrSlot {
 #[derive(Debug, Default, Clone)]
 pub struct PcrSelections {
     size_of_select: u8,
-    items: HashMap<HashingAlgorithm, HashSet<PcrSlot>>,
+    items: HashMap<HashingAlgorithm, BitFlags<PcrSlot>>,
 }
 
 impl From<PcrSelections> for TPML_PCR_SELECTION {
     fn from(pcr_selections: PcrSelections) -> TPML_PCR_SELECTION {
         let mut ret: TPML_PCR_SELECTION = Default::default();
-        for (hash_algorithm, pcr_slots) in &pcr_selections.items {
-            ret.pcrSelections[ret.count as usize].hash = hash_algorithm.clone().into();
-            ret.pcrSelections[ret.count as usize].sizeofSelect = pcr_selections.size_of_select;
-            for &pcr_slot in pcr_slots {
-                let index: usize = (pcr_slot as usize) / 8;
-                let value: u8 = 1 << ((pcr_slot as u8) % 8);
-                ret.pcrSelections[ret.count as usize].pcrSelect[index] |= value;
-            }
+        for (hash_algorithm, pcr_slots) in pcr_selections.items {
+            let tpms_pcr_selection = &mut ret.pcrSelections[ret.count as usize];
+            tpms_pcr_selection.hash = hash_algorithm.clone().into();
+            tpms_pcr_selection.sizeofSelect = pcr_selections.size_of_select;
+            tpms_pcr_selection.pcrSelect = pcr_slots.bits().to_le_bytes();
             ret.count += 1;
         }
         ret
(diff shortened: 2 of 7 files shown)


ionut-arm:
Thanks for the changes! I'll have a look in the morning (will have to get a deep-dive in those TPM spec structures as I'm not that familiar with them :) )

src/utils/mod.rs
```diff
@@ -1010,3 +1011,164 @@ pub fn create_unrestricted_signing_rsa_public(
...
+pub struct PcrSelections {
+    items: HashMap<HashingAlgorithm, HashSet<PcrSlot>>,
+}
+
+impl From<PcrSelections> for TPML_PCR_SELECTION {
+    fn from(pcr_selections: PcrSelections) -> TPML_PCR_SELECTION {
+        let mut ret: TPML_PCR_SELECTION = Default::default();
+        for (hash_algorithm, pcr_slots) in &pcr_selections.items {
+            ret.pcrSelections[ret.count as usize].hash = hash_algorithm.clone().into();
+            ret.pcrSelections[ret.count as usize].sizeofSelect = 3;
```
ionut-arm, line 1104: I'm weary about that hardcoded constant; from the spec:

> sizeofSelect indicates  the  number  of  octets  in pcrSelect.  The  allowable  value  for sizeofSelectis determined  by  the  number  of  PCR  required  by  the  applicable  platform-specific  specification  and  the number of PCR implemented in the TPM

[Mobile TPMs](https://trustedcomputinggroup.org/wp-content/uploads/TPM_2.0_Mobile_Common_Profile_v2r31_FINAL.pdf), for example, can contain as little as 8 PCRs. 

I think the safest bet here would be to either: 
* move this code into a method implemented on `PcrSelection` so that you can take this value in as a param
* add another field in `PcrSelections` that holds this value and then you can default to 3 if it's not set in the Builder

src/utils/mod.rs
```diff
@@ -1010,3 +1011,164 @@ pub fn create_unrestricted_signing_rsa_public(
...
+                }
+            }
+            let _ = ret.items.insert(
+                HashingAlgorithm::try_from(selection.hash).unwrap(),
+                pcr_slots,
+            );
+        }
+        ret
+    }
+}
```
ionut-arm, line 1137: I strongly suggest modifying this into a `TryFrom` because there are a few fields inside that structure that could be wrong and a `From` is strictly for cases where there is no possible failure or error (like in the `From` above where we control all inputs and we know they're safe - although that one has another issue as I noted) - in [docs](https://doc.rust-lang.org/std/convert/trait.From.html): "**Note: This trait must not fail.**"

You need to check that `tpml_pcr_selection.count`, for example, is a valid value. 

You're also expected to not use `.unwrap()` in cases where it's not provable that the result is `Ok(...)` - like when doing `HashingAlgorithm::try_from(...)`, as there's no guarantee that a user doesn't call this on a "bad" `TPML_PCR_SELECTION`. But that's easier to deal with if you swap for `TryFrom`

src/utils/mod.rs
```diff
@@ -1010,3 +1011,164 @@ pub fn create_unrestricted_signing_rsa_public(
...
+            for &pcr_slot in pcr_slots {
+                let index: usize = (pcr_slot as usize) / 8;
+                let value: u8 = 1 << ((pcr_slot as u8) % 8);
+                ret.pcrSelections[ret.count as usize].pcrSelect[index] |= value;
+            }
+            ret.count += 1;
+        }
+        ret
+    }
+}
```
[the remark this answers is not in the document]
Superhepper: Initially I was thinking of using [bitflags](https://docs.rs/bitflags/1.2.1/bitflags/). Backed by u32. But the I think at the time I did not want to deal with endian stuff (or add more dependecies to the project). So I chose not do it that way. Maybe I will reconsider.

src/utils/mod.rs
Superhepper: I think you are right and this is good idea and it has now been fixed.

src/utils/mod.rs
Superhepper: Yea, I agree. I have fixed that now.

src/utils/mod.rs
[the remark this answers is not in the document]
Superhepper: I ended up using [enumflags2::BitFlags](https://docs.rs/enumflags2/0.6.2/enumflags2/struct.BitFlags.html). Though there was a bug that I reported so now I need to wait for a release where it has been fixed.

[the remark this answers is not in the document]
Superhepper: This have now been fixed.

src/utils/mod.rs
```diff
@@ -1010,3 +1011,164 @@ pub fn create_unrestricted_signing_rsa_public(
...
+            18 => Ok(PcrSlot::Slot18),
+            19 => Ok(PcrSlot::Slot19),
+            20 => Ok(PcrSlot::Slot20),
+            21 => Ok(PcrSlot::Slot21),
+            22 => Ok(PcrSlot::Slot22),
+            23 => Ok(PcrSlot::Slot23),
+            _ => Err(Error::local_error(WrapperErrorKind::InvalidParam)),
+        }
+    }
+}
```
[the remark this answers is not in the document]
Superhepper: This have been fixed and changed when moving to [numflags2::BitFlags](https://docs.rs/enumflags2/0.6.2/enumflags2/struct.BitFlags.html).

src/utils/mod.rs
```diff
@@ -1010,3 +1011,164 @@ pub fn create_unrestricted_signing_rsa_public(
...
+        hash_algorithm: HashingAlgorithm,
+        pcr_slots: &[PcrSlot],
+    ) -> Self {
+        let selected_pcr_slots: HashSet<PcrSlot> = pcr_slots.iter().cloned().collect();
+        match self.items.get_mut(&hash_algorithm) {
+            Some(previously_selected_pcr_slots) => {
+                *previously_selected_pcr_slots = previously_selected_pcr_slots
+                    .union(&selected_pcr_slots)
+                    .cloned()
+                    .collect();
```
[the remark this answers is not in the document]
Superhepper: Extend would probably been the correct choice but it is no longer needed.

src/utils/mod.rs
Superhepper: Your are absolutely right. It was me being sloppy. It has been fixed now.

Superhepper:
Sorry for all the forced pushes but it was not trivial to try to add a signoff to commit that was a couple of commits back.

src/utils/mod.rs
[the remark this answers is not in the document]
ionut-arm: Nice!! I just looked at the issue you raised (linking [here](https://github.com/NieDzejkob/enumflags2/issues/22) for completeness) and it's a nice thing to catch for them! :)

ionut-arm:
> Sorry for all the forced pushes but it was not trivial to try to add a signoff to commit that was a couple of commits back.

Don't worry, we use force-push all the time 😄 It's probably more common than normal pushes, since we tend to work within the same commit for everything

src/utils/mod.rs
ionut-arm: Hey! Sorry for the late reply!

I've had a look, the swap to `TryFrom` is good, but it still needs the calls to `unwrap` to be changed to `?` - and for the case where you unwrap a `BitFlags::try_from` you can use [`or`](https://doc.rust-lang.org/std/result/enum.Result.html#method.or). Also, `selection_index` can trigger an "index out of bounds" panic as `tpml_pcr_selection.count` can be larger than the length of `tpml_pcr_selection.pcrSelections`

ionut-arm:
Just as a FYI - we're thinking that, since the interface is clearly not stable, we'll revert the crate version number to smething like `0.xx.yy`. It won't impact any of the functionality, but hopefully it won't confuse anyone. Will probably add a not about that on the main `README`.

README.md
```diff
@@ -49,3 +49,4 @@ This project uses the following third party crates:
 * mbox (MIT)
 * bitfield (MIT and Apache-2.0)
 * pkg-config (MIT and Apache-2.0)
+* enumfags2 (MIT OR Apache-2.0)
```
ionut-arm, line 52: I know that the repo/crate states that its license is MIT **or** Apache-2.0, but we decided to display it as **and** - the motivation being that since the license text allows both, contributors can choose either one, so technically the library will have both MIT and Apache-2.0 licensed code inside of it.

Superhepper: This have now been fixed. Also fixed the incorrect spelling of the name.

src/utils/mod.rs
```diff
@@ -1010,3 +1012,154 @@ pub fn create_unrestricted_signing_rsa_public(
...
+    fn try_from(tpml_pcr_selection: TPML_PCR_SELECTION) -> Result<PcrSelections> {
+        let mut ret: PcrSelections = Default::default();
+        let mut size_of_select: Option<u8> = None;
+        // Loop over available selections
+        for selection_index in 0..(tpml_pcr_selection.count as usize) {
+            let selection = &tpml_pcr_selection.pcrSelections[selection_index];
+            let pcr_slots: BitFlags<PcrSlot> =
+                BitFlags::<PcrSlot>::try_from(u32::from_le_bytes(selection.pcrSelect)).unwrap();
+            // Check for variations in sizeofSelect.
+            // Something that currently is not supported.
```
hug-dev, line 1084: Is is something that we do not support or is not supported in general? I am asking because we have the `UnsupportedParam` error as well.

src/utils/mod.rs
```diff
@@ -1010,3 +1012,154 @@ pub fn create_unrestricted_signing_rsa_public(
...
+impl TryFrom<TPML_PCR_SELECTION> for PcrSelections {
+    type Error = Error;
+    fn try_from(tpml_pcr_selection: TPML_PCR_SELECTION) -> Result<PcrSelections> {
+        let mut ret: PcrSelections = Default::default();
+        let mut size_of_select: Option<u8> = None;
+        // Loop over available selections
+        for selection_index in 0..(tpml_pcr_selection.count as usize) {
+            let selection = &tpml_pcr_selection.pcrSelections[selection_index];
+            let pcr_slots: BitFlags<PcrSlot> =
+                BitFlags::<PcrSlot>::try_from(u32::from_le_bytes(selection.pcrSelect)).unwrap();
```
hug-dev, line 1082: We try to not use panicking function in our code. Is it possible to remove the `unwrap` with something like:
```rust
.or_else(|e| {
    error!("Error converting pcrSelect to a BitFlags<PcrSlot>: {}.", e);
    Err(Error::local_error(WrapperErrorKind::InvalidParam))
})?;
```
And similar kind of things for all the `unwrap`. Ideally we always want to return an error instead of panicking.

src/utils/mod.rs
```diff
@@ -1010,3 +1012,154 @@ pub fn create_unrestricted_signing_rsa_public(
...
+    /// provided.
+    ///
+    /// If no size of select have been provided then it will
+    /// be defaulted to 3. This may not be the correct size for
+    /// the current platform. The correct values can be obtained
+    /// by quering the tpm for its capabilities.
+    pub fn build(self) -> PcrSelections {
+        let select_size = match self.size_of_select {
+            Some(value) => value,
+            None => 3,
```
hug-dev, line 1158: Do you think that's fine to put 3 there? We could also return an error if this value not set.

src/utils/mod.rs
hug-dev: Sorry did not mention and I think it is not imported in that file, but `error!` is imported with `use log::error;`

Superhepper: You are right I was thinking that it could not fail. Because I only made it possible to select 24 slots through the enum. But it is also possible to send in something that does not come from the tpm and try to convert it and in that case that would panic. I use your code and make it more robust against conversion of invalid data.

src/utils/mod.rs
Superhepper: 3 covers the slots specified by the enum. Future proof value according to the specification is 4. But right now I do not think that there is any one using that.

Returning an error would be nice if people using this knows what the value should be. Would it be enough to just put a recommended value in the documentation?

src/utils/mod.rs
Superhepper: I was actually not sure if it was even allowed to have different values for the sizeofSelect in the different selections. **So you are right it should be `UnsupportedParam`** until some one come and tell me that it is either something they need and it is allowed. Because then we need to rewrite the code or it is something that is not allowed by the specification and then we need to change the error to `InvalidParam`.

src/utils/mod.rs
Superhepper: We could also do something like this 

```rust
pub enum SelectSize {
One_Byte = 1,
Two_Bytes = 2,
Three_Bytes = 3,
Four_Bytes = 4,
}

impl Default for SelectSize {
fn default() -> SelecSize {
// Three bytes is a commonly used value
// That work on most platforms.
SelectSize::Three_Bytes
}
}
```

hug-dev: I am fine with just using 3 in that case! I was just worried where it came from. It might be nice to put it in a `const` value with a comment saying what you wrote: "Three bytes is a commonly used value that works on most plaftorms."

Superhepper: I kind of like the enum approach because it makes it harder to send in a value that is completely wrong. So i think it will be safer that way.

hug-dev: Are those four values the only ones allowed?

Superhepper: Yes. They dictate the size of a variable bit mask. The buffer for the bit mask is [4: u8]. With the sizeofSelect parameter to tell how many bytes to use. So no other values makes any sense.

hug-dev: Great! Then the enum with the `Default` seems like a good option to me!

Superhepper:
I think I have fixed all concerns and issues now.

src/utils/mod.rs
```diff
@@ -1062,7 +1084,7 @@ impl From<PcrSelections> for TPML_PCR_SELECTION {
         for (hash_algorithm, pcr_slots) in pcr_selections.items {
             let tpms_pcr_selection = &mut ret.pcrSelections[ret.count as usize];
             tpms_pcr_selection.hash = hash_algorithm.clone().into();
-            tpms_pcr_selection.sizeofSelect = pcr_selections.size_of_select;
+            tpms_pcr_selection.sizeofSelect = pcr_selections.size_of_select.to_u8().unwrap();
```
hug-dev, line 1087: Sorry to be annoying about `unwrap` again 😬 
We try to apply the following rule about panicking in our repos:
* remove all panicking behaviour if possible (`panic`, `expect`, `unwrap`, `unreachable`, etc) and return errors instead
* if in some location it is sure (proven) that a panic will not happen, we put a comment next to it saying why it is the case

We do it that way to make our libraries more stable as a whole, write them in what we think is a gode code quality practice but most importantly for security reasons: any panicking is a potential path opened to create a denial of service attack.

In that case, `enum PcrSelectSize` will probably not have a value outside of 0 - 255 so I'd say it is fine to panic, even better if there is a comment saying that 😋 

Don't worry about changing that now though! Just wanted to point out our "panicking guidelines".

src/utils/mod.rs
```diff
@@ -1074,32 +1096,63 @@ impl TryFrom<TPML_PCR_SELECTION> for PcrSelections {
...
-            let pcr_slots: BitFlags<PcrSlot> =
-                BitFlags::<PcrSlot>::try_from(u32::from_le_bytes(selection.pcrSelect)).unwrap();
+            // Parse the pcr slots from the pcrSelect bit mask.
+            let parsed_pcr_slots: BitFlags<PcrSlot> = BitFlags::<PcrSlot>::try_from(
+                u32::from_le_bytes(selection.pcrSelect),
+            )
+            .or_else(|e| {
+                error!("Error parsing pcrSelect to a BitFlags<PcrSlot>: {}.", e);
+                Err(Error::local_error(WrapperErrorKind::UnsupportedParam))
+            })?;
```
hug-dev, line 1110: Thanks for that!

PR #69, opened 2020-05-05
Add support for elliptic curves for transient context

ionut-arm:
This commit adds support for elliptic curve handling of the
TransientKeyContext.

It also adds and improves many of the underlying structures used to
access or provision said keys.


src/utils/mod.rs
```diff
@@ -274,6 +308,117 @@ impl TpmsRsaParmsBuilder {
 /// Supported sizes for RSA key modulus
 pub const RSA_KEY_SIZES: [u16; 4] = [1024, 2048, 3072, 4096];
 
+/// Builder for `TPMS_ECC_PARMS` values.
+// Most of the field types are from bindgen which does not implement Debug on them.
+#[allow(missing_debug_implementations)]
```
ionut-arm, line 313: This is no longer the case.

src/abstraction/transient.rs
```diff
@@ -456,3 +476,30 @@ impl Default for TransientKeyContextBuilder {
...
 }
+
+/// Parameters for the kinds of keys supported by the context
+#[derive(Debug, Clone, Copy)]
+pub enum KeyParams {
+    Rsa {
+        /// Size of key in bits
+        ///
+        /// Can only be one of: 1024, 2048, 3072 or 4096
+        size: u16,
```
hug-dev, line 487: Could we have a `RsaSize` enumeration for those four possible values so that the check line 104 is no longer needed?

src/utils/mod.rs
hug-dev: Nice! Does that mean we could remove this `allow`?

src/utils/mod.rs
```diff
@@ -274,6 +308,117 @@ impl TpmsRsaParmsBuilder {
...
+    pub for_signing: bool,
+    /// Flag indicating whether the key shall be used for decryption
+    pub for_decryption: bool,
+    /// Flag indicating whether the key is restricted
+    pub restricted: bool,
+}
+
+impl TpmsEccParmsBuilder {
+    /// Create parameters for a restricted decryption key (i.e. a storage key)
+    pub fn new_restricted_decryption_key(symmetric: Cipher, curve: EllipticCurve) -> Self {
```
hug-dev, line 332: What is the justification for the decryption one to be restricted and the signing one to be unrestricted?

src/utils/mod.rs
```diff
@@ -274,6 +308,117 @@ impl TpmsRsaParmsBuilder {
...
+        }
+
+        if self.for_decryption
+            && scheme.scheme != TPM2_ALG_SM2
+            && scheme.scheme != TPM2_ALG_ECDH
+            && scheme.scheme != TPM2_ALG_ECMQV
+            && scheme.scheme != TPM2_ALG_NULL
+        {
+            return Err(Error::local_error(WrapperErrorKind::InconsistentParams));
+        }
```
hug-dev, line 393: I ask this mostly because I am not an expert but it seems like here we are checking that the appropriate algorithm is select for each key type. Does that mean that different algorithms are used for signing and verification?

src/utils/mod.rs
```diff
@@ -705,13 +990,15 @@ impl TryFrom<Signature> for TPMT_SIGNATURE {
...
-                Err(Error::local_error(WrapperErrorKind::UnsupportedParam))
-            }
             AsymSchemeUnion::AnySig(_) => {
                 Err(Error::local_error(WrapperErrorKind::UnsupportedParam))
             }
+            AsymSchemeUnion::ECDH(_)
+            | AsymSchemeUnion::RSAES
+            | AsymSchemeUnion::RSAOAEP(_)
+            | AsymSchemeUnion::ECMQV(_) => {
+                Err(Error::local_error(WrapperErrorKind::InconsistentParams))
```
hug-dev, line 1000: Should that be `UnsupportedParam`?

src/utils/mod.rs
```diff
@@ -1013,7 +1300,33 @@ pub fn create_unrestricted_signing_rsa_public(
...
+}
+
+/// Create the TPM2B_PUBLIC structure for an ECC unrestricted signing key.
+///
+/// * `scheme` - Asymmetric scheme to be used for signing
+/// * `curve` - identifier of the precise curve to be used with the key
+pub fn create_unrestricted_signing_ecc_public(
+    scheme: AsymSchemeUnion,
+    curve: EllipticCurve,
+) -> Result<TPM2B_PUBLIC> {
```
hug-dev, line 1313: Is `TPM2B_PUBLIC` the last structure that we use and do not abstract? (Just for my information)

.gitignore
```diff
@@ -5,3 +5,4 @@ Cargo.lock
 *DS_Store
 *.patch
 *NVChip
+.vscode
```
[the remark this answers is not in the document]
ionut-arm: Sorry, I can't hear you over the sound of my IDE checking for and displaying any cargo issues automatically 😛

src/abstraction/transient.rs
ionut-arm: Yeah, I know, tried that with `Cipher` some time ago, but it feels overkill tbh... Could change if you want. Also the variant names feel weird af because you can't start a name with digits.

src/utils/mod.rs
ionut-arm: Restricted decryption keys are essentially Parent keys - they could be either used for derivation or storage of other object underneath it. The ECC one isn't used like that, so I could remove this

src/utils/mod.rs
ionut-arm: No, `for_signing` actually means `for_signing_and_verification` and decryption is the same with encryption. It essentially specifies what the private key is used for. 

This split here is because some scheme are for signing and others are for encryption. `SM2` can do both

src/utils/mod.rs
ionut-arm: Yes, and put `Debug` in derive

src/utils/mod.rs
ionut-arm: This is in a `impl TryFrom<Signature> for TPMT_SIGNATURE ` - meaning that only asymmetric schemes that are for signing make sense here; if the scheme on the `Signature` is an encryption one then there's an inconsistency.

That does, however, bring the question of whether the `AsymSchemeUnion` should be split into multiple enums - `AsymSignScheme` and `AsymEncryptScheme`, maybe even further, `RsaSignScheme`, `EccSignScheme` etc.

src/utils/mod.rs
ionut-arm: Yeah, `TPM2B_PUBLIC` is really deeply nested and convoluted - I guess it could be made up of a hierarchy of Rust native types, but it would be a large effort to allow all the possibilities that the original interface allows, while making it easier for a developer to build a structure that just works™️.

src/abstraction/transient.rs
hug-dev: I think it is worth to do it!
We could write the number in clear letters like:
```rust
enum RsaSize {
    OneThousandTwentyFour,
    TwoThousandFortyEight,
    ...
```

(joking)

Actually it could be possible

src/utils/mod.rs
hug-dev: Ahhh got it thanks!

src/utils/mod.rs
hug-dev: > This is in a impl TryFrom<Signature> for TPMT_SIGNATURE - meaning that only asymmetric schemes that are for signing make sense here; if the scheme on the Signature is an encryption one then there's an inconsistency.

That makes sense!

> That does, however, bring the question of whether the AsymSchemeUnion should be split into multiple enums - AsymSignScheme and AsymEncryptScheme, maybe even further, RsaSignScheme, EccSignScheme etc.

That would mean that the `scheme` field of `Signature` could be `AsymSignScheme` so you would not even have to do that check?

src/utils/mod.rs
hug-dev: Got it.

src/utils/mod.rs
ionut-arm: > That would mean that the scheme field of Signature could be AsymSignScheme so you would not even have to do that check?

True, but I've also realised that this would only work if we keep separate methods for creating/using signing and decryption keys. Otherwise in the `TransientKeyContext`, for example, we'd have to have different params types or variants for signing and encryption. 

I'm not a huge fan of it, but we can keep the thing as is and simply have these checks here - we have to do a full `match` in most cases anyway. I could implement methods `is_signing` and `is_decryption` on `AsymSchemeUnion` to make checks faster.

src/abstraction/transient.rs
ionut-arm: Yeah, for the symmetric ones I went for `Bits128`, `Bits192`... but anyway... I'll skip it for now

PR #70, opened 2020-05-07
Fix ECDSA signature variant name

ionut-arm:
This commit fixes the variant name for ECDSA-specific signatures.

### DIFF
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -865,7 +865,7 @@ pub struct Signature {
 #[derive(Debug, PartialEq)]
 pub enum SignatureData {
     RsaSignature(Vec<u8>),
-    EccSignature { r: Vec<u8>, s: Vec<u8> },
+    EcdsaSignature { r: Vec<u8>, s: Vec<u8> },
 }
 
 impl Signature {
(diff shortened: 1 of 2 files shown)


src/utils/mod.rs
```diff
@@ -865,7 +865,7 @@ pub struct Signature {
 #[derive(Debug, PartialEq)]
 pub enum SignatureData {
     RsaSignature(Vec<u8>),
-    EccSignature { r: Vec<u8>, s: Vec<u8> },
+    EcdsaSignature { r: Vec<u8>, s: Vec<u8> },
```
hug-dev, line 868: Should we increase the minor version number instead of the revision one because of that public change?

hug-dev: Assuming we follow the same rules than if we were before `1.0.0`:

> Before you reach 1.0.0, anything goes, but if you make breaking changes, increment the minor version. In Rust, breaking changes include adding fields to structs or variants to enums.

from [here](https://doc.rust-lang.org/cargo/reference/manifest.html#the-version-field)

So I think it is fine like this and happy to see this merged!

Issue #64, opened 2020-04-15
Add policy pcr support.


Superhepper:
Currently it is not possible to create any policies(I think). The pull #63 adds some nice improvement to the creation of pcr selections that will be very useful for making it possible to create a policy from pcr values. I suggest that this issue will include the following.

1. Improve the return value of the `pcr_read` api.
2. Create a new api for creating a policy from pcr values (policy_pcr) that can use the improved return value from read_pcr as input.

hug-dev:
Can this now be closed or more work is needed on this topic?

Superhepper:
This can be closed now.

PR #71, opened 2020-05-08
Pcr read return improvement

Superhepper:
I have attempted to make some sense of the return value of pcr_read. So that it will no longer be necessary to manually match the data in the digest with selection.

I also added the method for policy_pcr. All of this pretty much corresponds to #64 

Tell me what you think.

### DIFF
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -22,9 +22,8 @@ use log::error;
 use num_derive::{FromPrimitive, ToPrimitive};
 use num_traits::{FromPrimitive, ToPrimitive};
 use serde::{Deserialize, Serialize};
-use std::collections::HashMap;
+use std::collections::{BTreeMap, HashMap};
 use std::convert::{TryFrom, TryInto};
-
 /// Helper for building `TPM2B_PUBLIC` values out of its subcomponents.
 ///
 /// Currently the implementation is incomplete, focusing on creating objects of RSA type.
@@ -1322,7 +1321,7 @@ pub fn create_unrestricted_signing_ecc_public(
 }
 
 // Enum with the bit flag for each PCR slot.
-#[derive(BitFlags, Hash, Debug, PartialEq, Clone, Copy)]
+#[derive(BitFlags, Hash, Debug, Eq, PartialEq, Ord, PartialOrd, Clone, Copy)]
 #[repr(u32)]
 pub enum PcrSlot {
     Slot0 = 0x0000_0001,
@@ -1352,7 +1351,7 @@ pub enum PcrSlot {
 }
 
 // Enum with the possible values for sizeofSelect.
-#[derive(FromPrimitive, ToPrimitive, Debug, Copy, Clone, PartialEq)]
+#[derive(FromPrimitive, ToPrimitive, Debug, Copy, Clone, PartialEq, Eq)]
 #[repr(u8)]
 pub enum PcrSelectSize {
     OneByte = 1,
@@ -1377,10 +1376,110 @@ impl Default for PcrSelectSize {
 /// number of PCR required by the platform-specific
 /// specification with which the TPM is compliant or by the implementer if
 /// not adhering to a platform-specific specification.
-#[derive(Debug, Default, Clone)]
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
 pub struct PcrSelections {
     size_of_select: PcrSelectSize,
     items: HashMap<HashingAlgorithm, BitFlags<PcrSlot>>,
+}
+
+impl PcrSelections {
+    /// Function for retrieiving the number of banks in the selection
+    pub fn len(&self) -> usize {
+        self.items.len()
+    }
+
+    /// Returns true if the selection is empty.
+    pub fn is_empty(&self) -> bool {
+        self.items.is_empty()
+    }
+
+    /// Removes items in `other` from `self.
+    ///
+    /// # Arguments
+    ///
+    /// * `other` - A PcrSelections containing items
+    ///             that will be removed from `self`.
+    ///     
+    ///
+    /// # Constraints
+    ///
+    /// * Cannot be called with `other` that contains items that
+    ///   are not present in `self`.
+    ///
+    /// * Cannot be called with `other` that has a size_of_select
+    ///   that is different from the one in `self`.
+    ///
+    /// # Errors
+    ///
+    /// * Calling the method with `other` that contains items
+    ///   not present in `self` will result in an InvalidParam error.
+    ///
+    /// * Calling the method with `other` that contains a size_of_select
+    ///   that is different then the one in `self` will result in
+    ///   a InvalidParam error.
+    ///
+    /// # Examples
+    /// ```
+    /// use tss_esapi::utils::{PcrSelectionsBuilder, PcrSlot, algorithm_specifiers::HashingAlgorithm};
+    /// // pcr selections
+    /// let mut pcr_selections = PcrSelectionsBuilder::new()
+    ///     .with_size_of_select(Default::default())
+    ///     .with_selection(HashingAlgorithm::Sha256, &[PcrSlot::Slot0, PcrSlot::Slot8])
+    ///     .build();
+    ///
+    /// // Another pcr selections
+    /// let other = PcrSelectionsBuilder::new()
+    ///     .with_size_of_select(Default::default())
+    ///     .with_selection(
+    ///         HashingAlgorithm::Sha256, &[PcrSlot::Slot0],
+    ///     )
+    ///     .build();
+    /// pcr_selections.subtract(&other).unwrap();
+    /// assert_eq!(pcr_selections.len(), 1);
+    /// ```
+    pub fn subtract(&mut self, other: &Self) -> Result<()> {
+        if self == other {
+            self.items.clear();
+            return Ok(());
+        }
+
+        if self.is_empty() {
+            error!("Error: Trying to remove item that did not exist.");
+            return Err(Error::local_error(WrapperErrorKind::InvalidParam));
+        }
+
+        if self.size_of_select != other.size_of_select {
+            error!("Error: Non matching size of select.");
+            return Err(Error::local_error(WrapperErrorKind::InconsistentParams));
+        }
+
+        for hashing_algorithm in other.items.keys() {
+            // Get selection from difference in order to modify it.
+            let selection = match self.items.get_mut(&hashing_algorithm) {
+                Some(val) => val,
+                None => {
+                    error!("Error: Trying to remove item that did not exist.");
+                    return Err(Error::local_error(WrapperErrorKind::InconsistentParams));
+                }
+            };
+            // Check if value exists in other and if not then nothing needs to be done
+            if let Some(val) = other.items.get(&hashing_algorithm) {
+                if !selection.contains(*val) {
+                    error!("Error: Trying to remove item that did not exist.");
+                    return Err(Error::local_error(WrapperErrorKind::InvalidParam));
+                }
+
+                if selection == val {
+                    // They are equal so everything can be removed.
+                    let _ = self.items.remove(&hashing_algorithm);
+                } else {
+                    // Remove values
+                    selection.remove(*val);
+                }
+            }
+        }
+        Ok(())
+    }
 }
 
 impl From<PcrSelections> for TPML_PCR_SELECTION {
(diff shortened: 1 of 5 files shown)


hug-dev:
Hi @Superhepper 
Do tell us if you would like us to review your PR or if you are still working on it 😃!

Superhepper:
@hug-dev It was under development but it is finished now. I was trying to write a proper example in the documentation of the difference method and that took a little bit extra time and some extra commits.

src/utils/mod.rs
```diff
@@ -1377,12 +1376,115 @@ impl Default for PcrSelectSize {
...
+    ///     .with_size_of_select(Default::default())
+    ///     .with_selection(
+    ///         HashingAlgorithm::Sha256, &[PcrSlot::Slot0],
+    ///     )
+    ///     .build();
+    /// let diff = pcr_selections.difference(&other).unwrap();
+    /// assert_eq!(diff.unwrap().len(), 1);
+    /// ```
+    pub fn difference(&self, other: &Self) -> Result<Option<Self>> {
+        if self == other {
```
hug-dev, line 1442: That would not be needed if the `Option` is dropped: we would end up creating an empty `HashMap`.

src/utils/mod.rs
```diff
@@ -1377,12 +1376,115 @@ impl Default for PcrSelectSize {
...
+                }
+
+                // They are equal so everything can be removed.
+                if selection == val {
+                    let _ = difference.items.remove(&hashing_algorithm);
+                    continue;
+                }
+
+                // Remove values
+                selection.remove(*val);
```
hug-dev, line 1481: I think this is easier to understand:
```rust
// They are equal so everything can be removed.
if selection == val {
    let _ = difference.items.remove(&hashing_algorithm);
} else {
    // Remove values
    selection.remove(*val);
}
```

src/utils/mod.rs
```diff
@@ -1524,3 +1626,176 @@ pub enum PublicKey {
...
+            value: tss_digest.buffer[..size].to_vec(),
+        })
+    }
+}
+
+/// Struct for holding PcrSlots and their
+/// corresponding values.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct PcrBank {
+    bank: BTreeMap<PcrSlot, PcrValue>,
```
hug-dev, line 1661: I am curious of the use of a `BTreeMap` 😃 Is it more efficient than a `HashMap` in this case?

tests/pcr_selections_tests.rs
```diff
@@ -164,4 +169,100 @@ mod test_pcr_selection_builder {
         // The try_from should then fail.
         PcrSelections::try_from(tpml_pcr_selection).unwrap_err();
     }
+
+    #[test]
```
hug-dev, line 173: It would be nice to add a test for `difference` where the two operators have multiple `HashingAlgorithm` keys!

src/utils/mod.rs
```diff
@@ -1377,12 +1376,115 @@ impl Default for PcrSelectSize {
...
+    pub fn difference(&self, other: &Self) -> Result<Option<Self>> {
+        if self == other {
+            return Ok(None);
+        }
+
+        if self.is_empty() {
+            error!("Error: Trying to remove item that did not exist.");
+            return Err(Error::local_error(WrapperErrorKind::InvalidParam));
+        }
+
```
hug-dev, line 1450: That also would not be needed if we iterate through `other.items.keys()`

src/utils/mod.rs
```diff
@@ -1377,12 +1376,115 @@ impl Default for PcrSelectSize {
...
+        self.items.len()
+    }
+
+    /// Returns true if the selection is empty.
+    pub fn is_empty(&self) -> bool {
+        self.items.is_empty()
+    }
+
+    /// Returns the the elements present in `self` but not in `other` or None
+    /// if they are equal.
```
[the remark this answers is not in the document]
ionut-arm: I'd lean towards a `subtract` method to mimick, for example, the [`split_off`](https://doc.rust-lang.org/std/collections/struct.LinkedList.html#method.split_off) in `LinkedList`. Alternatively an iterator that only points to the different elements could be returned. But I agree, it depends on the use case.

src/utils/mod.rs
```diff
@@ -1377,12 +1376,115 @@ impl Default for PcrSelectSize {
...
+    /// # Arguments
+    ///
+    /// * `other` - A PcrSelections containing elements
+    ///             that will be removed from `self`.
+    ///     
+    ///
+    /// # Constraints
+    ///
+    /// * Cannot be called with `other` that contains elements that
+    ///   is not present in `self`.
```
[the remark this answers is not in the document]
ionut-arm: I suggest using [`other.items.iter`](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.iter) - then you don't need to do a `get` on the key value.

src/utils/mod.rs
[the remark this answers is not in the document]
Superhepper: The use-case for the method it is to get the remaining items when you do I pcr_read for example.

In the case  that the you have a PcrSelections that contains more then 16 items then it will not be able to get all of them in one read. So I was thinking you sould be able to do something like
```rust
let mut remaining = Some(pcr_selections);
while remaining.is_some() {
   let (_update_counter, read_pcr_selections, data) = ctx.pcr_read(remaining.unwrap());
   remaining = remaining.map(|x| x.difference(read_pcr_selections);
}
```

But in order to be able to do that properly. There should probably be some kind of merge funtion for the PcrData struct.

src/utils/mod.rs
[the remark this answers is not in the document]
Superhepper: Yeah that looked super weird. I do not really know what I was thinking.

src/utils/mod.rs
Superhepper: I kind of like that the function is returning an Option. Either there is a Some difference or there is None. But that is matter of taste and I think both works.

src/utils/mod.rs
Superhepper: Yeah, I think you are absolutely right.

src/utils/mod.rs
Superhepper: No the reason for the BTreeMap here is because it is ordered. And if I am not mistaken when I read the specification is that the returned values from pcr_read are also in order. So the lowest selected slot number for a specific bank will end up first in the returned digest.

When I did this I did not take performance into account. I was more thinking that for the banks I did not care about ordering (= HashMap) for the pcr slot values I did care about the order because it made sense there (=BTreeMap).
If we are concerned about performance here then the HashMap needs to be removed. Because it will probably be slower for a small amount of items.

src/utils/mod.rs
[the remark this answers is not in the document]
hug-dev: Could this:
```rust
remaining = remaining.map(|x| x.difference(read_pcr_selections));
```
be replaced with something like (as an idea)
```rust
remaining.substract(read_pcr_selections);
```

src/utils/mod.rs
hug-dev: Sure, if you prefer having an `Option` there, just go with it, that is totally fine with me 😃

src/utils/mod.rs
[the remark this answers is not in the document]
Superhepper: yeah absolutly something like:
```rust
let mut remaining = pcr_selections;
while !remaining.is_empty() {
    let (update_counter, read_pcr_selections, pcr_data) = ctx.read_pcr(remaining).unwrap();
    remaining.subtract(read_pcr_selections);
}
```
I think that works just as well.

src/utils/mod.rs
Superhepper: But if its going to change be a subtract then the option is not needed.

src/utils/mod.rs
hug-dev: Thanks for the explanation, I am good with it!

src/utils/mod.rs
Superhepper: You are right it will be caught in the loop though I think I would like to keep it as an "early out".

src/utils/mod.rs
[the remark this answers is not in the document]
Superhepper: Fixed

src/utils/mod.rs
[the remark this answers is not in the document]
Superhepper: Fixed

src/utils/mod.rs
Superhepper: Option is removed

src/utils/mod.rs
Superhepper: Fixed

tests/pcr_selections_tests.rs
Superhepper: fixed

Superhepper:
@hug-dev @ionut-arm Do any of you understand why the CI failed?

hug-dev:
It seems that one of our dependency either does not compile on nightly or on version 1.38.
However I tried locally on your branch and did not see the bug 🤯 I will try to restart the CI.

Superhepper:
Nice, everything seams to work now.

tests/pcr_selections_tests.rs
```diff
@@ -1,167 +1,158 @@
...
 
-        // The try_from should then fail.
-        PcrSelections::try_from(tpml_pcr_selection).unwrap_err();
+        pcr_selections.subtract(&pcr_selections_1).unwrap_err();
     }
 
     #[test]
-    fn test_conversion_of_data_with_invalid_hash_alg_id() {
-        let mut tpml_pcr_selection: TPML_PCR_SELECTION = PcrSelectionsBuilder::new()
+    fn test_subtract_with_larg_selection_ramining_value() {
```
hug-dev, line 98: Nice test 👏

src/utils/mod.rs
```diff
@@ -1377,12 +1376,112 @@ impl Default for PcrSelectSize {
...
+    /// let other = PcrSelectionsBuilder::new()
+    ///     .with_size_of_select(Default::default())
+    ///     .with_selection(
+    ///         HashingAlgorithm::Sha256, &[PcrSlot::Slot0],
+    ///     )
+    ///     .build();
+    /// pcr_selections.subtract(&other).unwrap();
+    /// assert_eq!(pcr_selections.len(), 1);
+    /// ```
+    pub fn subtract(&mut self, other: &Self) -> Result<()> {
```
hug-dev, line 1440: Thanks, I prefer this version!

genofire:
I am missing to get `TPML_DIGEST` again (by e.g. `std::convert::into`).

Superhepper:
> I am missing to get `TPML_DIGEST` again (by e.g. `std::convert::into`).

What is it you want to to do? 

If we are talking about the return value from pcr_read (i.e. PcrData)? Then no not really, it cannot be converted into anything if you do not do it manually. You can extract all the values from that structure rather easily because you can look them up (everything is mapped to HashingAlgorithm and PcrSlot).

p.s.
It would be better to move this discussion to either the slack channel or create an issue for what you think is missing or want to accomplish. It is hard for people to find things i comment to a pull request that have been closed.

genofire:
> > I am missing to get `TPML_DIGEST` again (by e.g. `std::convert::into`).
> 
> What is it you want to to do?
I need the `TPML_DIGEST`  after a `pcr_read` in my "strange" use case.
Compared to `PcrSelections` there exists a `impl core::convert::From<PcrSelections>`
```
fn from(pcr_selections: PcrSelections) -> TPML_PCR_SELECTION
```

current i have to use this legacy version for it: 1438bacf38a1d52eed549af7f8709318cb7e296e
Maybe i will late create a full issue or maybe a PR, current i just wanted a feedback  from developer of the PR / creator of `PcrData` (maybe i do not get everthing).

> p.s.
> It would be better to move this discussion to either the slack channel or create an issue for what you think is missing or want to accomplish. It is hard for people to find things i comment to a pull request that have been closed.
Sorry but i do not like slack, it is zero privacy awareness and so on ...

Superhepper:
> > > I am missing to get `TPML_DIGEST` again (by e.g. `std::convert::into`).
> > 
> > 
> > What is it you want to to do?
> 
> I need the `TPML_DIGEST` after a `pcr_read` in my "strange" use case.
> Compared to `PcrSelections` there exists a `impl core::convert::From<PcrSelections>`
> 
> ```
> fn from(pcr_selections: PcrSelections) -> TPML_PCR_SELECTION
> ```
> 
> current i have to use this legacy version for it: [1438bac](https://github.com/parallaxsecond/rust-tss-esapi/commit/1438bacf38a1d52eed549af7f8709318cb7e296e)
> Maybe i will late create a full issue or maybe a PR, current i just wanted a feedback from developer of the PR / creator of `PcrData` (maybe i do not get everything).
> 
> > p.s.
> > It would be better to move this discussion to either the slack channel or create an issue for what you think is missing or want to accomplish. It is hard for people to find things i comment to a pull request that have been closed.
> 
> Sorry but i do not like slack, it is zero privacy awareness and so on ..

I am sorry. My main goal when I did the improvement of return value for pcr_read was to get away from TPML_PCR_SELECTION and TPML_DIGEST and how they related to each other because to me it appeared to a thing that could easily be done wrong.

If you just want that data in the form of TPML_DIGEST and you are not interested that it correctly relates to the original selection you passed into the pcr_read then you could probably do something like this (OBS not tested):

```rust
let mut tpml_digest: TPML_DIGEST = Default::default();

for pcr_bank in pcr_data {
    for (_pcr_slot, pcr_value) in pcr_bank {
        let mut& tpmb_digest = &tpml_digest.digests[tpml_digest.count as usize];
        tpmb_digest.size = pcr_value.value().len() as u16;
        tpmb_digest.buffer.copy_from_slize(pcr_value.value());
        tpml_digest.count = tpml_digest.count + 1;
    }
}
```

genofire:
thx @Superhepper for the idea - i implemented it in #106

PR #72, opened 2020-05-11
Hash method

Superhepper:
I added context method for the Esys_Hash method. The return value from that method is a digest and a TPMT_TK_HASHCHECK ticket. So in order to create HashcheckTicket type I had to create an enum representing StructureTag(TPM_ST).

### DIFF
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -9,6 +9,8 @@
 //! type name. Unions are converted to Rust `enum`s by dropping the `TPMU` qualifier and appending
 //! `Union`.
 pub mod algorithm_specifiers;
+pub mod tags;
+pub mod tickets;
 
 use crate::constants::*;
 use crate::response_code::{Error, Result, WrapperErrorKind};
@@ -865,7 +867,7 @@ pub struct Signature {
 #[derive(Debug, PartialEq)]
 pub enum SignatureData {
     RsaSignature(Vec<u8>),
-    EccSignature { r: Vec<u8>, s: Vec<u8> },
+    EcdsaSignature { r: Vec<u8>, s: Vec<u8> },
 }
 
 impl Signature {
@@ -933,7 +935,7 @@ impl Signature {
 
                 Ok(Signature {
                     scheme,
-                    signature: SignatureData::EccSignature { r, s },
+                    signature: SignatureData::EcdsaSignature { r, s },
                 })
             }
             TPM2_ALG_SM2 | TPM2_ALG_ECSCHNORR | TPM2_ALG_ECDAA => {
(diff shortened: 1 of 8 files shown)


src/utils/tickets.rs
```diff
@@ -0,0 +1,110 @@
...
+                    }
+                    Err(why) => {
+                        error!("Failed to parsed tag: {}", why);
+                        return Err(Error::local_error(WrapperErrorKind::InvalidParam));
+                    }
+                };
+
+                let len = tss_ticket.digest.size.into();
+                let mut digest = tss_ticket.digest.buffer.to_vec();
+                digest.truncate(len);
```
ionut-arm, line 51: Reckon it should be an error if `len` is longer than the buffer size? It `truncate` won't fail in any way, just asking from a programming experience point of view

src/utils/mod.rs
```diff
@@ -865,7 +867,7 @@ pub struct Signature {
 #[derive(Debug, PartialEq)]
 pub enum SignatureData {
     RsaSignature(Vec<u8>),
-    EccSignature { r: Vec<u8>, s: Vec<u8> },
+    EcdsaSignature { r: Vec<u8>, s: Vec<u8> },
```
ionut-arm, line 870: These seem to be remnants from a change I made (and since merged) that you might've merged into your branch

src/utils/mod.rs
```diff
@@ -9,6 +9,8 @@
 //! type name. Unions are converted to Rust `enum`s by dropping the `TPMU` qualifier and appending
 //! `Union`.
 pub mod algorithm_specifiers;
+pub mod tags;
+pub mod tickets;
```
ionut-arm, line 13: 💯 nice, I think we'd do well to start splitting the rest of this file out, it's a mess of random types and utils for different things at the moment.

tests/context_tests.rs
```diff
@@ -1234,3 +1235,30 @@ mod test_test_parms {
...
+                data.as_bytes(),
+                HashingAlgorithm::Sha256,
+                expected_hierarchy,
+            )
+            .unwrap();
+        assert_eq!(expected_hashed_data.len(), actual_hashed_data.len());
+        assert_eq!(&expected_hashed_data[..], &actual_hashed_data[..]);
+        assert_eq!(ticket.hierarchy(), expected_hierarchy);
+        assert_ne!(ticket.digest().len(), 0); // Should do some better checking of the digest
+    }
```
ionut-arm, line 1263: 👍

src/utils/tickets.rs
Superhepper: Fixed

PR #74, opened 2020-05-29
Also add the TSS2-MU library

puiterwijk:
This also makes the Marshalling/Unmarshalling API available via tss2_esys.
This makes it possible to use those functions with the tss2_esys types.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -16,8 +16,12 @@ fn main() {
             .atleast_version(MINIMUM_VERSION)
             .probe("tss2-tctildr")
             .expect("Error with pkg-config finding tss2-tctildr.");
+        let tss2_mu = pkg_config::Config::new()
+            .atleast_version(MINIMUM_VERSION)
+            .probe("tss2-mu")
+            .expect("Error with pkg-config finding tss2-mu.");
 
-        // These two pkg-config files should contain only one include/lib path.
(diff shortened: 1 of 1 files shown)


hug-dev:
> Also, just as a note for the future, we have a plan to decouple the bindgen-generated types from the Rust-native abstractions into a *-sys crate.

Just created #75 for that work.

puiterwijk:
@ionut-arm Thanks! My use is for compatibility with keys created externally, which store the TPM2B_PUBLIC structures as their Marshalled versions on disk.
This allows me to unmarshal those objects as `tss_esapi::tss2_esys::TPM2B_PUBLIC` types for use in e.g. `Context.load`

puiterwijk:
I might at some point also add a native Rust wrapper (possibly in the `utils` module), but for now I'm first trying to get some plumbing for stuff I'm working on upstreamed in small patches.

ionut-arm:
> 
> 
> @ionut-arm Thanks! My use is for compatibility with keys created externally, which store the TPM2B_PUBLIC structures as their Marshalled versions on disk.
> This allows me to unmarshal those objects as `tss_esapi::tss2_esys::TPM2B_PUBLIC` types for use in e.g. `Context.load`

That sounds interesting, sounds like something that could be integrated with `serde` for Rust-native serialization support!

PR #78, opened 2020-06-08
Implement std::error::Error on Error

### LINKED ISSUE #77: the trait `std::error::Error` is not implemented for `tss_esapi::response_code::Error
An implementation for the std::error::Error should be available for our Error type. Not having that makes it complicated to wrap or box our errors.

ionut-arm:
This commit implements the native Error trait on the Error hierarchy of
types.

@Superhepper 
Fixes #77

### DIFF
--- a/src/response_code.rs
+++ b/src/response_code.rs
@@ -28,6 +28,15 @@ bitfield! {
     severity, _: 11;
 }
 
+impl std::fmt::Display for FormatZeroResponseCode {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        // Display response code as "Response code value: 0xdeadbeef"
+        write!(f, "Response code value: {:#x}", self.0)
+    }
+}
+
+impl std::error::Error for FormatZeroResponseCode {}
+
 bitfield! {
     #[derive(PartialEq, Copy, Clone)]
     pub struct FormatOneResponseCode(TSS2_RC);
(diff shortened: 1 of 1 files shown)


src/response_code.rs
```diff
@@ -439,6 +457,16 @@ impl From<TSS2_RC> for Tss2ResponseCode {
     }
 }
 
+impl std::error::Error for Tss2ResponseCode {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        match self {
+            Tss2ResponseCode::FormatOne(response_code) => Some(response_code),
```
Superhepper, line 463: Shouldn't it be 
```rust
Tss2ResponseCode::FormatOne(ref response_code) => Some(response_code),
Tss2ResponseCode::FormatZero(ref response_code) => Some(response_code),
```

Or doesn't it matter?

src/response_code.rs
```diff
@@ -476,6 +504,15 @@ impl std::fmt::Display for Error {
     }
 }
 
+impl std::error::Error for Error {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        match self {
+            Error::WrapperError(wrapper_error) => Some(wrapper_error),
```
Superhepper, line 510: Same thing here ...
```rust
Error::WrapperError(ref wrapper_error) => Some(wrapper_error),
Error::Tss2Error(ref response_code) => Some(response_code),
```

But again I am not sure it is necessary. Maybe it is implicitly converted to a reference due to the way the return type is declared?

ionut-arm: I think it's implicit because the function takes in `&self`, so `match self` actually returns `Error::WrapperError(&WrapperErrorKind)`, not an actual object!

Superhepper: Yeah makes sense. In that case I have no more questions. Nice job.

PR #79, opened 2020-06-09
Minor change to remove usage of raw TSS type in load_external API.

Superhepper:


Superhepper:
Hehe, I am using tss 2.4.1(>= 2.4.0, have breaking changes) and they have changed the type of the Hierarchy so when the tests succeeds for me locally they fail here in CI. So it is a little bit of 'trial and horror'.

PR #84, opened 2020-06-20
Updated tpm2-tss version to 2.4.1 and tpm2-tools version to 4.2.1

Superhepper:

### DIFF
--- a/tests/Dockerfile
+++ b/tests/Dockerfile
@@ -3,7 +3,7 @@ FROM tpm2software/tpm2-tss:ubuntu-18.04
 ENV PKG_CONFIG_PATH /usr/local/lib/pkgconfig
 
 # Download and install TSS 2.0
-RUN git clone https://github.com/tpm2-software/tpm2-tss.git --branch 2.3.3
+RUN git clone https://github.com/tpm2-software/tpm2-tss.git --branch 2.4.1
 RUN cd tpm2-tss \
 	&& ./bootstrap \
 	&& ./configure \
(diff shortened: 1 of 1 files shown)


ionut-arm:
> 
> 
> It's good that it works! If it is non-breaking, we do not even have to change the `MINIMUM_VERSION` variable in `build.rs`.

I'm wondering, though. We should maybe have two builds with two different Dockerfiles, one with the minimum change and one with the maximum. Otherwise we might introduce changes that break the minimum compatibility and we wouldn't know

hug-dev:
> we wouldn't know

I think the problem is there! Ideally we should not export any type from the ones generated by `bindgen` or if we do, manually and carefully select them. That way the external interface of this crate won't change.
That might be a problem though because I believe a lot of people using this crate directly use the `bindgen` type.

ionut-arm:
I wasn't actually talking about interface compatibility, I was more thinking about bugs or changes in the library implementation that might not work with our implementations of the `TransientKeyContext`, for example (we chose 2.3.0 minimum because it included some changes that we needed for our code to actually work).

Superhepper:
I actually like that. Two builds one with maximum and one with minimum (within the same major version of the release of course).

And at some point 3.0.x will go from being just dev to becoming a release and then we probably need to handle that in some nice way. Depending on when that happen we have perhaps not got as far as we would have liked and it will be a non issue to just set minimum to 3.0.x at that point.

hug-dev:
Please feel free to move the PRs that you want us to actively review as Ready! I am not sure what was the state on this one.

Superhepper:
This one could probably be removed. It was merely to test that it worked. I liked @ionut-arm suggestion and think we should go with that one instead.

PR #88, opened 2020-07-01
Add TCTI configuration

ionut-arm:
This commit adds functionality for the TCTI to be configured when
creating a `Context`.

### DIFF
--- a/src/utils/tcti.rs
+++ b/src/utils/tcti.rs
@@ -1,0 +1,450 @@
+// Copyright 2020 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+use crate::response_code::{Error, WrapperErrorKind};
+use regex::Regex;
+use std::convert::TryFrom;
+use std::ffi::CString;
+use std::net::IpAddr;
+use std::path::PathBuf;
+use std::str::FromStr;
+
+const DEVICE: &str = "device";
+const MSSIM: &str = "mssim";
+const TABRMD: &str = "tabrmd";
+
+// Possible TCTI to use with the ESYS API.
+/// Placeholder TCTI types that can be used when initialising a `Context` to determine which
+/// interface will be used to communicate with the TPM.
+#[derive(Clone, Debug, PartialEq)]
+pub enum Tcti {
+    /// Connect to a TPM available as a device node on the system
+    ///
+    /// For more information about configuration, see [this page](https://www.mankier.com/3/Tss2_Tcti_Device_Init)
+    Device(DeviceConfig),
+    /// Connect to a TPM (simulator) available as a network device
+    ///
+    /// For more information about configuration, see [this page](https://www.mankier.com/3/Tss2_Tcti_Mssim_Init)
+    Mssim(MssimConfig),
+    /// Connect to a TPM through an Access Broker/Resource Manager daemon
+    ///
+    /// For more information about configuration, see [this page](https://www.mankier.com/3/Tss2_Tcti_Tabrmd_Init)
+    Tabrmd(TabrmdConfig),
+}
+
+impl TryFrom<Tcti> for CString {
+    type Error = Error;
+
+    fn try_from(tcti: Tcti) -> Result<Self, Error> {
+        let tcti_name = match tcti {
+            Tcti::Device(..) => DEVICE,
+            Tcti::Mssim(..) => MSSIM,
+            Tcti::Tabrmd(..) => TABRMD,
+        };
+
+        let tcti_conf = match tcti {
+            Tcti::Mssim(config) => {
+                if let ServerAddress::Hostname(name) = &config.host {
+                    if !hostname_validator::is_valid(name) {
+                        return Err(Error::WrapperError(WrapperErrorKind::InvalidParam));
+                    }
+                }
+                format!("host={},port={}", config.host, config.port)
+            }
+            Tcti::Device(DeviceConfig { path }) => path
+                .to_str()
+                .ok_or(Error::WrapperError(WrapperErrorKind::InvalidParam))?
+                .to_owned(),
+            Tcti::Tabrmd(config) => {
+                format!("bus_name={},bus_type={}", config.bus_name, config.bus_type)
+            }
+        };
+
+        if tcti_conf.is_empty() {
+            CString::new(tcti_name).or(Err(Error::WrapperError(WrapperErrorKind::InvalidParam)))
+        } else {
+            CString::new(format!("{}:{}", tcti_name, tcti_conf))
+                .or(Err(Error::WrapperError(WrapperErrorKind::InvalidParam)))
+        }
+    }
+}
+
+impl FromStr for Tcti {
+    type Err = Error;
+
+    fn from_str(config_str: &str) -> Result<Self, Self::Err> {
+        let device_pattern = Regex::new(r"^device(:(.*))?$").unwrap(); //should not fail
+        if let Some(captures) = device_pattern.captures(config_str) {
+            return Ok(Tcti::Device(DeviceConfig::from_str(
+                captures.get(2).map_or("", |m| m.as_str()),
+            )?));
+        }
+
+        let mssim_pattern = Regex::new(r"^mssim(:(.*))?$").unwrap(); //should not fail
+        if let Some(captures) = mssim_pattern.captures(config_str) {
+            return Ok(Tcti::Mssim(MssimConfig::from_str(
+                captures.get(2).map_or("", |m| m.as_str()),
+            )?));
+        }
+
+        let tabrmd_pattern = Regex::new(r"^tabrmd(:(.*))?$").unwrap(); //should not fail
+        if let Some(captures) = tabrmd_pattern.captures(config_str) {
+            return Ok(Tcti::Tabrmd(TabrmdConfig::from_str(
+                captures.get(2).map_or("", |m| m.as_str()),
+            )?));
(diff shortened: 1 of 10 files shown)


src/utils/tcti.rs
```diff
@@ -0,0 +1,150 @@
...
+}
+
+/// Configuration for a Device TCTI context
+///
+/// The default configuration lets the library use its default path.
+#[derive(Clone, Debug, Default)]
+pub struct DeviceConfig {
+    /// Path to the device node to connect to
+    ///
+    /// If set to `None`, the default location is used
```
hug-dev, line 73: It would be nice to list what default locations are used, if you can find it somewhere standard. When browsing the code I think I saw that `/dev/tpmrm0` was tested first, the `/dev/tpm0`

src/utils/tcti.rs
```diff
@@ -0,0 +1,150 @@
+// Copyright 2019 Contributors to the Parsec project.
```
hug-dev, line 1: 2020 I think 😬

src/utils/tcti.rs
ionut-arm: I wanted to say that but I had no idea where they are on Windows! Will check

src/utils/tcti.rs
```diff
@@ -0,0 +1,451 @@
...
+    type Err = Error;
+
+    fn from_str(config_str: &str) -> Result<Self, Self::Err> {
+        let device_pattern = Regex::new(r"^device(:(.*))?$").unwrap(); //should not fail
+        if device_pattern.is_match(config_str) {
+            match device_pattern.captures(config_str) {
+                None => unreachable!("Branch can never be reached"),
+                Some(captures) => {
+                    return Ok(Tcti::Device(DeviceConfig::from_str(
+                        captures.get(2).map_or("", |m| m.as_str()),
```
hug-dev, line 87: What does the `map_or` handle?

src/utils/tcti.rs
```diff
@@ -0,0 +1,451 @@
...
+}
+
+impl FromStr for Tcti {
+    type Err = Error;
+
+    fn from_str(config_str: &str) -> Result<Self, Self::Err> {
+        let device_pattern = Regex::new(r"^device(:(.*))?$").unwrap(); //should not fail
+        if device_pattern.is_match(config_str) {
+            match device_pattern.captures(config_str) {
+                None => unreachable!("Branch can never be reached"),
```
hug-dev, line 84: Why is it unreachable here? What if `config_str` is only `"device"`?

src/utils/tcti.rs
```diff
@@ -0,0 +1,451 @@
...
+                }
+            }
+        }
+
+        Err(Error::WrapperError(WrapperErrorKind::InvalidParam))
+    }
+}
+
+#[test]
+fn validate_from_str_tcti() {
```
hug-dev, line 122: Nice, thanks for the tests!

src/utils/tcti.rs
```diff
@@ -0,0 +1,451 @@
...
+        let host = match host_pattern.captures(config_str) {
+            None => None,
+            Some(captures) => Some(ServerAddress::from_str(captures.get(2).unwrap().as_str())?),
+        };
+
+        let port_pattern = Regex::new(r"(,|^)port=(.*?)(,|$)").unwrap(); // should not fail
+        let port = match port_pattern.captures(config_str) {
+            None => None,
+            Some(captures) => Some(
+                u16::from_str(captures.get(2).unwrap().as_str())
```
hug-dev, line 240: Is the `unwrap` always ok?

src/utils/tcti.rs
```diff
@@ -0,0 +1,451 @@
...
+impl FromStr for MssimConfig {
+    type Err = Error;
+
+    fn from_str(config_str: &str) -> Result<Self, Self::Err> {
+        if config_str.is_empty() {
+            return Ok(Default::default());
+        }
+        let host_pattern = Regex::new(r"(,|^)host=(.*?)(,|$)").unwrap(); // should not fail
+        let host = match host_pattern.captures(config_str) {
+            None => None,
```
hug-dev, line 232: Just an idea: instead of having `Option` in the configuration, we could always put the default value instead of `None`.

src/utils/tcti.rs
ionut-arm: It handles the case where that 2nd capture group is empty (in which case the get would return a `None`), when it returns "" instead (what we want => default config).

src/utils/tcti.rs
ionut-arm: It's unreachable because I check two lines above that there is a match - even if `config_str` is only "device", there is at least one capture group, which is the whole matched string (in this case "device").

src/utils/tcti.rs
ionut-arm: No, actually it isn't - I updated it locally but didn't feel like pushing again. It's essentially the same fix as for the first comment you posted. I'll add a test with something like `host=,port=1234`, that would've caught it.

src/utils/tcti.rs
ionut-arm: We could, but what if a user wants to set one value to something and the other leave it as the default? You can't implement `Default` on `u16`, for example, so you have to set them both manually when creating the object. I guess you could initialise with `Default::default()` and then change the one you want to change, but it feels nicer to say "no, I only want to set this one, the other doesn't matter"

src/utils/tcti.rs
ionut-arm: (I also updated all the other places that do it like this)

src/utils/tcti.rs
```diff
@@ -0,0 +1,451 @@
...
+            Tcti::Tabrmd(config) => format!(
+                ":bus_name={},bus_type={}",
+                config
+                    .bus_name
+                    .unwrap_or_else(|| String::from("com.intel.tss2.Tabrmd")),
+                config.bus_type.unwrap_or(BusType::System)
+            ),
+        };
+
+        CString::new(format!("{}{}", tcti_name, tcti_conf))
```
Superhepper, line 72: Sorry for been 'that guy' here but. I think it would be clearer if you left out the limiter when you create tcti_conf and add it here in the format instead.
```rust
CString::new(format!("{}:{}", tcti_name, tcti_conf))
```

src/utils/tcti.rs
Superhepper: Would it be some kind of performance hit to do:
```rust
if let Some(captures) = device_pattern.captures(config_str) {
    return Ok(Tcti::Device(DeviceConfig::from_str(
                     captures.get(2).map_or("", |m| m.as_str()),
                     )?));
}
```
Because I think that would look nicer.

src/utils/tcti.rs
Superhepper: Nice!

src/utils/tcti.rs
```diff
@@ -0,0 +1,451 @@
...
+                    Regex::new(r"^[a-zA-Z0-9\-_]+(\.[a-zA-Z0-9\-_]+)+$").unwrap(); //should not fail
+                if !valid_bus_name_pattern.is_match(captures.get(2).unwrap().as_str()) {
+                    return Err(Error::WrapperError(WrapperErrorKind::InvalidParam));
+                }
+                Some(captures.get(2).unwrap().as_str().to_owned())
+            }
+        };
+
+        let bus_type_pattern = Regex::new(r"(,|^)bus_type=(.*?)(,|$)").unwrap(); // should not fail
+        let bus_type = match bus_type_pattern.captures(config_str) {
```
Superhepper, line 378: Could you not use .map() instead?

src/utils/tcti.rs
hug-dev: I think that if you implement `Default` on the different config structures, then an user can do:

```rust
let mssim_config = MssimConfig {
    port: 1234,
    ..Default::default()
};
```

but as I said, this looks fine as it is now and is not necessary.

src/utils/tcti.rs
ionut-arm: The problem with that is when `tcti_conf` is empty. I could branch before, I guess

src/utils/tcti.rs
ionut-arm: Nice, I didn't know that works, thought `Default::default()` would set the type-specific default values, not those defined for the container type.

src/utils/tcti.rs
ionut-arm: That sounds good!

src/utils/tcti.rs
```diff
@@ -0,0 +1,450 @@
...
+    path: PathBuf,
+}
+
+impl Default for DeviceConfig {
+    fn default() -> Self {
+        DeviceConfig {
+            path: PathBuf::from("/dev/tpm0"),
+        }
+    }
+}
```
hug-dev, line 162: 🆗

ionut-arm: I'm legit afraid that would need to be changed to be OS-dependent

src/utils/tcti.rs
ionut-arm: Done

src/utils/tcti.rs
ionut-arm: Done

src/utils/tcti.rs
Superhepper: I share your concern but I think it is ok for now.

PR #89, opened 2020-07-01
Add unsealing

puiterwijk:

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -129,7 +129,8 @@ use std::ptr::{null, null_mut};
 use tss2_esys::*;
 use utils::{
     algorithm_specifiers::HashingAlgorithm, tickets::HashcheckTicket, Hierarchy, PcrData,
-    PcrSelections, PublicParmsUnion, Signature, TpmaSession, TpmaSessionBuilder, TpmsContext,
+    PcrSelections, PublicParmsUnion, SensitiveData, Signature, TpmaSession, TpmaSessionBuilder,
+    TpmsContext,
 };
 
 #[macro_use]
@@ -489,6 +490,31 @@ impl Context {
             Ok((*outprivate, *outpublic))
         } else {
             error!("Error in creating derived key: {}.", ret);
+            Err(ret)
+        }
+    }
+
+    /// Unseal and return data from a Sealed Data Object
+    pub fn unseal(&mut self, item_handle: ESYS_TR) -> Result<SensitiveData> {
+        let mut out_data = null_mut();
+
+        let ret = unsafe {
+            Esys_Unseal(
+                self.mut_context(),
+                item_handle,
+                self.sessions.0,
+                self.sessions.1,
+                self.sessions.2,
+                &mut out_data,
+            )
+        };
+        let ret = Error::from_tss_rc(ret);
+
+        if ret.is_success() {
+            let out_data = unsafe { MBox::from_raw(out_data) };
+            Ok(SensitiveData::try_from(*out_data)?)
+        } else {
+            error!("Error in unsealing: {}.", ret);
             Err(ret)
         }
     }
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -853,6 +853,54 @@ impl AsymSchemeUnion {
     }
 }
 
+/// Rust native representation of sensitive data.
+///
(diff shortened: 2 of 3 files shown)


src/lib.rs
```diff
@@ -493,6 +493,31 @@ impl Context {
...
+                self.sessions.1,
+                self.sessions.2,
+                &mut out_data,
+            )
+        };
+        let ret = Error::from_tss_rc(ret);
+
+        if ret.is_success() {
+            let out_data = unsafe { MBox::from_raw(out_data) };
+            Ok(TPM2B_SENSITIVE_DATA::try_from(*out_data).unwrap())
```
ionut-arm, line 514: Instead of `TPM2B_SENSITIVE_DATA::try_from(*out_data).unwrap()`, simply doing `*out_data` also works here with the same result!

src/lib.rs
```diff
@@ -493,6 +493,31 @@ impl Context {
         }
     }
 
+    /// Unseal and return data from a Sealed Data Object
+    pub fn unseal(&mut self, item_handle: ESYS_TR) -> Result<TPM2B_SENSITIVE_DATA> {
```
Superhepper, line 497: I would really like to see that the TPM2B_SENSITIVE_DATA is not used as a return value if not necessary. I would like it better if there was a simple struct containing the data and then there was conversion functions.

```rust
pub struct SensitiveData {
    sensitive_data: Vec<u8>,
}

impl SesitiveData {
    pub fn new() -> Self {
        SensitiveData { sesitive_data: Vec::<u8>::new() }
    }

    pub fn data(&mut self) -> &[u8] {
        self.sensitive_data
    }
}

impl TryFrom<TPM2B_SENSITIVE_DATA> for SesitiveData {
    fn try_from(tpm2b_sesitive_data: TPM2B_SENSITIVE_DATA) -> Result<Self> {
        if tpm2b_sesnsitive_data.size > 255 {
             // return some error
        }
        Ok(SesitiveData { sensitive_data: tpm2b_sensitive_data.buffer[..tpm2b_sesnsitive_data.size as usize].to_vec() } )
    }
}

impl TryFrom<SensitiveData> for TPM2B_SENSITIVE_DATA {
    fn try_from(sensitive_data: SensitiveData) -> Result<Self> {
        if sensitive_data.data().len() > 255 {
              // return an error
        }
        let mut rv: TPM2B_SENSITIVE_DATA = Default::default();
        rv.size = sensitive_data.len() as u16;
        rv.buffer[..rv.size as usize].copy_from_slice(sensitive_data.data());
        Ok(rv)
    }
}
```

Then the return of the unseal method should be
```rust
Result<SensitiveData>
```

And the method should end with
```rust

Ok(SensitiveData::try_from(out_data)?)

Comment (puiterwijk) [reply src/lib.rs:497]: Well, the reason I went for a `TPM2B_SENSITIVE` return is that other functions also return the inner `TPM2B_` types.
Example: `pub fn read_public(&mut self, key_handle: ESYS_TR) -> Result<TPM2B_PUBLIC> {`.
And in other cases, there's functions that accept `TPM2B_SENSITIVE`, like:
```
pub fn load_external(
        &mut self,
        private: &TPM2B_SENSITIVE,
        public: &TPM2B_PUBLIC,
        hierarchy: Hierarchy,
    ) -> Result<ESYS_TR> {
```

And in order to be compatible with those, and the tpm2-mu library, I think that keeping the `TPM2B_SENSITIVE` would be most consistent.

puiterwijk: (Note: I'm not opposed, just trying to explain why I went with the inner structure. If the goal is to move the other uses of `TPM2B_SENSITIVE` over to another structure type as well, I personally think that should all be changed in one time.)

puiterwijk: I just realized that this is `TPM2B_SENSITIVE_DATA` and that there's no other functions using that version.
So I guess then it'll be fine, and I will take your suggestion (thanks!) and update my PR.

hug-dev: Thanks for creating the abstraction structure as suggested by @Superhepper ! A goal that we have for this crate is to eventually abstract all the ESYS types generated by bindgen into our own native, idiomatic and friendly types 😃

puiterwijk: @hug-dev Good to know, thanks! Then I'll make sure to do this as much as reasonable for my own changes in the future.

PR #90, opened 2020-07-02
Added policy_or

puiterwijk:

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -128,9 +128,9 @@ use std::ffi::CString;
 use std::ptr::{null, null_mut};
 use tss2_esys::*;
 use utils::{
-    algorithm_specifiers::HashingAlgorithm, tickets::HashcheckTicket, Hierarchy, PcrData,
-    PcrSelections, PublicParmsUnion, SensitiveData, Signature, TpmaSession, TpmaSessionBuilder,
-    TpmsContext,
+    algorithm_specifiers::HashingAlgorithm, tickets::HashcheckTicket, Digest, DigestList,
+    Hierarchy, PcrData, PcrSelections, PublicParmsUnion, SensitiveData, Signature, TpmaSession,
+    TpmaSessionBuilder, TpmsContext,
 };
 
 #[macro_use]
@@ -921,6 +921,39 @@ impl Context {
         }
     }
 
+    /// Cause conditional gating of a policy based on an OR'd condition.
+    ///
+    /// The TPM will ensure that the current policy digest equals at least
+    /// one of the digests.
+    /// If this is the case, the policyDigest of the policy session is replaced
+    /// by the value of the different hashes.
+    ///
+    /// # Constraints
+    /// * `hash_list` must be at least 2 and at most 8 elements long
+    ///
+    /// # Errors
+    /// * if the hash list provided is too short or too long, a `WrongParamSize` wrapper error will be returned
+    pub fn policy_or(&mut self, policy_session: ESYS_TR, digest_list: DigestList) -> Result<()> {
+        let digest_list = TPML_DIGEST::try_from(digest_list)?;
+
+        let ret = unsafe {
+            Esys_PolicyOR(
+                self.mut_context(),
+                policy_session,
+                self.sessions.0,
+                self.sessions.1,
+                self.sessions.2,
+                &digest_list,
+            )
+        };
+        let ret = Error::from_tss_rc(ret);
+        if ret.is_success() {
+            Ok(())
+        } else {
+            Err(ret)
+        }
+    }
+
     // TODO: Should we really keep `num_bytes` as `u16`?
     /// Get a number of random bytes from the TPM and return them.
     ///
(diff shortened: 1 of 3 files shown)


puiterwijk:
Note that this also changes the result type of `policy_get_digest` to a `Result<Digest>`, to hide the `TPM2B_DIGEST` type.

Superhepper:
> Note that this also changes the result type of `policy_get_digest` to a `Result<Digest>`, to hide the `TPM2B_DIGEST` type.

That is perfect!

src/lib.rs
```diff
@@ -1036,8 +1069,7 @@ impl Context {
         };
         let ret = Error::from_tss_rc(ret);
         if ret.is_success() {
-            let policy_digest = unsafe { MBox::<TPM2B_DIGEST>::from_raw(policy_digest_ptr) };
-            Ok(policy_digest.buffer[..policy_digest.size as usize].to_vec())
+            Ok(unsafe { Digest::try_from(*policy_digest_ptr)? })
```
hug-dev, line 1072: I am a little bit worried about removing the `MBox` here. The reason of it, I think, was to free the memory that was allocated at `policy_digest_ptr` by the callee. I am afraid we would leak memory if removed?
Would it be possible to keep that line (1039) and dereference `policy_digest` in the `Digest::try_from`? That way, when dropped, the `MBox` would still free the memory.

src/utils/mod.rs
```diff
@@ -610,6 +610,73 @@ pub enum PublicIdUnion {
...
+impl DigestList {
+    pub fn new() -> Self {
+        DigestList {
+            digests: Vec::new(),
+        }
+    }
+
+    pub fn value(&self) -> &[Digest] {
+        &self.digests
+    }
```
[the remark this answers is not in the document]
puiterwijk: Ah, right, thanks for the idea!

src/lib.rs
puiterwijk: Thanks for the feedback, fixed that!

src/utils/mod.rs
```diff
@@ -610,6 +611,73 @@ pub enum PublicIdUnion {
...
+
+impl TryFrom<DigestList> for TPML_DIGEST {
+    type Error = Error;
+    fn try_from(digest_list: DigestList) -> Result<Self> {
+        if digest_list.digests.len() < 2 {
+            return Err(Error::local_error(WrapperErrorKind::WrongParamSize));
+        }
+        if digest_list.digests.len() > 8 {
+            return Err(Error::local_error(WrapperErrorKind::WrongParamSize));
+        }
```
ionut-arm, line 667: Given the restrictions placed within the `DigestList` type, are these checks actually needed? Or is it more of a safety net in case the code above changes?

src/utils/mod.rs
```diff
@@ -610,6 +611,73 @@ pub enum PublicIdUnion {
     Ecc(Box<TPMS_ECC_POINT>),
 }
 
+// The digests field type is from bindgen which does not implement Debug on them.
```
ionut-arm, line 614: I'm confused by this - isn't `Digest` the structure defined below on line 1745, as that seems to have `Debug`?

src/utils/mod.rs
```diff
@@ -1674,31 +1742,58 @@ pub enum PublicKey {
...
+            buffer[pos] = *val;
+        }
+        Ok(TPM2B_DIGEST {
+            size: digest.len() as u16,
+            buffer,
+        })
+    }
+}
+
+type PcrValue = Digest;
```
ionut-arm, line 1795: Apart from slightly enhanced reading, is this actually needed? I don't think the type is actually referred to outside this file.

src/utils/mod.rs
```diff
@@ -1674,31 +1742,58 @@ pub enum PublicKey {
 
 /// Struct holding a pcr value.
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct PcrValue {
+pub struct Digest {
```
ionut-arm, line 1745: Nice! This is a really useful structure!

I was wondering if it's worth creating a generic `List` structure, but I'll put that idea in another issue :)

src/utils/mod.rs
puiterwijk: That was intended more as a safety net.

src/utils/mod.rs
puiterwijk: Yep, it sure is. This was from an earlier version of the code where it had `Vec<TPM2B_DIGEST>` as `digests`.

src/utils/mod.rs
puiterwijk: Given that a `PcrValue` is not exactly the same use case, I personally like to have this aliased name, just for readability.
If you disagree, I'm happy to change it.

src/utils/mod.rs
puiterwijk: And for consistency with the other `TryFrom` implementations that also perform checks on the values.

src/utils/mod.rs
ionut-arm: Happy to keep it!

PR #91, opened 2020-07-02
Add Context.policy_authorize

puiterwijk:
This abstracts the TPM2_PolicyAuthorize command to enable
use of signed policies.


puiterwijk:
@Superhepper I had an initial version of that written, just decided to then go with a trial policy. I'll see if I can add a test for a signed policy in a future PR.

Superhepper:
> @Superhepper I had an initial version of that written, just decided to then go with a trial policy. I'll see if I can add a test for a signed policy in a future PR.

Yeah, I approve of this with or without that extensive test.

ionut-arm:
@puiterwijk @Superhepper Are you ok with me releasing a new alpha version in the meanwhile? Needed some changes I merged earlier somewhere else!

puiterwijk:
@ionut-arm Sure, though I'd like it if this could get merged before, since I need this in one of my projects.
I hope to fix this in the next hour or so, just fixing the test now to be the full test (I am using the GitHub test runner as a way to run the tests, since podman doesn't want to work with the test image, and Docker (moby-engine) doesn't run on Fedora because of lack of cgroups v2 support).

ionut-arm:
> since podman doesn't want to work with the test image

Anything we can do to fix that/add support for it? Not necessarily for now, feel free to open an issue on the topic!

puiterwijk:
> > since podman doesn't want to work with the test image
> 
> Anything we can do to fix that/add support for it? Not necessarily for now, feel free to open an issue on the topic!

The problem for this is in the origin layer (`tpm2software/tpm2-tss:ubuntu-18.04`), it's basically storing files as a user whose uid/gid are outside of the range allowed by (rootless) podman.
I'm planning to work with the people maintaining that to get that fixed.

Superhepper:
I have nothing I want comitted right now so go ahead. Make a new release.

puiterwijk:
@Superhepper I ended up writing an actual (non-trial) test anyway in this PR, just with an empty `policyRef`. So this should be a good demonstration of how this works.

@ionut-arm The code should now be ready for review. If this could go into the next release, I'd be very thankful, although I would understand if I'll need to wait.

ionut-arm:
Out of curiosity, I saw that a signature verification failed on the CI test - what was the issue?

puiterwijk:
@ionut-arm I didn't see signature failures on the last run. The run before that, I was making a signature over `ahash` but validating a signature over `policy_digest`, so just a small typo there.
Before that, one failure was because I was signing `policy_digest` directly, forgetting the extra digest calculation (over `policy_ref || policy_digest`).

PR #95, opened 2020-07-07
Add functions for more Policy commands

puiterwijk:
This implements abstractions for almost all TPM2_Policy* commands.

The ones not implemented are:
- PolicyNV
- PolicyCounterTime
- PolicyDuplicationSelect
- PolicyTemplate
- PolicyAuthorizeNV

These weren't implemented because some of them require complicated structures
and others are not implemented in any of the TPMs I have available for testing.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -934,6 +934,192 @@ impl Context {
         }
     }
 
+    /// Cause conditional gating of a policy based on locality.
+    ///
+    /// The TPM will ensure that the current policy can only complete in the specified
+    /// locality (extended) or any of the specified localities (non-extended).
+    pub fn policy_locality(
+        &mut self,
+        policy_session: ESYS_TR,
+        locality: TPMA_LOCALITY,
+    ) -> Result<()> {
+        let ret = unsafe {
+            Esys_PolicyLocality(
+                self.mut_context(),
+                policy_session,
+                self.sessions.0,
+                self.sessions.1,
+                self.sessions.2,
+                locality,
+            )
+        };
+        let ret = Error::from_tss_rc(ret);
+        if ret.is_success() {
+            Ok(())
+        } else {
+            Err(ret)
+        }
+    }
+
+    /// Cause conditional gating of a policy based on command code of authorized command.
+    ///
+    /// The TPM will ensure that the current policy can only be used to complete the command
+    /// indicated by code.
+    pub fn policy_command_code(&mut self, policy_session: ESYS_TR, code: TPM2_CC) -> Result<()> {
+        let ret = unsafe {
+            Esys_PolicyCommandCode(
+                self.mut_context(),
+                policy_session,
+                self.sessions.0,
+                self.sessions.1,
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -1172,7 +1358,7 @@ impl Context {
     ///////////////////////////////////////////////////////////////////////////
 
     /// Returns a mutable reference to the native ESYS context handle.
-    fn mut_context(&mut self) -> *mut ESYS_CONTEXT {
+    pub fn mut_context(&mut self) -> *mut ESYS_CONTEXT {
```
Superhepper, line 1361: Why is this necessary?

The question regarding exposing the esys context have been discussed previously(#50, #53). The reason for not doing was to not leak the state of the context. See @ionut-arm response in #53 .

puiterwijk: Sorry, that was a test from me that I forgot to undo.

tests/context_tests.rs
```diff
@@ -640,6 +636,216 @@ mod test_policy_or {
...
+        let trial_session_attr = TpmaSessionBuilder::new()
+            .with_flag(TPMA_SESSION_DECRYPT)
+            .with_flag(TPMA_SESSION_ENCRYPT)
+            .build();
+        context
+            .tr_sess_set_attributes(trial_session, trial_session_attr)
+            .unwrap();
+
+        // There should be no errors setting an Or for a TRIAL session
+        context.policy_locality(trial_session, 3).unwrap();
```
hug-dev, line 662: What does `3` mean here?

src/lib.rs
```diff
@@ -934,6 +934,192 @@ impl Context {
...
+        } else {
+            Err(ret)
+        }
+    }
+
+    /// Cause conditional gating of a policy based on command code of authorized command.
+    ///
+    /// The TPM will ensure that the current policy can only be used to complete the command
+    /// indicated by code.
+    pub fn policy_command_code(&mut self, policy_session: ESYS_TR, code: TPM2_CC) -> Result<()> {
```
hug-dev, line 968: Can `code` be any value from the `TPM2_CC` type values [here](https://github.com/parallaxsecond/rust-tss-esapi/blob/291180c19baa35936d87058fcfeace8aa0ea3879/src/constants.rs#L121)?

tests/context_tests.rs
puiterwijk: That was just a random value for testing.
Technically, it means the non-extended locality `1 + 2` (since anything <33 is non-extended localities).

src/lib.rs
puiterwijk: Yes.

hug-dev: Ok! My thought is that since the `TPM2_CC` and `TPMA_LOCALITY` types seem to be specific values amoung a range, it would be easy to implement an abstraction for those types as Rust enumerations with the `TryFrom` and `From` traits converting the enum variant from/to the specific values.
Just an idea, let's not do that in this PR!

Issue #87, opened 2020-06-24
Refactoring


Superhepper:
The files in the library is getting way to big. It is time to try to do something about that. And while we are at it we should probably come up with some better name and structure. This is just something I go and think about while doing other stuff so I have no proposal yet how to do it. But I am putting here so that we can start a discussion about it.

One thing I can say that personally like is that lib and mod files does not contain any code. They just contain documentation, settings, clippy-settings, use statements and mod statements.

hug-dev:
Definitely agree with refactoring and your two points.

ionut-arm:
Yeah, unfortunately when I wrote a lot of the code, especially the utils stuff, I didn't pay much attention to how it would scale 😞 

The suggestion sounds pretty good, we can have a big `context` module that splits up the `Context` into the areas you mentioned. For the utils maybe something similar, or maybe we can draw some inspiration from the way the spec (especially the one on structures) is split.

Superhepper:
I will begin a 'phase 1' of this work.
This will involve the following:
1. Move context out of the lib file.
2. Move pcr selection stuff out of mod in utils.

Superhepper:
For phase 2 I am thinking if doing the following:
- Move tickets to structures.
- Create new constants folder.
- Rename constants.rs to tss.rs and move it into constants.
- Add tags to constants folder.

Superhepper:
Only one thing left to do now are the things in utils. And that is to move the remaining items out of utils/mod.rs and into their correct locations. I guess there are things in there that are not tied to any TPM structures and those I think should remain there
though in their own files.

Any one who feels they got the hang of those structures feel free to do this. I had some ideas regarding some of the things. I have been thinking that for consistency reasons there should be representations for TPMT_PUBLIC and TPM2B_PUBLIC(e.t.c.).

And if we look at the specification it should be something like:
```
|-- src
    |-- complex (key/object Complex)
        |-- mod.rs
        |-- public
            |--mod.rs
            |--tagged.rs (TPMT_PUBLIC, struct TaggedPublic)
            |--public.rs (Not sure about the name could be called buffer.rs as well. TPM2B_PUBLIC, struct Public).
```

Take these suggestions with a barrel of salt. As I said I have only been pondering about it. But not really committed any energy to it hehe.

ionut-arm:
[...]

PR #96, opened 2020-07-14
Refactoring Phase 1 increment 2.

Superhepper:
Refactoring phase one increment 2. Is part of the #87 issue.

-- Renamed PcrSelections to PcrSelectionList and moved it to
   structures/lists
-- Moved DigestList to structures/lists
-- Added dir structures/pcr
-- Created PcrSelection and PcrSelect.
-- Added Copyright info to the new files.


hug-dev:
Thanks a lot for doing the re-organisation into something that makes sense and respects the TSS structure 👌 

For the lists: @ionut-arm suggested in an older PR that we could also have our own list wrapper type over a generic parameter type. Are TSS lists defined somewhere? If it can always be represented as a `Vec` of something, then maybe we can defined our `List<T>` type and only having to define the `T` abstractions and the conversions between our `List<T>` and the `TPML_...` types? That would avoid code duplication if it can work in all cases!

About copyrights for new files: it might be better to put 2020 as the year.

hug-dev:
Yes sure, I don't want to put that as a work to do for this PR 😃 Let's create an issue though!

Superhepper:
Fixed some of the copyright and added a test for DigestList.

src/structures/lists/pcr_selection.rs
```diff
@@ -0,0 +1,223 @@
...
+        match self.items.get_mut(&hash_algorithm) {
+            Some(previously_selected_pcr_slots) => {
+                *previously_selected_pcr_slots |= selected_pcr_slots;
+            }
+            None => {
+                let _ = self.items.insert(hash_algorithm, selected_pcr_slots);
+            }
+        }
+        self
+    }
```
ionut-arm, line 204: Should there be a limit here, to check if `self.items` contains more than `PcrSelectionList::MAX_SIZE`?

Superhepper: Hrrmm...
Yes, I think there should be.

Superhepper: Or rather it should be but I am not sure it will have any effect. Because there are currently only eight items in the HashingAlgorithm enum. And the MAX_SIZE is 16. And because they are squashed together it is never possible to create a tss list with a count greater then 8.

hug-dev:
Hi @Superhepper ,

If you rebase your PR, then we can merge it!

src/structures/lists/pcr_selection.rs
ionut-arm: > Because there are currently only eight items in the HashingAlgorithm enum. And the MAX_SIZE is 16.

Ha, yes, I didn't realise that, good point!

Superhepper:
Done

PR #97, opened 2020-07-17
Add get_capabilities

puiterwijk:

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -162,6 +162,46 @@ impl Context {
         self.sessions
     }
 
+    /// Get current capability information about the TPM.
+    pub fn get_capabilities(
+        &mut self,
+        capability: TPM2_CAP,
+        property: u32,
+        property_count: u32,
+    ) -> Result<(TPMS_CAPABILITY_DATA, bool)> {
+        let mut outcapabilitydata = null_mut();
+        let mut outmoredata: u8 = 0;
+
+        let ret = unsafe {
+            Esys_GetCapability(
+                self.mut_context(),
+                self.sessions.0,
+                self.sessions.1,
+                self.sessions.2,
+                capability,
+                property,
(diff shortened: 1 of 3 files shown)


puiterwijk:
A fix for the Clippy errors is in #99.

puiterwijk:
It does not abstract the response yet, since that is a quite complicated structure.
I'm planning to do that later, just wanted to send this to see whether this abstraction is wanted.

src/context.rs
```diff
@@ -162,6 +162,47 @@ impl Context {
...
+                &mut outmoredata,
+                &mut outcapabilitydata,
+            )
+        };
+        let moredata = if outmoredata == 0 {
+            false
+        } else if outmoredata == 1 {
+            true
+        } else {
+            panic!("Invalid outmoredata value");
```
ionut-arm, line 193: It's probably better to return an error here instead of panicking. Maybe creating a new variant in `WrapperErrorKind` - `WrongValueFromTpm` or something like that?

puiterwijk: Hah, that was my idea indeed. And then I left this from my first iteration.

puiterwijk:
> Thanks for the addition!
> 
> > It does not abstract the response yet, since that is a quite complicated structure.
> > I'm planning to do that later, just wanted to send this to see whether this abstraction is wanted.
> 
> It's not strictly necessary - and especially with these vast, complicated structures (like `TPMS_CAPABILITY_DATA` and `TPM2B_PUBLIC`) the effort that needs to go into making a fully Rust-native version is demoralizing; it's probably not worth it if you don't plan to interact in complex ways with it in Rust.

I may still do it, I'll see, but I'll first fix your suggestions and then mark this as final.

puiterwijk:
This is rebased on current (post-refactor step 2) master.

PR #98, opened 2020-07-17
Remove the aes-256-symdef from TpmtSymdefBuilder

puiterwijk:
Some hardware TPMs do not support AES-256, and being able to get an easy
aes-128 is useful for those.

### DIFF
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -533,24 +533,6 @@ impl TpmtSymDefBuilder {
 
         Ok((key_bits, mode))
     }
-
-    /// Generate a `TPMT_SYM_DEF` object defining 256 bit AES in CFB mode.
-    pub fn aes_256_cfb() -> TPMT_SYM_DEF {
-        TPMT_SYM_DEF {
-            algorithm: TPM2_ALG_AES,
-            keyBits: TPMU_SYM_KEY_BITS { aes: 256 },
-            mode: TPMU_SYM_MODE { aes: TPM2_ALG_CFB },
-        }
-    }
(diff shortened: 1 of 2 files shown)


puiterwijk:
A fix for the Clippy errors is in #99.

ionut-arm:
May I suggest using the [`tss_esapi::algorithm::specifiers::Cipher`](https://docs.rs/tss-esapi/4.0.5-alpha.1/tss_esapi/utils/algorithm_specifiers/struct.Cipher.html) (be aware that the documentation version is in a different place due to recent refactorings that haven't been upstreamed) type instead? It has support for all cipher types and actually includes these the two AES CFB ciphers.

You can easily convert to `TPMT_SYM_DEF` or `TPMT_SYM_DEF_OBJECT` using `into()`

puiterwijk:
Would it in that case be an idea to just remove this function instead and point people towards using that?

ionut-arm:
Yes, I think I left it there by mistake

hug-dev:
Just as a notice: we will merge #96 relatively soon, which changes directory structure, and will trigger conflicts on your PR. Hope you don't mind :)

puiterwijk:
@hug-dev nah, that's fine. This PR is small enough, and especially if it just becomes a code removal patchset :)

puiterwijk:
The commit is updated to just remove the `aes_256_cfb` shortcut, and use `Cipher::aes_256_cfb` everywhere.

puiterwijk:
This is rebased on current (post-refactor step 2) master.

PR #99, opened 2020-07-17
Fix clippy errors

puiterwijk:
This fixes the Clippy errors that map_err is more succint than or_else
Err

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -1036,7 +1036,7 @@ impl Context {
                 self.sessions.2,
                 num_bytes
                     .try_into()
-                    .or_else(|_| Err(Error::local_error(ErrorKind::WrongParamSize)))?,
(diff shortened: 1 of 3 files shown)


puiterwijk:
@ionut-arm It was a relatively simple change to make :)

puiterwijk:
Seems I missed one in `context.rs`. This should fix that.

puiterwijk:
And next one (also fixed): clippy doesn't like useless type conversions (I guess this is a newer clippy than my version, because mine is totally fine with the code. So it may be a few more iterations).

ionut-arm:
Indeed, with every iteration of the compiler some lints are added/removed to clippy (usually) and sometimes it breaks things :) We were talking about whether we should make compiler version updates manually to prevent these kind of build failures

PR #100, opened 2020-07-20
Refactoring Phase 2

Superhepper:
This relates to #87 

- Moved tickets into structres dir.
- Created new dir constants
- Renamed constants.rs into tss.rs and moved into constants dir.
- Moved specifiers.rs from algorithm dir into constants dir and renamed
  it to algorithm.rs.
- Created tags as subdir to constants.
- Renamed utils/tags.rs to structure.rs and moved it into the
  constants/tags dir.
- Moved error part of respopnse_code.rs into sepparate file error.rs
  and moved response_code.rs into constants dir.
- Changed lints in lib.rs to make more sense.
- Moved lint from lib.rs to constants/mod.rs


src/error.rs
```diff
@@ -0,0 +1,91 @@
+use crate::constants::response_code::Tss2ResponseCode;
```
ionut-arm, line 1: I like the split, it makes sense.

Superhepper:
> 
> 
> That all looks good to me, thanks a lot for working on this! It will be primordial for more stable releases.
> 
> I have only one request: could you please check all new files if they all have the copyright notice on top and add them if not?

Done

PR #101, opened 2020-07-31
Add RSA decryption and encryption to Context

puiterwijk:
This adds `Context::rsa_encrypt` and `Context::rsa_decrypt`.

### DIFF
--- a/src/structures/buffers/publickeyrsa.rs
+++ b/src/structures/buffers/publickeyrsa.rs
@@ -1,0 +1,65 @@
+// Copyright 2020 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+use crate::tss2_esys::TPM2B_PUBLIC_KEY_RSA;
+use crate::{Error, Result, WrapperErrorKind};
+use log::error;
+use std::convert::TryFrom;
+/// Struct holding the largest RSA public key supported by the TPM
+#[derive(Debug, Clone, PartialEq, Eq, Default)]
+pub struct PublicKeyRSA {
+    value: Vec<u8>,
+}
+
+impl PublicKeyRSA {
+    const MAX_SIZE: usize = 512;
+    pub fn value(&self) -> &[u8] {
+        &self.value
+    }
+}
+
+impl TryFrom<Vec<u8>> for PublicKeyRSA {
+    type Error = Error;
+    fn try_from(bytes: Vec<u8>) -> Result<Self> {
+        if bytes.len() > PublicKeyRSA::MAX_SIZE {
+            error!("Error: Invalid Vec<u8> size(> {})", PublicKeyRSA::MAX_SIZE);
+            return Err(Error::local_error(WrapperErrorKind::WrongParamSize));
+        }
+        Ok(PublicKeyRSA { value: bytes })
+    }
+}
+
+impl TryFrom<TPM2B_PUBLIC_KEY_RSA> for PublicKeyRSA {
+    type Error = Error;
+    fn try_from(tss_data: TPM2B_PUBLIC_KEY_RSA) -> Result<Self> {
+        let size = tss_data.size as usize;
+        if size > PublicKeyRSA::MAX_SIZE {
+            error!(
+                "Error: Invalid TPM2B_PUBLIC_KEY_RSA size(> {})",
+                PublicKeyRSA::MAX_SIZE
+            );
+            return Err(Error::local_error(WrapperErrorKind::WrongParamSize));
+        }
+        Ok(PublicKeyRSA {
+            value: tss_data.buffer[..size].to_vec(),
+        })
+    }
+}
+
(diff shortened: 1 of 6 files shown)


src/structures/buffers/publickeyrsa.rs
```diff
@@ -0,0 +1,73 @@
+// Copyright 2020 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+use crate::tss2_esys::TPM2B_PUBLIC_KEY_RSA;
+use crate::{Error, Result, WrapperErrorKind};
+use log::error;
+use std::convert::TryFrom;
+/// Struct holding a data value.
```
ionut-arm, line 7: Probably worth mentioning that this is intended to hold "the largest RSA public key supported by the TPM"

src/structures/buffers/publickeyrsa.rs
```diff
@@ -0,0 +1,73 @@
...
+    }
+}
+
+impl Default for PublicKeyRSA {
+    fn default() -> Self {
+        PublicKeyRSA {
+            value: Vec::<u8>::new(),
+        }
+    }
+}
```
ionut-arm, line 26: I think just deriving `Default` on `PublicKeyRSA` gives you this as well!

src/structures/buffers/publickeyrsa.rs
```diff
@@ -0,0 +1,73 @@
...
+use log::error;
+use std::convert::TryFrom;
+/// Struct holding a data value.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct PublicKeyRSA {
+    value: Vec<u8>,
+}
+
+impl PublicKeyRSA {
+    const MAX_SIZE: usize = 512;
```
ionut-arm, line 14: I'm not sure this is always strictly the maximum allowed. I think keys of size 1024 and 2048 bits are mandated by the spec, but 4096 is up to implementation. Then again, I'm not really sure how we would check anyway... ¯\\\_(ツ)_/¯

puiterwijk: Yeah, I decided to go with the largest size acceptable. The TPM may still say "Sorry, we can't do this size".

src/structures/buffers/publickeyrsa.rs
puiterwijk: Right. The `publickeyrsa.rs` file started as a copy of `data.rs`, as you might've noticed from the other comment.
I think we may want to abstract the buffer types, just as we had an issue for the List types.

puiterwijk:
@ionut-arm yeah, this wasn't too hard. The `TransientKeyContext` side is going to be much harder, since you need an unrestricted sign + decrypt key with scheme set to `ALG_NULL` in order to be able to use both `Esys_RSA_Encrypt` and `Esys_RSA_Decrypt`. So existing keys unfortunately will not be able to use these functions, and the `TransientKeyContext` is currently heavily biased around signing.
I'm now working on implementing that layer of the story.

ionut-arm:
> 
> 
> @ionut-arm yeah, this wasn't too hard. The `TransientKeyContext` side is going to be much harder, since you need an unrestricted sign + decrypt key with scheme set to `ALG_NULL` in order to be able to use both `Esys_RSA_Encrypt` and `Esys_RSA_Decrypt`. So existing keys unfortunately will not be able to use these functions, and the `TransientKeyContext` is currently heavily biased around signing.
> I'm now working on implementing that layer of the story.

That is, unfortunately, true. Maybe creating a parallel set of methods, types, functions to the signing ones is the best/easiest option. After a brief read of the file, seems that only `load_external_rsa_public_key` is not explicitly marked for signing/verifying, but only works for that.

src/utils/mod.rs
```diff
@@ -1242,6 +1250,43 @@ pub fn create_restricted_decryption_rsa_public(
...
+    object_attributes.set_sign_encrypt(true);
+    object_attributes.set_restricted(false);
+
+    Tpm2BPublicBuilder::new()
+        .with_type(TPM2_ALG_RSA)
+        .with_name_alg(TPM2_ALG_SHA256)
+        .with_object_attributes(object_attributes)
+        .with_parms(PublicParmsUnion::RsaDetail(rsa_parms))
+        .build()
+}
```
ionut-arm, line 1288: Excellent

src/structures/buffers/publickeyrsa.rs
ionut-arm: Yeah, I'll try and think of a way, although Rust might make it somewhat complicated.

puiterwijk:
> That is, unfortunately, true. Maybe creating a parallel set of methods, types, functions to the signing ones is the best/easiest option. After a brief read of the file, seems that only `load_external_rsa_public_key` is not explicitly marked for signing/verifying, but only works for that.

I have now actually gotten an API and implementation that works within the current API.
I'll send it up for PR shortly as a different PR, so that this layer can be reviewed/merged separately after this one got merged.

PR #103, opened 2020-07-31
Implement rsa_encrypt and rsa_decrypt for TransientKeyContext

puiterwijk:
The used key will need to be created via create_signing_encryption_decryption_key
since it needs both encrypt and decrypt flags set.

### DIFF
--- a/src/abstraction/transient.rs
+++ b/src/abstraction/transient.rs
@@ -15,11 +15,12 @@
 //! and persist the underlying data.
 use crate::constants::algorithm::{Cipher, EllipticCurve, HashingAlgorithm};
 use crate::constants::tss::*;
-use crate::structures::{Auth, Digest, VerifiedTicket};
+use crate::structures::{Auth, Data, Digest, PublicKeyRSA, VerifiedTicket};
 use crate::tcti::Tcti;
 use crate::tss2_esys::*;
 use crate::utils::{
-    self, create_restricted_decryption_rsa_public, create_unrestricted_signing_ecc_public,
+    self, create_restricted_decryption_rsa_public,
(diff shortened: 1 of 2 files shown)


src/abstraction/transient.rs
```diff
@@ -99,7 +102,67 @@ impl TransientKeyContext {
...
+        })?;
+        self.context.flush_context(key_handle)?;
+        Ok((key_context, key_auth))
+    }
+
+    fn get_public_from_params(
+        &self,
+        params: KeyParams,
+        encryption_decryption: bool,
+    ) -> Result<TPM2B_PUBLIC> {
```
ionut-arm, line 165: I think instead of using a `bool` to signal which path to take you can change the `KeyParams` enum to have `RsaSign` and `RsaEncrypt` variants, and for the latter you don't allow a scheme, just size and public exponent.

src/abstraction/transient.rs
```diff
@@ -207,6 +278,86 @@ impl TransientKeyContext {
...
+        let key_handle = self.context.context_load(key_context)?;
+        if let Some(key_auth_value) = key_auth {
+            self.context
+                .tr_set_auth(key_handle, &key_auth_value)
+                .or_else(|e| {
+                    self.context.flush_context(key_handle)?;
+                    Err(e)
+                })?;
+        }
+        let scheme = AsymSchemeUnion::RSAOAEP(HashingAlgorithm::Sha256).get_rsa_decrypt_struct();
```
ionut-arm, line 305: Hmm, should the hash algorithm be chosen by the user? I don't know how widespread use of non-SHA256 is.

ionut-arm: If we keep it like this, it's worth adding a line in the docs above to mention the encryption scheme (since it's not chosen when the key is created, like with signing).

Issue #102, opened 2020-07-31
Add support for non volatile storage api


Superhepper:
Currently neither context APIs nor the necessary structs for handling non volatile storage exists. This needs to be added in order to support for the non volatile storage in the TPM.

PR #104, opened 2020-08-02
First step to support NV operations.

Superhepper:
This is a first step in order to add support for non volatile storage in the TPM(#102).

- Added folder src/nv (Non volatile).
- Added folder src/nv/storage (Non volatile storage).
- Added file index.rs to src/nv/storage:
  This file contains NvIndex, NvIndexType, NvIndexHandle, and NvIndexAttributes types.
- Added file public.rs to src/nv/storage:
  This file contains the NvPublic, NvPublicBuilder types.
- Added file authorization.rs to src/nv/storage:
  This file contains the NvAuthroization type.
- Added file max_nv.rs to src/structures/buffers:
  This file contains the MaxNvBuffer type.
- Added Context APIs:
    -- nv_define_space
    -- nv_undefine_space
    -- nv_read_public
    -- nv_read
    -- nv_write

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -2,9 +2,10 @@
 // SPDX-License-Identifier: Apache-2.0
 use crate::algorithm::structures::SensitiveData;
 use crate::constants::algorithm::HashingAlgorithm;
+use crate::nv::storage::{NvAuthorization, NvIndexHandle, NvPublic};
 use crate::structures::{
-    Auth, Data, Digest, DigestList, HashcheckTicket, MaxBuffer, Name, Nonce, PcrSelectionList,
-    PublicKeyRSA,
+    Auth, Data, Digest, DigestList, HashcheckTicket, MaxBuffer, MaxNvBuffer, Name, Nonce,
+    PcrSelectionList, PublicKeyRSA,
 };
 use crate::tcti::Tcti;
 use crate::tss2_esys::*;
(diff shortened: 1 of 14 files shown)


src/nv/storage/index.rs
```diff
@@ -0,0 +1,129 @@
...
+            TPM2_NT_PIN_PASS => Ok(NvIndexType::PinPass),
+            _ => {
+                error!("Error: Found invalid when trying to parse Nv Index Type.");
+                Err(Error::local_error(WrapperErrorKind::InvalidParam))
+            }
+        }
+    }
+}
+
+bitfield! {
```
hug-dev, line 80: Pretty nice use of bitfields 🆗

src/nv/storage/index.rs
```diff
@@ -0,0 +1,129 @@
...
+    pub pp_write, set_pp_write: 0;
+    pub owner_write, set_owner_write: 1;
+    pub auth_write, set_auth_write: 2;
+    pub policy_write, set_policy_write: 3;
+    TPM2_NT, tss_index_type, _: 7, 4; // private getter
+    pub TPM2_NT, from into NvIndexType, _, set_index_type: 7, 4;
+    // Reserved 9,8
+    pub policy_delete, set_policy_delete: 10;
+    pub write_locked, set_write_locked: 11;
+    pub wriate_all, set_write_all: 12;
```
hug-dev, line 95: Typo of `write_all` I think

src/nv/storage/public.rs
```diff
@@ -0,0 +1,60 @@
...
+    type Error = Error;
+    fn try_from(nv_public: NvPublic) -> Result<TPM2B_NV_PUBLIC> {
+        Ok(TPM2B_NV_PUBLIC {
+            size: NvPublic::STRUCTURE_SIZE as u16,
+            nvPublic: TPMS_NV_PUBLIC {
+                nvIndex: nv_public.index.into(),
+                nameAlg: nv_public.index_name_alg.into(),
+                attributes: nv_public.index_attributes.into(),
+                authPolicy: nv_public.index_auth_policy.try_into()?,
+                dataSize: nv_public.data_area_size as u16,
```
hug-dev, line 56: Should there be a check if `nv_public.data_area_size` is lesser or equal than `u16::MAX`

src/nv/storage/index.rs
Superhepper: Nice catch thanks

src/nv/storage/public.rs
Superhepper: Yes! It should. Thank you.

hug-dev:
Looks good to me, thanks! Will approve once ready to review.

Superhepper:
I just need to add a few more tests.

Superhepper:
It is finally ready for review.

src/context.rs
```diff
@@ -1328,6 +1329,165 @@ impl Context {
...
 
+    ///////////////////////////////////////////////////////////////////////////
+    /// TPM Non Volatile Section
+    ///////////////////////////////////////////////////////////////////////////
+
+    /// NV Define Space
+    ///
+    /// Allocates an index in the non volatile
+    /// storage.
+    pub fn nv_define_space(
```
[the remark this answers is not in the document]
ionut-arm: Yeah, was thinking of doing that for `Context`, as it's becoming a very large file

src/nv/storage/authorization.rs
```diff
@@ -0,0 +1,64 @@
...
+impl TryFrom<Hierarchy> for NvAuthorization {
+    type Error = Error;
+
+    fn try_from(hierarchy: Hierarchy) -> Result<NvAuthorization> {
+        match hierarchy {
+            Hierarchy::Owner => Ok(NvAuthorization::Owner),
+            Hierarchy::Platform => Ok(NvAuthorization::Platform),
+            _ => {
+                error!("Error: Found invalid value when trying to convert Hierarchy to NV Authroization.");
+                Err(Error::local_error(WrapperErrorKind::InvalidParam))
```
ionut-arm, line 36: Obviously not something for this PR, but I'm wondering if the concept of `InvalidParam` is waaay too dilluted in this crate. We use it for a lot of different things, maybe we should split it out into more fine-grained errors.

src/nv/storage/index.rs
```diff
@@ -0,0 +1,195 @@
...
+    }
+}
+
+impl From<TPMI_RH_NV_INDEX> for NvIndex {
+    fn from(tss_nv_index: TPMI_RH_NV_INDEX) -> NvIndex {
+        NvIndex {
+            value: tss_nv_index,
+        }
+    }
+}
```
ionut-arm, line 75: I don't think this one is correct, you should be passing the `tss_nv_index` through `create_from_value` because there's no guarantee that the `u32` is has the correct format - you can pass any `u32` in here since the `TPMI_RH_NV_INDEX` is just an alias.

Superhepper: You are right. I was reading the specification which says that that is the interface type used and it is guaranteed to only have values in the correct range. Though it does not prevent some one from creating incorrect ones so it needs to be checked.

PR #105, opened 2020-08-03
add abstraction::QuoteContext

genofire:
- easy quote
- verify signature of a quote
- create and load keys from file

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ documentation = "https://docs.rs/crate/tss-esapi"
 [dependencies]
 bitfield = "0.13.2"
 serde = { version = "1.0.103", features = ["derive"] }
+hex = "0.4"
 mbox = "0.5.0"
 log = "0.4.8"
 enumflags2 = "0.6.3"
--- a/src/abstraction/mod.rs
+++ b/src/abstraction/mod.rs
@@ -1,4 +1,5 @@
 // Copyright 2019 Contributors to the Parsec project.
 // SPDX-License-Identifier: Apache-2.0
 
+pub mod quote;
 pub mod transient;
--- a/src/abstraction/quote.rs
+++ b/src/abstraction/quote.rs
@@ -1,0 +1,629 @@
+// Copyright 2019 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Module for abstracting resource handle management
+//!
+//! This module presents an abstraction over the TPM functionality exposed through the core
+//! `Context` structure. The abstraction works by hiding resource handle management from the
+//! client. This is achieved by passing objects back and forth in the form of contexts. Thus, when
+//! an object is created, its saved context is returned and the object is flushed from the TPM.
+//! Whenever the client needs to use said object, it calls the desired operation with the context
+//! as a parameter - the context is loaded in the TPM, the operation performed and the context
+//! flushed out again before the result is returned.
+//!
+//! Object contexts thus act as an opaque handle that can, however, be used by the client to seralize
+//! and persist the underlying data.
+use crate::constants::algorithm::{Cipher, EllipticCurve, HashingAlgorithm};
+use crate::constants::tss::*;
+use crate::structures::{Auth, Data, Digest, MaxBuffer, PcrSelectionList, VerifiedTicket};
+use crate::tcti::Tcti;
+use crate::tss2_esys::*;
+use crate::utils::{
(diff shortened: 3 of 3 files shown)


ionut-arm:
Hi! Thanks for the changes - support for quote and signing/verifying signatures over quotes is very welcome! However, the `TransientKeyContext` is intended to handle the context saving, loading and evictions for users, so removing that feature isn't in line with the goal of the abstraction. The main feature of it is that all operations can be performed in a way that treats the `TpmsContext` as an opaque handle for the key. Doing the creation and loading in different steps blows away that abstraction.

If you **need** that functionality, though, I'd be more than happy with another abstraction type that operates in the way that you need - you can copy the code you want from `TransientKeyContext` and implement whatever other functionality you need on it.

genofire:
@ionut-arm 
would you merge the commits for:
- `abstraction::TransientKeyContext - add support for loading private keys`
- `abstraction::TransientKeyContext - create only keys without loading
???

and drop `signature` and `verifiy_signature` ?

ionut-arm:
> 
> 
> @ionut-arm
> would you merge the commits for:
> 
>     * `abstraction::TransientKeyContext - add support for loading private keys`
> 
>     * `abstraction::TransientKeyContext - create only keys without loading
>       ???
> 
> 
> and drop `signature` and `verifiy_signature` ?

* `create only keys without loading` is definitely a no-go - same argument I outlined above, it breaks the abstraction, but you could create a separate abstraction layer to handle things your way!
* `add support for loading private keys` looks somewhat better than the method we currently have (`load_external_rsa_public_key`) which should really be rewritten to make it more flexible. I think I'd prefer your `load_rsa_key` method, but instead of hardcoding `AsymSchemeUnion::RSASSA(HashingAlgorithm::Sha256)`, it should be a param.

If you want a proper review of the code I could do that, but be aware that some tests will also be needed

ionut-arm:
Hi!

Can I ask what the use case is that you're trying to achieve? I see a lot of methods for creating and loading keys, plus the ones for signing and verifying quotes - are you just interested in having the quote in an easier format?

I think some of the methods seem more like application logic than belonging to this library - the one for loading a key from a file or creating it in particular. Especially with the (somewhat random) file formatting, I think it makes more sense to implement that in your application code! You could try and implement a format based on a specification (maybe something like [DER](https://en.wikipedia.org/wiki/X.690) ) and that would fit in this crate, but if you don't care too much about that then your app is probably a better place for handling the file storage.

Also, what kind of keys are you interested in - both RSA and ECC, or just one of them?

genofire:
Yes, my goal is to get a quote on a easy way, signed with the same key over restarts.
And of course to verifiy the signature of the quote.

Therefore i write fast this easy code just with RSA support and a easy file - at the moment.

in my project i use it this way ATM - but i am willing to refactor it (again) to a state that the esapi could could handle it on a better way.
Therefore i wish to get some useful tips.
- like the DER ... on which scheme should it be stored, without to be a complete Certificate? (maybe is PKCS 8 more usefull?)
- and on which place (with function signature) should it be placed in the esapi

ionut-arm:
Quick question - do you intend to use the keys externally from the TPM or with some other library? The ones that you store on disk. If you only want to use them with this library, then simply obtaining the key as a `TpmsContext` is probably the easiest way to go. that's what we do in Parsec , we write `TpmsContext` to a file and when we need the key we read it back into a variable and use it with the `TransientKeyContext`. 

I have some ideas, will write them down in an issue, maybe we can create an abstract context on top of the `TransientKeyContext` that saves all keys in files instead of returning the `TpmsContext`.

genofire:
No, the generated keys should only be usable with the expected TPM (otherwise it is a security problem and make this part of a TPM useless ...)
It is just to be restart / reboot save, otherwise the `TransientKeyContext` will generate a new key, where the public key changed, too.
On this way, it is not possible to verify signatures without having a secure mechanism to exchange the new public key.

(Current i am able to fetch the public key at beginning and be safe for the signatures will would be generated in the future)

ionut-arm:
> It is just to be restart / reboot save, otherwise the TransientKeyContext will generate a new key, where the public key changed, too.

The `TransientKeyContext` works for that too! We rely on it for Parsec. The main key (`root_key`) is just a primary key that helps with seeding the rest of them, but once the keys are created and saved as `TpmsContext` you can load them again even after a reboot! That's because they're not dependent on that `root_key` anymore. You can try it out, see if it works for you.

ionut-arm:
Hi! Did you manage to test the `TransientKeyContext`?

If you want to chat, we're more easily reachable on Slack - find a link [here](https://github.com/parallaxsecond/community)

PR #106, opened 2020-08-03
Convert PcrData to TPML_DIGEST

genofire:


src/utils/mod.rs
```diff
@@ -1496,3 +1496,20 @@ impl<'a> IntoIterator for &'a PcrData {
         self.data.iter()
     }
 }
+
+impl Into<TPML_DIGEST> for PcrData {
+    fn into(self) -> TPML_DIGEST {
```
ionut-arm, line 1501: The implementation of the function looks good, but I'd recommend implementing `From<PcrData> for TPML_DIGEST` instead, as it's more idiomatic. From the [documentation of `Into`](https://doc.rust-lang.org/std/convert/trait.Into.html):

> One should avoid implementing `Into` and implement `From` instead. Implementing `From` automatically provides one with an implementation of `Into` thanks to the blanket implementation in the standard library.

genofire:
Any reason, why it is not merged yet?

PR #108, opened 2020-08-06
Added rust types for handles.

Superhepper:
This is an initial work that adds the basic
handle rust types for ESYS ESAPI(ESYS_TR) handles and
TPM handles(TPM2_HANDLE). Some API will require more specific
handle types which can be implemented later.

### DIFF
--- a/src/handles/tpm.rs
+++ b/src/handles/tpm.rs
@@ -1,0 +1,181 @@
+// Copyright 2020 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+use crate::{
+    constants::tss::{
+        TPM2_HT_AC, TPM2_HT_HMAC_SESSION, TPM2_HT_LOADED_SESSION, TPM2_HT_NV_INDEX, TPM2_HT_PCR,
+        TPM2_HT_PERMANENT, TPM2_HT_PERSISTENT, TPM2_HT_POLICY_SESSION, TPM2_HT_SAVED_SESSION,
+        TPM2_HT_TRANSIENT,
+    },
+    tss2_esys::TPM2_HANDLE,
+    Error, Result, WrapperErrorKind,
+};
+use log::error;
+use std::convert::{From, TryFrom};
+use std::stringify;
+
+/// Enum representing the different types of tpm handles
+/// of a TPM handle.
+///
+/// * Details
+/// The TPM handles are used
+/// to reference shielded locations of various
+/// types within the TPM.
+///
+/// * OBS
+/// Do not confuse the TpmHandles with the
+/// ESYS [ObjectHandle](crate::handles::esys::ObjectHandle).
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub enum TpmHandle {
+    Pcr(PcrTpmHandle),
+    NvIndex(NvIndexTpmHandle),
+    HmacSession(HmacSessionTpmHandle),
+    LoadedSession(LoadedSessionTpmHandle),
+    PolicySession(PolicySessionTpmHandle),
+    SavedSession(SavedSessionTpmHandle),
+    Permanent(PermanentTpmHandle),
+    Transient(TransientTpmHandle),
+    Persistent(PersistentTpmHandle),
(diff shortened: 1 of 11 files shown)


src/handles/tpm.rs
```diff
@@ -0,0 +1,60 @@
...
+        let most_significant_byte = value.to_be_bytes()[0];
+        match most_significant_byte {
+            TPM2_HT_PCR => Ok(TpmHandle { value }),            // 0x00
+            TPM2_HT_NV_INDEX => Ok(TpmHandle { value }),       // 0x01
+            TPM2_HT_HMAC_SESSION => Ok(TpmHandle { value }), // 0x02, Same as TPM2_HT_LOADED_SESSION
+            TPM2_HT_POLICY_SESSION => Ok(TpmHandle { value }), // 0x03, Same as TPM2_HT_SAVED_SESSION
+            TPM2_HT_PERMANENT => Ok(TpmHandle { value }),      // 0x04
+            TPM2_HT_TRANSIENT => Ok(TpmHandle { value }),      // 0x80
+            TPM2_HT_PERSISTENT => Ok(TpmHandle { value }),     // 0x81
+            TPM2_HT_AC => Ok(TpmHandle { value }),             // 0x90
```
hug-dev, line 40: If the TPM handles are part of a set, maybe it would be nice for the `TpmHandle` type to be an enum of those different categories. Maybe that will help later with static checks.

Superhepper: My thinking is that TpmHandle is the base and all specific version of TpmHandle can always be converted to TpmHandle. 

So if we take the NvIndex type from the NV storage PR(#104)

I would rename NvIndex to NvIndexTpmHandle and move that type into src/handles/tpm.rs
```rust
pub struct NvIndexTpmHandle {
    tpm_handle: TpmHandle,
}

impl NvIndexTpmHandle {
   pub fn new(value: u32) -> Result<NvIndexTpmHandle> {
       if value.to_be_bytes()[0] != TPM2_HT_NV_INDEX {
            error!("Error: Invalid most significant byte is invalid (!={})", TPM2_HT_NV_INDEX);
            Err....
       }
       NvIndexTpmHandle { value: TpmHandle {value} }
    }
}

imp From<NvIndexTpmHandle> for TpmHandle {
    fn from(nv_index_tpm_handle: NvIndexTpmHandle) -> TpmHandle {
         nv_index_tpm_handle.value    
    }
}
```

Then the functions that only uses nv index tpm handles can clearly state that and any attempts at trying to use the wrong type of handle will be catched at compile time.

I am note sure I understood what it was you suggested so could you please elaborate?

hug-dev: > Then the functions that only uses nv index tpm handles can clearly state that and any attempts at trying to use the wrong type of handle will be catched at compile time.

Agree with that and what you said!
I was thinknig taking it a little step further and having:
```rust
pub enum TpmHandle {
    Pcr(Pcr),
    NvIndex(NvIndex),
    ... etc
}

pub struct NvIndex {
    value: u32,
}

imp From<NvIndex> for TpmHandle {
    fn from(nv_index: NvIndex) -> TpmHandle {
         TpmHandle::NvIndex(nv_index)    
    }
}
```
We use this idiom quite a lot in other repos and I think it's nice 😃 But agree with yours as well.

Superhepper: Hrrm. Yeah I agree with you. It will be easier to do it your way for APIs that can return any type of TPM handle. I changed it. I also fixed that none of the new types were publicly available (hehe oops).

src/handles/tpm.rs
```diff
@@ -0,0 +1,181 @@
...
+        }
+
+        impl TryFrom<TPM2_HANDLE> for $handle_type_name {
+            type Error = Error;
+            fn try_from(tss_tpm_handle: TPM2_HANDLE) -> Result<$handle_type_name> {
+                $handle_type_name::new(tss_tpm_handle)
+            }
+        }
+    };
+}
```
ionut-arm, line 149: Nice, this is exactly what I've just done for Buffer as well, shall push a PR today

PR #112, opened 2020-08-07
Create a generic buffer implementation

ionut-arm:
This commit implements a generic implementation of buffer types (those
starting with `TPM2B_` in the spec) and migrates the existing native
buffers to the new implementation.


src/structures/buffers.rs
```diff
@@ -0,0 +1,79 @@
...
+
+#[allow(unused_macros)]
+macro_rules! buffer_type {
+    ($native_type:ident,$MAX:expr,$tss_type:ident) => {
+        use crate::tss2_esys::$tss_type;
+
+        #[derive(Debug, Clone, PartialEq, Eq, Default)]
+        pub struct $native_type(Vec<u8>);
+        impl $native_type {
+            const MAX_SIZE: usize = $MAX;
```
Superhepper, line 17: If there is no point having the max sized stored like that in the macro unless we make it public. And I think it should be public in order for user to avoid creating errors.

src/structures/buffers.rs
```diff
@@ -0,0 +1,79 @@
...
+            type Target = Vec<u8>;
+            fn deref(&self) -> &Self::Target {
+                &self.0
+            }
+        }
+
+        impl TryFrom<Vec<u8>> for $native_type {
+            type Error = Error;
+
+            fn try_from(bytes: Vec<u8>) -> Result<Self> {
```
Superhepper, line 34: I wish there was some way we could make a try_from from a slice. But unfortunatly I do not think it is possible.

src/structures/mod.rs
```diff
@@ -25,37 +25,37 @@ pub use names::name::Name;
 mod buffers;
 pub use self::auth_buffer::Auth;
 pub mod auth_buffer {
-    pub use super::buffers::auth::*;
+    pub use super::buffers::Auth;
```
Superhepper, line 28: These make no sense now. They should be removed.

ionut-arm: Should it just be `pub mod buffers` then?

Superhepper: Yes. Because previously the xxx_buffer mod was used in order to access extra things that had not been declared pub use.

The macro could add each buffer type in its own pub mod. So that it would be possible to add stuff to the macro without having to explicitly declare it with pub use in order to make it available.

ionut-arm: I've simply made them public from the `structures` module - reckon the modules themselves are useful?

Superhepper: The only reason I can see for keeping the modules is the following is if you wish to add like free functions or something similar where the name does not change depending buffer type. Or if you add free functions with the same name but different return types. That won't work unless they are 1. Generic or 2. in different modules.

Superhepper: I am actually starting to lean against that it would be a good idea to keep the different buffer types in different modules. That way you can leave the src/structures/buffers/mod.rs unchanged and just add

```rust
macro_rules! buffer_type {
    ($module_name:idnet,$native_type:ident,$MAX:expr,$tss_type:ident) => {
        pub mod $module_name {
            use super::*;
            use crate::tss2_esys::$tss_type;
```

ionut-arm: I don't think it would work that way since it would be impossible to add things in the module that are not in the macro. I'll just wrap each macro call in a module.

ionut-arm: Will not recreate the previous export hierarchy, though - will only export the types from `structures/mod.rs`, not the modules as well, since they only contain the type.

Superhepper: Yeah, I think that is fine. I do not think we have a reason for having the previous export structure so removing it is fine and each type is confined in its own mod with its own imports which is nice to.

Issue #113, opened 2020-08-11
NV contexts methods takes incorrect authorization argument.


Superhepper:
The
```nv_define_space```
and
```nv_undefine_space```
behaves correctly with regard to authorization because the take an NvAuthorization.

From specification: 
>TPM2_NV_DefineSpace Command:
>|Type | Name | Description|
>|:---|:---|:---|
>|TPMI_RH_PROVISION| @authHandle |TPM_RH_OWNER or TPM_RH_PLATFORM+{PP}<BR /> Auth Index: 1<BR />Auth Role: USER|

And the TPM2_NV_UndefineSpace looks similair.

The bug regards the authorization for commands such as ```nv_read``` and ```nv_write``` (and the remaining commands that will be implmented later). Currently they look like this:
```rust
    pub fn nv_read(
        &mut self,
        nv_authorization: NvAuthorization,
        nv_index_handle: NvIndexHandle,
        size: u16,
        offset: u16,
    ) -> Result<MaxNvBuffer>
```
and 
```rust
    pub fn nv_write(
        &mut self,
        nv_authorization: NvAuthorization,
        nv_index_handle: NvIndexHandle,
        data: &MaxNvBuffer,
        offset: u16,
    ) -> Result<()>
```

Here the specification says something different:
> Table 220 — TPM2_NV_Write Command
>|Type | Name | Description|
>|:---|:---|:---|
>|TPMI_RH_NV_AUTH| @authHandle |handle indicating the source of the authorization value<BR /> Auth Index: 1<BR /> Auth Role: USER|

So for those commands the authorization is not limited to ```Platform``` or ```Owner``` and therefor should not take a NvAuthorization as argument.

Issue #24, opened 2020-01-09
Create enum wrappers around resource handle types


ionut-arm:
Resource handle types that have various variants (e.g. hierarchies) would benefit from better static enforcing through Rust enum wrappers.

Superhepper:
Work for the TPM handles where made in #108 but that left the Esys handles to just one type which isn't really correct.

Something similar needs to be done for the esys handles.

Superhepper:
This have in general been solved. There might still be things to do but that can be handled as separate issues.

PR #114, opened 2020-08-11
Improve esys handles handling

Superhepper:
This PR aims to improve handling of esys handles(ESYS_TR) [ref #24] and solve the bug with nv authorization argument in the nv API:s [ref #113 ]


src/handles/handle.rs
```diff
@@ -0,0 +1,330 @@
...
+pub mod object {
+    impl_basic_handle!(
+        /// General handle type.
+        ObjectHandle
+    );
+
+    impl ObjectHandle {
+        pub const fn new(value: u32) -> ObjectHandle {
+            ObjectHandle { value }
+        }
```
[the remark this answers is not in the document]
ionut-arm: The same check should probably be done in any conversion from `ESYS_TR` too

src/handles/handle.rs
```diff
@@ -0,0 +1,330 @@
...
+    use super::object::ObjectHandle;
+    impl_basic_handle!(
+        /// NV Index Handle
+        ///
+        /// Represents an esys resource handle
+        /// for a nv index.
+        NvIndexHandle
+    );
+    impl_handle_conversion!(NvIndexHandle, ObjectHandle);
+    impl_handle_conversion!(NvIndexHandle, AuthHandle);
```
ionut-arm, line 301: Is there a place where these equivalences are listed/mentioned?

src/handles/handle.rs
```diff
@@ -0,0 +1,330 @@
...
+
+/// Auth handle module
+pub mod auth {
+    use super::object::ObjectHandle;
+    impl_basic_handle!(
+        /// Auth handle
+        ///
+        /// Represents an esys handle to resources
+        /// that can be used for authentication.
+        AuthHandle
```
ionut-arm, line 284: I'm very confused as to what this type is supposed to represent. I'm not sure you've intended it the same way `ESYS_TR_RH_AUTH` is defined in the ESAPI spec

src/handles/handle.rs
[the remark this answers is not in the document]
ionut-arm: Actually, now that I think of it, this shouldn't really be public. `ESYS_TR` handles are only created by the ESAPI layer - we have `Copy` and `Clone` implemented on the `ObjectHandle` type, so it can be duplicated if necessary. Are there use cases where the client is the one specifying what handle an object should have?

src/handles/handle.rs
```diff
@@ -0,0 +1,330 @@
...
+    use super::object::ObjectHandle;
+    impl_basic_handle!(
+        /// Session Handle
+        ///
+        /// Represents an esys handle used for
+        /// referencing session resources.
+        SessionHandle
+    );
+    impl_handle_conversion!(SessionHandle, ObjectHandle);
+    impl_handle_conversion!(SessionHandle, AuthHandle);
```
ionut-arm, line 329: Shouldn't there be a conversion from `PasswordHandle` to `SessionHandle` too?

src/handles/handle.rs
Superhepper: Not really. Because they are not a real thing in terms of the specification. The specifications I think mentions that the nv operations needs an esys handle to the nv index but for esys everything is ESYS_TR. So that is taken from how I have had to use it. If you look at the tests. There you can see how those conversions are used.

src/handles/handle.rs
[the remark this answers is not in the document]
Superhepper: Your trail of thought are exactly like mine when I wrote this!

I decided to leave it there in the draft but I do not have anything against removing it.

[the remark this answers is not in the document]
ionut-arm: I don't think we need to remove it, but `pub(crate)` is probably safer - at least that way we can still use it.

src/handles/handle.rs
Superhepper: Nope I have not. An AuthHandle is a handle to anything that have metadata in it that can be used for authorization.
So for example when you do
```rust
context.set_sessions((ESYS_TR_PASSWORD, ESYS_TR_NONE, ESYS_TR_NONE));
let nv_index_handle = context.nv_define_space(NvAuthorization::Owner, Some(&Auth::try_from(vec![0,1,2,3])?), &nv_public).unwrap();
```
Then the auth is stored in esys and is associated with the handle returned from esapi. And the nv index handle can then be used as an AuthHandle.


And I was a little bit unsure how to interpret that range of values in the specification:
> #define ESYS_TR_RH_AUTH_FIRST 0x110U
> #define ESYS_TR_RH_AUTH(x) (ESYS_TR_RH_AUTH_FIRST + (ESYS_TR)(x))

Didn't understand if that was something I had to bother with or not.

Superhepper: So those would have been added to the TpmConstantsHandle(terrible name I know but it is what esapi spec calls them).

src/handles/handle.rs
Superhepper: I was thinking of making PasswordHandle and NoneHandles into new types with only one value. In order to be more consistent....

src/handles/handle.rs
Superhepper: They are mainly there to make sure that a key handle is not sent into an API that requires an nv index handle without that 'conversion' explicitly being stated.

Issue #80, opened 2020-06-09
Handle breaking changes in tpm2-tss from 2.4.0


Superhepper:
I am just putting this here to remind us. That tpm2-tss 2.4.0 and greater introduces some breaking changes. Because this is in alpha we have the possibility to just increase the required version of tpm2-tss. But I am not sure which path is the best.

Superhepper:
I will have look and see if I can improve things.

Superhepper:
Unfortunately there is not much to improve. The different types in this case are just type aliases for u32. So I could not think of any trick / hack in order to figure out what values to use.

1. We move everything to the latest tpm2-tss 2.4.X. Fix all tests and things that breaks with this change. And be done with it.

2. We try to maintain backwards compatibility by looking at the version in pkgconfig and do necessary adjustments in the code with regard to version.

Personally I would vote for number 1. But what do you guys think @hug-dev @ionut-arm ?

hug-dev:
I think since we are still on alpha releases, it is fine and expected to do breaking changes. So I would say that we can improve the `tpm2-tss` version we support. Is the hierarchy type the only breaking change?

Superhepper:
I have begun doing some testing with running the CI container with updated versions. And it seams to be more things that breaks. I will look into it further and see why those things stops working.

Superhepper:
It seams as the breaking change in 2.4.X was removed, And they moved it into 3.0.0-dev version of the library. But I am going to continue to look at upgrading I am still finding some issues with it that I need to resolve.

The following tests are still failing and I need to understand why:
    two_signatures_different_digest
    verify_wrong_digest
    verify_wrong_key

ionut-arm:
I'll try it out as well when I get a bit of free time!

Are those failing with a reasonable error or is it something obscure/non-intuitive?

Superhepper:
Well now it is only failing on my machine. In #84 I uploaded a draft PR to see if the CI would be able to run it and that worked fine.

hug-dev:
Can we close this issue?

Superhepper:
Yes. I am not sure if we have made another issue for the things we discussed with regard to having two different builds in CI. One with MIN_VERSION = 2.3.3 and one with latest compatible 2.x.x?

hug-dev:
Ok I will rename that issue to be just that.

hug-dev:
Actually created a new one.

PR #121, opened 2020-08-18
bugfix: hierarchy use esys_rh not tpm2_rh in esys calls

genofire:
it fix log output / fallback in `esys_iutil.c` like:
```
ERROR:esys:src/tss2-esys/esys_iutil.c:389:iesys_handle_to_tpm_handle() Error: Esys invalid ESAPI handle (40000007).
WARNING:esys:src/tss2-esys/esys_iutil.c:410:iesys_is_platform_handle() Convert handle from TPM2_RH to ESYS_TR, got: 0x40000007
```

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -604,7 +604,7 @@ impl Context {
                 self.sessions.2,
                 private,
                 public,
-                hierarchy.rh(),
+                hierarchy.esys_rh(),
                 &mut key_handle,
             )
         };
@@ -635,7 +635,7 @@ impl Context {
                 self.sessions.2,
                 null(),
                 public,
-                hierarchy.rh(),
+                hierarchy.esys_rh(),
                 &mut key_handle,
             )
         };
@@ -1214,7 +1214,7 @@ impl Context {
                 self.sessions.2,
                 &in_data,
                 hashing_algorithm.into(),
-                hierarchy.rh(),
+                hierarchy.esys_rh(),
                 &mut out_hash_ptr,
                 &mut validation_ptr,
             )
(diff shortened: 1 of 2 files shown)


genofire:
@Superhepper do you like to review? i believe you know it at best ....

Superhepper:
What version of tpm2-tss lin are you usling?

Superhepper:
Because with version 2.x.x of the tpm-tss lib they will not make any breaking changes. Like what kind of handle that they are using in the API. Even though they are not compliant with the specification. All breaking changes are there for added to the 3.0-dev version. The version you will probably get if you just take the latest from their master.

genofire:
i am using tpm2-tss version 3.0.0

genofire:
LOG_ERROR("Error: Esys invalid ESAPI handle (%x).", esys_handle);

Superhepper:
Because changing the handle type is a breaking change. The tpm2-tss did not want to have any breaking changes within the same major version. So originally this fix was merged into 2.4.x of tpm2-tss but was later removed from there becuase it was breaking the backward compatibility. So it was moved into 3.x.x instead.

The rust-tss-esapi library is at the moment only compatible with major version 2 tpm2-tss (> 2.3.3 to be precise). The CI container is using version 2.3.3 when building and testing the library. That is why the tests in this PR is failing.

How to add support for the breaking changes in version > 3.0.0 of tpm2-tss is whole different issues that have not fully been discussed. We had some initial thoughts regarding this (#80 ) when the breaking change was introduced in 2.4.0 but since it was dropped it was not discussed any further.

Issue #123, opened 2020-08-20
No private key analog to `load_external_rsa_public_key`


joechrisellis:
Hi!

I'm currently looking at [this issue](https://github.com/parallaxsecond/parsec/issues/126) regarding adding private key support to the `PsaImportKey` operation. For public keys, Parsec uses the `load_external_rsa_public_key` to [get the public key context](https://github.com/parallaxsecond/parsec/blob/92216ee683bd6a021366d1354cceb00eefd5efb2/src/providers/tpm_provider/key_management.rs#L186). It looks like there is currently no analog for private keys (e.g. `load_external_rsa_private_key` or `load_external_rsa_keypair`). Is this something we need? 🙂

joechrisellis:
@hug-dev I'm no expert but that seems like a good approach. I have a patch ready that I'll submit a draft PR for but it'll 100% need some tweaking.

PR #124, opened 2020-08-20
Add `load_external_rsa_keypair` function

joechrisellis:
Draft PR -- would like to see if this is anywhere close to what would be needed to solve [issue 123](https://github.com/parallaxsecond/rust-tss-esapi/issues/123) or whether a different approach would be preferred. Would be nice to get some eyes on this. Feel free to nag me if this is way off the mark! 😅


src/abstraction/transient.rs
```diff
@@ -179,6 +179,79 @@ impl TransientKeyContext {
...
+        let private = TPM2B_SENSITIVE {
+            size: private_key_len,
+            sensitiveArea: TPMT_SENSITIVE {
+                sensitiveType: TPM2_ALG_RSA,
+                authValue: Default::default(),
+                seedValue: Default::default(),
+                sensitive: TPMU_SENSITIVE_COMPOSITE {
+                    rsa: TPM2B_PRIVATE_KEY_RSA {
+                        size: private_key_len,
+                        buffer: private_key_buffer,
```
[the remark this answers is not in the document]
ionut-arm: So I was having a look at [`tpm2_loadexternal`](https://www.mankier.com/1/tpm2_loadexternal) code and from there I can tell that they 
* load the public key from a PEM file (modulus and exponent) into `TPM2B_PUBLIC`
* load the first prime in the private key (called `p` in the [PKCS #1 RFC](https://tools.ietf.org/html/rfc8017#section-3.2) - disregard the comments about "private key exponent", [`RSA_get0_factors`](https://manpages.debian.org/experimental/libssl-doc/RSA_get0_factors.3ssl.en.html) does return the primes) from the same PEM file, and populate the `TPM2B_SENSITIVE` with it

So overall they pull in those 3 pieces of data and then the TPM is responsible for deriving all the rest. I think it's nice and efficient, tbh, though not entirely to standard 😁 Also, I did actually try to create a PEM with OpenSSL, load it in the SW TPM and sign something, then verify with OpenSSL and it's all ok, so sanity-test passed.

The code for loading the key in `tpm2_loadexternal` is [here](https://github.com/tpm2-software/tpm2-tools/blob/f50f4ce3b2d8edd3474229003a4036327f52f421/lib/tpm2_openssl.c#L1060)

src/abstraction/transient.rs
```diff
@@ -179,6 +180,105 @@ impl TransientKeyContext {
...
+    /// # Errors
+    /// * errors are returned if any method calls return an error.
+    /// `TransientKeyContext::`set_session_attrs`, `Context::load_external`,
+    /// `Context::context_save`, `Context::flush_context`
+    pub fn load_external_rsa(
+        &mut self,
+        private_key: &[u8],
+        public_modulus: &[u8],
+        exponent: u32,
+    ) -> Result<TpmsContext> {
```
ionut-arm, line 202: I'm wondering what the best method would be to make this generic. At the moment it seems all keys are loaded as signing keys for RSA PKCS 1v5 with SHA256, but ideally this would be configurable. We can start by just hardcoding things, but that probs needs some detailing in the documentation.

src/abstraction/transient.rs
```diff
@@ -179,6 +180,105 @@ impl TransientKeyContext {
...
+    ///
+    /// # Errors
+    /// * errors are returned if any method calls return an error.
+    /// `TransientKeyContext::`set_session_attrs`, `Context::load_external`,
+    /// `Context::context_save`, `Context::flush_context`
+    pub fn load_external_rsa(
+        &mut self,
+        private_key: &[u8],
+        public_modulus: &[u8],
+        exponent: u32,
```
ionut-arm, line 201: This should be `public_exponent` 😄

ionut-arm (commented):
Thanks! I've left a few comments, I think we need to decide whether we want this method to be more generic (like `create_key`) or if we want to only support some very specific use cases for now

src/abstraction/transient.rs
joechrisellis: Agreed, detailed it in the documentation. I can address this in a new patch at some point if you want. 🙂

src/abstraction/transient.rs
```diff
@@ -179,6 +180,105 @@ impl TransientKeyContext {
...
+
+        // Handle public key...
+        let mut public_key_buffer = [0_u8; 512];
+        public_key_buffer[..public_modulus.len()]
+            .clone_from_slice(&public_modulus[..public_modulus.len()]);
+
+        let pub_buffer = TPM2B_PUBLIC_KEY_RSA {
+            size: public_modulus.len().try_into().unwrap(),
+            buffer: public_key_buffer,
+        };
```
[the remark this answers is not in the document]
joechrisellis: I don't think we can, at least not right now -- slightly below, when we create the `pub_id_union` variable, it expects a `Box<TPM2B_PUBLIC_KEY_RSA>`, so I don't think there's any point wrapping it with that type right now. 🙂

src/abstraction/transient.rs
```diff
@@ -179,6 +180,105 @@ impl TransientKeyContext {
...
+    ///   bytes (i.e. slice elements) long for public keys of length 128, 256, 384 or 512 respectively (which in turn correspond
+    ///   to 512/1024/1536/2048 bits for `private_key`).
+    ///
+    /// # Errors
+    /// * errors are returned if any method calls return an error.
+    /// `TransientKeyContext::`set_session_attrs`, `Context::load_external`,
+    /// `Context::context_save`, `Context::flush_context`
+    pub fn load_external_rsa(
+        &mut self,
+        private_key: &[u8],
```
ionut-arm, line 199: This should probs be `key_prime` or `prime_p` or something along those lines - `private_key` might be misinterpreted by someone not familiar with TPMs

src/abstraction/transient.rs
```diff
@@ -179,6 +180,105 @@ impl TransientKeyContext {
         Ok(key_context)
     }
 
+    /// Load a previously generated RSA keypair. Currently only supports signing keys for RSA PKCS 1v5 with SHA256.
+    ///
+    /// Returns the key context.
+    ///
+    /// # Constraints
+    /// * `public_key` must be 128, 256, 384 or 512 bytes (i.e. slice elements) long, corresponding to 1024, 2048, 3072 or 4096 bits
```
ionut-arm, line 188:
```suggestion
    /// * `public_modulus` must be 128, 256, 384 or 512 bytes (i.e. slice elements) long, corresponding to 1024, 2048, 3072 or 4096 bits
```

ionut-arm (approved):
Looks good overall 👌

src/abstraction/transient.rs
joechrisellis: Good point!

tests/transient_key_context.rs
```diff
@@ -142,6 +164,33 @@ fn verify() {
...
+        0xF6, 0x94, 0x95, 0x35, 0x2F, 0x2A, 0xB5, 0x8D, 0xB8, 0x9A, 0x0A, 0x6D, 0xDB, 0x06, 0x0C,
+        0xA0, 0xBA, 0xA5, 0xEC, 0x19, 0x0D, 0x1D, 0x61, 0xF0, 0xFA, 0xE3, 0x2C, 0xDF, 0xB7, 0x51,
+        0x6F, 0xC9, 0xE4, 0x96, 0x8B, 0x5C, 0x49, 0x4C, 0x05, 0x7F, 0x35, 0xDF, 0xE6, 0x91, 0x36,
+        0xFE, 0x35, 0x43, 0x4F, 0x0A, 0x3B, 0x89, 0x79, 0x55, 0x13, 0x47, 0xC4, 0x7A, 0x35, 0x7A,
+        0xBA, 0xD0, 0xAD, 0x0B,
+    ];
+
+    let mut ctx = create_ctx();
+    // exponent doesn't really matter here...
+    let _key_context = ctx.load_external_rsa(&priv_key, &pub_key, 0).unwrap();
```
hug-dev, line 191: I guess `0` is allowed here because of (section 12.2.3.5 `TPMS_RSA_PARMS` of Part 2 "Structures" of the spec):

> A TPM compatible with this specification and supporting RSA shall support two primes and an exponentof zero. An exponent of zero indicates that the exponent is the default of (2^16)+1. Support for other values is  optional.  Use  of  other  exponents  in  duplicated  keys  is  not  recommended  because  the  resulting  keys would not be interoperable with other TPMs.

Shouldn't we handle that better here? As in, either:
* not having an exponent paramater at all
* indicating in the docs that 0 is a valid value and means that default will be selected (I think this is the minimum we should do)
* replace it with an `Option`? Or some other new type?

ionut-arm: Tbh I think we should document it and leave it there, we probs should support that being something other than 2^16 + 1 . I don't see `Option` as a better fit because `None` doesn't make more sense than `0` - and I think it's quite standard to regard `0` as that particular value by default. PSA Crypto generates all RSA keys with 2 ^ 16 + 1 public exponent, but I assume accepts any value for imports.

Superhepper: An optional RsaExponent type wouldn't be that bad:

```rust
pub struct RsaExponent {
    value: u32,
}

impl Default for RsaExponent {
    fn default() -> RsaExponent{
        RsaExponent {value: 65537 }
    }
}

impl RsaExponent {
    pub fn new(value: u32) -> RsaExponent {
        RsaExponent { value }
    }

    pub fn value(&self) -> u32 {
        self.value
    }
}
```

And then in the api:
```rust
/// If no exponent is provided then it the default value will be used.
/// -->make fancy link to to RsaExponent default <--
    pub fn load_external_rsa(
        &mut self,
        key_prime: &[u8],
        public_modulus: &[u8],
        public_exponent: Option<RsaExponent>,
    ) -> Result<TpmsContext> {
.
.
.
        let rsa_parms = TpmsRsaParmsBuilder::new_unrestricted_signing_key(
            AsymSchemeUnion::RSASSA(HashingAlgorithm::Sha256),
            u16::try_from(public_modulus.len()).unwrap() * 8_u16,
            public_exponent.unwrap_or_default(),
        )
.
.
.
}
```

And then in the call:
```rust
let _key_context = ctx.load_external_rsa(&priv_key, &pub_key, None).unwrap();
```

I am not sure if I completely missed the point but to me this looks nice.

hug-dev: Looks nice to me as well! I agree to both having an `Option` and passing `None` or just having the exponent so that callers directly pass `RsaExponent::default()`.

ionut-arm: I like the suggestion with the new type, but I still think I prefer not having an `Option` - `RsaExponent::default()` or `Default::default()` seems like a good fallback

Superhepper: I have been convinced. New type and using default() is the best option.

Superhepper (approved):

src/abstraction/transient.rs
```diff
@@ -21,14 +21,38 @@ use crate::tss2_esys::*;
...
 
+#[derive(Copy, Clone, Debug)]
+pub struct RsaExponent {
+    value: u32,
+}
+
+impl Default for RsaExponent {
+    fn default() -> RsaExponent {
+        RsaExponent {
+            value: (1 << 16) - 1,
```
hug-dev, line 41: Isn't that `0xFFF`? Do we want to put `0x101` here?

hug-dev (approved):
Thanks for the changes, looks good to me!

PR #125, opened 2020-09-01
Add wrappers for Esys_EvictControl, Esys_Clear, Esys_TR_FromTPMPublic

pajowu:
This PR also switches to using the ESYS_TR instead of the TPM2_RH in calls to TPM2_Hash. See tpm2-software/tpm2-tss#1796 for more details

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -1214,7 +1214,7 @@ impl Context {
                 self.sessions.2,
                 &in_data,
                 hashing_algorithm.into(),
-                hierarchy.rh(),
+                hierarchy.esys_rh(),
                 &mut out_hash_ptr,
                 &mut validation_ptr,
             )
--- a/src/context.rs
+++ b/src/context.rs
@@ -1523,6 +1523,74 @@ impl Context {
             Err(ret)
         }
     }
+
+    /// Make transient objects persistent or remove persistent objects
+    pub fn evict_control(
+        &mut self,
+        auth: ESYS_TR,
+        handle: ESYS_TR,
+        persistent_handle: TPMI_DH_PERSISTENT,
+    ) -> Result<ESYS_TR> {
+        let mut out_handle: ESYS_TR = Default::default();
+        let ret = unsafe {
+            Esys_EvictControl(
+                self.mut_context(),
+                auth,
+                handle,
+                self.sessions.0,
+                self.sessions.1,
+                self.sessions.2,
+                persistent_handle,
+                &mut out_handle,
+            )
+        };
+        let ret = Error::from_tss_rc(ret);
+        if ret.is_success() {
+            Ok(out_handle)
+        } else {
+            Err(ret)
+        }
+    }
+
+    /// Reset the TPM
+    pub fn clear(&mut self, auth_handle: ESYS_TR) -> Result<()> {
+        let ret = unsafe {
+            Esys_Clear(
+                self.mut_context(),
+                auth_handle,
+                self.sessions.0,
+                self.sessions.1,
+                self.sessions.2,
+            )
+        };
+        let ret = Error::from_tss_rc(ret);
+        if ret.is_success() {
+            Ok(())
+        } else {
+            Err(ret)
+        }
+    }
+
+    /// Create ESYS_TR from tpm metadata
+    pub fn from_tpm_public(&mut self, handle: ESYS_TR) -> Result<ESYS_TR> {
+        let mut out_handle: ESYS_TR = Default::default();
+        let ret = unsafe {
+            Esys_TR_FromTPMPublic(
+                self.mut_context(),
+                handle,
+                self.sessions.0,
+                self.sessions.1,
+                self.sessions.2,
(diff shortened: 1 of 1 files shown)


ionut-arm:
Hi! Thanks for the contribution, we'll probably get #120 merged first and then you can update this PR to use the new types defined there. Should make for a Rustier experience 😄

Superhepper:
The change of handle type for hash wont work. That will break backward compatibility. So unless we change the minimum required version to be latest tpm2-tss 2.4.x version then we cannot change the handle types to be used that easily. That is also why the CI fails. Because you are trying to use a handle that is not supported by the 2.3.3 version that is used for CI and that is the minimum required version.

src/context.rs
```diff
@@ -1214,7 +1214,7 @@ impl Context {
                 self.sessions.2,
                 &in_data,
                 hashing_algorithm.into(),
-                hierarchy.rh(),
+                hierarchy.esys_rh(),
```
Superhepper, line 1217: The change of handle type for hash wont work. That will break backward compatibility. So unless we change the minimum required version to be latest tpm2-tss 2.4.x version then we cannot change the handle types to be used that easily. That is also why the CI fails. Because you are trying to use a handle that is not supported by the 2.3.3 version that is used for CI and that is the minimum required version.

src/context.rs
```diff
@@ -1523,6 +1523,74 @@ impl Context {
             Err(ret)
         }
     }
+
+    /// Make transient objects persistent or remove persistent objects
+    pub fn evict_control(
+        &mut self,
+        auth: ESYS_TR,
```
Superhepper, line 1530: Use the Auth type for the auth parameter.

src/context.rs
```diff
@@ -1523,6 +1523,74 @@ impl Context {
...
+        handle: ESYS_TR,
```
Superhepper, line 1531: Check the handle types that we have and see any of the specific handle types are applicable otherwise use the ObjectHandle type.

src/context.rs
```diff
@@ -1523,6 +1523,74 @@ impl Context {
...
+        let ret = Error::from_tss_rc(ret);
+        if ret.is_success() {
+            Ok(())
+        } else {
+            Err(ret)
+        }
+    }
+
+    /// Create ESYS_TR from tpm metadata
+    pub fn from_tpm_public(&mut self, handle: ESYS_TR) -> Result<ESYS_TR> {
```
Superhepper, line 1575: This api already exists and is called tr_from_tpm_public

src/context.rs
```diff
@@ -1523,6 +1523,74 @@ impl Context {
...
+        let ret = Error::from_tss_rc(ret);
+        if ret.is_success() {
+            Ok(out_handle)
+        } else {
+            Err(ret)
+        }
+    }
+
+    /// Reset the TPM
+    pub fn clear(&mut self, auth_handle: ESYS_TR) -> Result<()> {
```
Superhepper, line 1556: Use the AuthHandle type for the auth_handle parameter.

src/context.rs
```diff
@@ -1523,6 +1523,74 @@ impl Context {
...
         }
     }
+
+    /// Make transient objects persistent or remove persistent objects
+    pub fn evict_control(
+        &mut self,
+        auth: ESYS_TR,
+        handle: ESYS_TR,
+        persistent_handle: TPMI_DH_PERSISTENT,
+    ) -> Result<ESYS_TR> {
```
Superhepper, line 1533: Should probably return am ObjectHandle here.

Superhepper (changes requested):
Thanks for the contribution. I left some comments.

src/context.rs
ionut-arm: > 
> 
> The change of handle type for hash wont work. That will break backward compatibility. So unless we change the minimum required version to be latest tpm2-tss 2.4.x version then we cannot change the handle types to be used that easily. That is also why the CI fails. Because you are trying to use a handle that is not supported by the 2.3.3 version that is used for CI and that is the minimum required version.

I'm wondering if a solution here would be to have features for the different version "sets", and set one of them by default. I.e. we'd have features like `tss_v2.3.x` and `tss_v2.4.x` and for each one we handle things differently. Would mean more testing, but that's fine by me, as long as we support a broader range of TSS lib versions.

Superhepper: Yes, I think that is actually a fantastic idea!
```
tss-esapi = {"4.0.7-alpha.1", features = ["min_compat_tss_v2.4.1"]}
```
Or something like that
 
We just need to be careful that we document all the different available compatibility versions somewhere.

puiterwijk:
@pajowu thank you very much for your patches! I just realized that all the wrappers you are adding have already been added in other patches. Shall we close this for now?

Superhepper:
Has Esys_Clear really been added?

Superhepper:
All the changes have been merged in other PR:s so this can be closed.

PR #127, opened 2020-09-02
Upgrade dependencies

hug-dev:

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,40 @@
+name: Nightly Checks
+
+on:
+  schedule:
+    # Every night at midnight
+    - cron:  '0 0 * * *'
+
+jobs:
+  # This won't build on GA environment because of missing TSS
+  # libraries. To be reactivated later or to pass using Docker.
+  dependencies:
+    name: Check for unused dependencies
+    if: false
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install latest Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly
+      - name: Fetch submodules
+        run: git submodule update --init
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -28,8 +28,8 @@ zeroize = { version = "1.1.0", features = ["zeroize_derive"] }
 env_logger = "0.7.1"
 
 [build-dependencies]
-bindgen = "0.54.0"
-pkg-config = "0.3.17"
+bindgen = "0.55.1"
```
hug-dev, line 31: `bindgen` can not be upgraded to `0.55.1` because it pulls in `clang-sys` version `1.0.0` which itself pulls in `libloading` version `0.6.3` which uses the `non_exhaustive` attribute which is not supported by `rustc` version `1.38.0` that we test on CI.
If we decide to increase the version of `bindgen` it means that our minimum supported Rust compiler would be [`1.40.0`](https://blog.rust-lang.org/2019/12/19/Rust-1.40.0.html#non_exhaustive-structs-enums-and-variants)

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,40 @@
...
+      - name: Install latest Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly
+      - name: Fetch submodules
+        run: git submodule update --init
+      - name: Install cargo udeps
+        run: cargo install cargo-udeps --locked
+      - name: Execute cargo udeps
+        run: cargo +nightly udeps
```
ionut-arm, line 26: I just noticed - do you need the `+nightly` if you specify `with: toolchain: nightly` above?

hug-dev: Hmm yeah probably not, I just copy-pasted the command to make sure it worked 🤠

ionut-arm (approved):

PR #128, opened 2020-09-03
Add zeroing to SensitiveData

### LINKED ISSUE #46: Implement secure data management
Currently all data passed to/through the library is released as-is once it is dropped. We should ideally clean up all the objects we use and all input we get from clients.

See https://github.com/parallaxsecond/parsec/issues/122 for more details.

hug-dev:
Also zeroize the authentication value after setting it.

This would fix #46

### DIFF
--- a/src/algorithm/structures/sensitive/data.rs
+++ b/src/algorithm/structures/sensitive/data.rs
@@ -4,10 +4,12 @@ use crate::tss2_esys::TPM2B_SENSITIVE_DATA;
 use crate::{Error, Result, WrapperErrorKind};
 use log::error;
 use std::convert::TryFrom;
+use zeroize::Zeroize;
 /// Rust native representation of sensitive data.
 ///
 /// The structure contains the sensitive data as a byte vector.
-#[derive(Clone, Debug, Eq, PartialEq)]
+#[derive(Clone, Debug, Eq, PartialEq, Zeroize)]
+#[zeroize(drop)]
 pub struct SensitiveData {
     value: Vec<u8>,
 }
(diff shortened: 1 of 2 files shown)


hug-dev:
Many buffer containing confidential/sensitive data are implemented with the `buffer_type` macro which wraps them around `Zeroizing`:
* `Auth`
* `Data`
* `PublicKeyRSA`
* `Digest`
* `MaxNvBuffer`

I am not sure about the response codes and the PCR values (`PcrData`): should they also be zeroized?

Superhepper (approved):
Nicely done!

Superhepper:
> Many buffer containing confidential/sensitive data are implemented with the `buffer_type` macro which wraps them around `Zeroizing`:
> 
>     * `Auth`
> 
>     * `Data`
> 
>     * `PublicKeyRSA`
> 
>     * `Digest`
> 
>     * `MaxNvBuffer`
> 
> 
> I am not sure about the response codes and the PCR values (`PcrData`): should they also be zeroized?

PCR values are not that sensitive. Because the pcr values cannot be set directly in the TPM. You can modify them using pcr_extend and such. But you cannot set them to a specific value (or you can but that would require extreme luck or an almost infinite amount of "trial and horror").

They are more sayng "this is the state of your computer" and if you use them as authorization then you are saying "your computer needs to be in this exact state in order to access the secret"

Still I do not think it would hurt to do it because the performance hit would not be that bad.

hug-dev:
I see, thanks for the explanation. If that is not that critical for now, I would say we can go with this PR and add zeroizing on Pcr structures later.

ionut-arm (approved):

Issue #126, opened 2020-09-01
Depoly Session


Superhepper:
In #120 Session was added. This now needs to be deployed.

Superhepper:
I think we have deployed the handles in most places now.

PR #129, opened 2020-09-17
Deployed Session and Handles in all places where they made sense.

Superhepper:
Made use of the Session and Handles that were introduced in #120 as described in #126.

### DIFF
--- a/src/abstraction/transient.rs
+++ b/src/abstraction/transient.rs
@@ -16,7 +16,7 @@
 use crate::constants::algorithm::{Cipher, EllipticCurve, HashingAlgorithm};
 use crate::constants::tss::*;
 use crate::constants::types::session::SessionType;
-use crate::handles::SessionHandle;
+use crate::handles::KeyHandle;
 use crate::structures::{Auth, Data, Digest, PublicKeyRSA, VerifiedTicket};
 use crate::tcti::Tcti;
 use crate::tss2_esys::*;
@@ -69,7 +69,7 @@ impl RsaExponent {
 #[derive(Debug)]
 pub struct TransientKeyContext {
     context: Context,
-    root_key_handle: ESYS_TR,
+    root_key_handle: KeyHandle,
 }
 
 impl TransientKeyContext {
@@ -121,11 +121,11 @@ impl TransientKeyContext {
         let key_handle = self.context.load(self.root_key_handle, key_priv, key_pub)?;
 
         self.set_session_attrs()?;
-        let key_context = self.context.context_save(key_handle).or_else(|e| {
-            self.context.flush_context(key_handle)?;
+        let key_context = self.context.context_save(key_handle.into()).or_else(|e| {
+            self.context.flush_context(key_handle.into())?;
             Err(e)
         })?;
-        self.context.flush_context(key_handle)?;
+        self.context.flush_context(key_handle.into())?;
         Ok((key_context, key_auth))
     }
 
@@ -197,11 +197,11 @@ impl TransientKeyContext {
             .context
             .load_external_public(&public, Hierarchy::Owner)?;
         self.set_session_attrs()?;
-        let key_context = self.context.context_save(key_handle).or_else(|e| {
-            self.context.flush_context(key_handle)?;
+        let key_context = self.context.context_save(key_handle.into()).or_else(|e| {
+            self.context.flush_context(key_handle.into())?;
             Err(e)
         })?;
-        self.context.flush_context(key_handle)?;
+        self.context.flush_context(key_handle.into())?;
         Ok(key_context)
     }
 
@@ -298,11 +298,11 @@ impl TransientKeyContext {
             .context
             .load_external(&private, &public, Hierarchy::Null)?;
         self.set_session_attrs()?;
-        let key_context = self.context.context_save(key_handle).or_else(|e| {
-            self.context.flush_context(key_handle)?;
+        let key_context = self.context.context_save(key_handle.into()).or_else(|e| {
+            self.context.flush_context(key_handle.into())?;
             Err(e)
         })?;
-        self.context.flush_context(key_handle)?;
+        self.context.flush_context(key_handle.into())?;
         Ok(key_context)
     }
 
@@ -319,7 +319,7 @@ impl TransientKeyContext {
         let key_handle = self.context.context_load(key_context)?;
 
         self.set_session_attrs()?;
-        let key_pub_id = self.context.read_public(key_handle).or_else(|e| {
+        let key_pub_id = self.context.read_public(key_handle.into()).or_else(|e| {
             self.context.flush_context(key_handle)?;
             Err(e)
         })?;
@@ -362,12 +362,15 @@ impl TransientKeyContext {
         label: Option<Data>,
     ) -> Result<PublicKeyRSA> {
         self.set_session_attrs()?;
-        let key_handle = self.context.context_load(key_context)?;
+        let key_handle = self
+            .context
+            .context_load(key_context)
+            .map(KeyHandle::from)?;
         if let Some(key_auth_value) = key_auth {
             self.context
                 .tr_set_auth(key_handle.into(), &key_auth_value)
                 .or_else(|e| {
-                    self.context.flush_context(key_handle)?;
+                    self.context.flush_context(key_handle.into())?;
                     Err(e)
                 })?;
         }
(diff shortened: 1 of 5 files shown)


src/context.rs
```diff
@@ -673,17 +791,26 @@ impl Context {
...
                 self.mut_context(),
-                key_handle,
-                self.sessions.0,
-                self.sessions.1,
-                self.sessions.2,
+                key_handle.into(),
+                self.sessions
+                    .0
+                    .map(|v| v.handle().into())
+                    .unwrap_or(ESYS_TR_NONE),
```
hug-dev, line 805: Might be worth to create a method somewhere to do the conversion `Option<Session>` -> `ESYS_TR`.

src/abstraction/transient.rs
```diff
@@ -362,12 +362,15 @@ impl TransientKeyContext {
         label: Option<Data>,
     ) -> Result<PublicKeyRSA> {
         self.set_session_attrs()?;
-        let key_handle = self.context.context_load(key_context)?;
+        let key_handle = self
+            .context
+            .context_load(key_context)
+            .map(KeyHandle::from)?;
```
hug-dev, line 368: Does it work if you just add an `into` at the end?
```rust
let key_handle = self.context.context_load(key_context)?.into();
```

hug-dev (commented):
Thank you! More abstractions 😄 !!

src/abstraction/transient.rs
Superhepper: I am not sure if the compiler can figure out what convert into. I think rust requires you to be specific. The following would work (I think):
```rust
let key_handle: KeyHandle = self.context.context_load(key_context).into();
```
or 
```rust
let key_handle = self.context.context_load(key_context).into::<KeyHandle>();
```

src/context.rs
Superhepper: Hrrmm...maybe I am not sure what I think about that. I have to think about that.

hug-dev: Note that my point is only to avoid code duplication as opposed as esthetics ;)

src/abstraction/transient.rs
hug-dev: Yes that what I meant! You decide if you prefer that rather than calling `map`!

ionut-arm (commented):
Thank for the effort, this is great!!! I think the question on a conversion to `ESYS_TR` is still pending, so I'm just waiting for your choice on that

src/context.rs
Superhepper: That things are just aliases for an integer type is something you will probably never get away from when the code is written in C unless it is written by a zealous programmer with a fetish for types. So conversions to and from ESYS_TR is something that is done for most handles and session types.

To be able to do:
```rust
self.sessions.0.into()
```
instead of 
```rust
self.sessions.0.map(|v| v.handle().into()).unwrap_or(ESYS_TR_NONE)
```

So how it is used inside the Context methods is of little interest really because that is an implementation detail that should not concern users unless there is a bug. Something that is more interesting is how it would be used used by some one that is using the lib and then suddenly need to value of the handle.

OBS! Pseudo code.
```rust
// bla bla code
let mut auth_session: Option<Session> = None;
if need_to_auth {
    auth_session = context.start_auth_session()
}

let actual_value: ESYS_TR = auth_session.into();
```

That is how I would expect it to work if I didn't know better. Which speaks in favor for adding conversions from Option<Session>. So I think I may be inclined into adding that conversion.

ionut-arm: Thanks!! That sounds good

Superhepper: Unfortunatly ```From<Option<Session>> for ESYS_TR``` is not possible. Because neither Option nor ESYS_TR are defined in the crate.

I have been going at this for a little while now and more I try the worse the code gets.

Superhepper:
I have added some different ways of conversion in Session. One thing they have in common is that they provide a way to get the SessionHandle from the Session or Option<Session. That handle can in turn be converted into ESYS_TR. The reason for this is because the handles are or first layer of abstraction of the ESYS_TR. So it wouldn't make any sense to convert a Session directly into a ESYS_TR.

So I would be really grateful for some inputs into which ones we should keep.

src/context.rs
ionut-arm: Yeah, I didn't realise that initially! Thanks for attempting 😅

ionut-arm:
> 
> 
> I have added some different ways of conversion in Session. One thing they have in common is that they provide a way to get the SessionHandle from the Session or `Option<Session>`. That handle can in turn be converted into ESYS_TR. The reason for this is because the handles are or first layer of abstraction of the ESYS_TR. So it wouldn't make any sense to convert a Session directly into a ESYS_TR.

I would say it's fine to even have the `ESYS_TR` hidden completely (i.e. get it with a method that's public only within the crate) and not allow users of the abstract version of the crate to get it from `SessionHandle`. Unless you think there are cases where one might need to access stuff we've not abstracted yet?

> 
> I wonder if we could do something more clever here with a deref?

Do you mean deref'ing from `SessionHandle` to `ESYS_TR` or from `Session` to `SessionHandle`? I think in either case we should only use typical conversions, as `Deref` seems to be intended for smart pointers which is not exactly what we have here.

Superhepper:
Hrrm I do not think we are there quite yet were we can completely remove the possibility of get hold of the ESYS_TR. 

I think that needs to be done at a later stage. Because that involves making ESAPI, the Esys types and all the conversion to esys types only internally available.

src/session.rs
```diff
@@ -12,34 +12,82 @@ pub enum Session {
...
-            Session::Policy(session_object) => session_object.handle(),
-            Session::Trial(session_object) => session_object.handle(),
+    ///
+    /// Function that creates a Option<Session>
+    ///
+    pub fn create(
+        session_type: SessionType,
+        session_handle: SessionHandle,
+        auth_hash: HashingAlgorithm,
+    ) -> Option<Session> {
```
ionut-arm, line 26: Is there any doc explaining why we use `Option`? It feels like we should mention that we map `ESYS_TR_NONE` to `Option::None` throughout the crate

src/context.rs
```diff
@@ -127,64 +127,64 @@ impl Context {
...
         let ret = unsafe {
             Esys_StartAuthSession(
                 self.mut_context(),
-                tpm_key,
+                tpm_key.map(|v| v.into()).unwrap_or(ESYS_TR_NONE),
                 bind,
-                self.sessions.0,
-                self.sessions.1,
-                self.sessions.2,
+                Session::handle(self.sessions.0).into(),
```
ionut-arm, line 149: Maybe I'm a bit late on this one 🤡  you could implement 3 methods on context:
```Rust
    fn session_0(&self) -> ESYS_TR {
        Session::handle(self.sessions.0).into()
    }
```
and use that everywhere, like we use the `self.mut_context()`

src/session.rs
Superhepper: Yes, I will add a detail section that explains that.

src/context.rs
Superhepper: Yes! That is much better. I think I will remove the ```From``` implementations. Because I do not like them and just keep static handle function that takes ```Option<Session>``` and maybe add another method to Session that retrieves can retrieve the handle.

src/context.rs
```diff
@@ -127,64 +131,64 @@ impl Context {
     #[allow(clippy::too_many_arguments)]
     pub fn start_auth_session(
         &mut self,
-        tpm_key: ESYS_TR,
+        tpm_key: Option<KeyHandle>,
         bind: ESYS_TR,
```
ionut-arm, line 135: Could this also be replaced with `Option<ObjectHandle>`?

ionut-arm (approved):
LGTM!

src/context.rs
Superhepper: Probably I just need to do some digging in the specification to see what it should be...

ionut-arm: > The  entity  referenced  with  the bind parameter  contributes  an  authorization  value  to  the sessionKeygeneration process.
>
> entity providing the authValue; may be TPM_RH_NULL

I think it can be any entity that sports a handle

hug-dev (approved):

src/context.rs
Superhepper: Yeah seams as you could send in anything that handle that refers to something that some kind of auth but the question is should None => ESYS_TR_NONE or ESYS_TR_RH_NULL?

Superhepper: Nevermind it should be none. Yeah, lets do that.

ionut-arm (approved):

PR #131, opened 2020-10-01
Implement pcr_{reset, extend} function wrappers

puiterwijk:


src/structures/lists/digest.rs
```diff
@@ -1,11 +1,72 @@
...
+    }
+}
+
+impl TryFrom<DigestValues> for TPML_DIGEST_VALUES {
+    type Error = Error;
+    fn try_from(digest_values: DigestValues) -> Result<Self> {
+        let mut tss_digest_values: TPML_DIGEST_VALUES = Default::default();
+        for (digest_hash, digest_val) in digest_values.digests.iter() {
+            tss_digest_values.digests[tss_digest_values.count as usize] =
+                (*digest_hash, digest_val.clone()).try_into()?;
```
ionut-arm, line 63: The `clone` here can be avoided since `digest_values` is dropped at the end of this method anyway, you can call `digest_values.digests.drain()` above and get ownership of the hashmap contents.

Superhepper (changes requested):

src/structures/lists/digest.rs
```diff
@@ -1,11 +1,72 @@
...
+        }
+    }
+
+    pub fn set(&mut self, alg: HashingAlgorithm, dig: Digest) {
+        let _ = self.digests.insert(alg, dig);
+    }
+}
+
+impl TryFrom<(HashingAlgorithm, Digest)> for TPMT_HA {
+    type Error = Error;
```
Superhepper, line 29: I like this part but I think it should be placed in a separate module.

In ```src/structures``` is where we have placed most of our rust representations of the TPM types (i.g. Name which is our way of representing TPM2B_NAME can be found in ```src/structures/names/name.rs```). So I would suggest you do ```src/structures/hash_digest.rs``` and put in your code there for your hash digest there.

puiterwijk: Thanks for the note. My intention had indeed been to split `DigestValues` out, but I forgot to do that before committing!

ionut-arm (approved):
LGTM 👍

Superhepper (approved):
Looks nice!

PR #132, opened 2020-10-01
Add method to use custom TCTI during test

puiterwijk:
This adds an environment variable (TEST_TCTI) that gets parsed via
`Tcti::from_str` to run the tests with another TCTI, for example when
Mssim is not available on a system.

### DIFF
--- a/tests/context_tests.rs
+++ b/tests/context_tests.rs
@@ -33,6 +33,8 @@ const KEY: [u8; 512] = [
 ];
 
 use std::convert::{TryFrom, TryInto};
+use std::env;
+use std::str::FromStr;
 use tss_esapi::{
     algorithm::structures::SensitiveData,
     constants::{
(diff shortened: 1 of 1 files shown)


ionut-arm (approved):

tests/context_tests.rs
```diff
@@ -103,7 +104,14 @@ fn create_public_sealed_object() -> tss_esapi::tss2_esys::TPM2B_PUBLIC {
 }
 
 fn create_ctx_without_session() -> Context {
-    unsafe { Context::new(Tcti::Mssim(Default::default())).unwrap() }
+    let tcti = match env::var("TEST_TCTI") {
+        Err(_) => Tcti::Mssim(Default::default()),
+        Ok(tpe) => match &tpe[..] {
```
hug-dev, line 109: I think it would be excellent to use the `FromStr` trait on `Tcti` here! The implementation is [here](https://github.com/parallaxsecond/rust-tss-esapi/blob/d55f3fca59d1c2ac40d49ff371732440c8b1c2f8/src/tcti.rs#L71), and I believe it uses the standard format 🚀

puiterwijk: Updated, I had totally forgotten about that!

hug-dev (approved):
Thanks for the change!

PR #133, opened 2020-10-02
Add startup method wrapper

puiterwijk:

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -6,7 +6,7 @@ use crate::{
         algorithm::{Cipher, HashingAlgorithm},
         tags::PropertyTag,
         tss::{TPMA_SESSION_DECRYPT, TPMA_SESSION_ENCRYPT},
-        types::{capability::CapabilityType, session::SessionType},
+        types::{capability::CapabilityType, session::SessionType, startup::StartupType},
     },
     handles::{
         handle_conversion::TryIntoNotNone, AuthHandle, KeyHandle, NvIndexHandle, ObjectHandle,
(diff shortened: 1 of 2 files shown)


src/context.rs
```diff
@@ -115,6 +115,16 @@ impl Context {
         }
     }
 
+    pub fn startup(&mut self, startup_type: TPM2_SU) -> Result<()> {
+        let ret = unsafe { Esys_Startup(self.mut_context(), startup_type) };
```
Superhepper, line 119: There is a Startup type already defined in src/constants/types.rs so you do not need to use the ESYS type in the api.

puiterwijk: Ah, right. But this function is going to be troublesome with the tests anyway (given that swtpm already runs a startup, and doesn't like having it rerun, as a normal TPM should), so I may very well just drop this PR.

ionut-arm: We do actually have to run the CLI version of this command before running tests: https://github.com/parallaxsecond/rust-tss-esapi/blob/master/tests/all.sh#L18 . Plus, even without a test, the method might be useful on `Context`

Superhepper: Yeah, I do not think it should be closed. I think this method needs to be available even without a test.

puiterwijk:
By popular demand, reopened this PR (sorry about closing it).
I rebased and made it use the `StartupType` as requested.

ionut-arm (approved):
Nice and simple

src/context.rs
```diff
@@ -130,6 +130,16 @@ impl Context {
...
     }
 
+    pub fn startup(&mut self, startup_type: StartupType) -> Result<()> {
+        let ret = unsafe { Esys_Startup(self.mut_context(), startup_type.into()) };
+        let ret = Error::from_tss_rc(ret);
+        if ret.is_success() {
+            Ok(())
+        } else {
+            Err(ret)
+        }
```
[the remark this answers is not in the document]
puiterwijk: Added!

Superhepper (approved):

PR #136, opened 2020-10-26
Add Context.execute_with_session(s) functions

puiterwijk:
These are convenience functions to help with executing sections of code (Rust closures)
with a specific set of sessions.
This can avoid a lot of manual handling of sessions, and helps keep the session
use of each function clear (like shown in the PCR test case).

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -194,6 +194,45 @@ impl Context {
         self.sessions
     }
 
+    /// Execute the closure in f with the specified set of sessions, and sets the original sessions back afterwards
+    /// This is a convenience function
+    pub fn execute_with_sessions<F, T>(
+        &mut self,
+        session_handles: (Option<Session>, Option<Session>, Option<Session>),
+        f: F,
+    ) -> T
+    where
+        // We only need to call f once, so it can be FnOnce
+        F: FnOnce(&mut Context) -> T,
+    {
+        let oldses = self.sessions();
+        self.set_sessions(session_handles);
+
+        let res = f(self);
+
+        self.set_sessions(oldses);
+
+        res
+    }
+
+    /// A special case of execute_with_sessions. Executes the closure with a single session set, and the others set to None
+    pub fn execute_with_session<F, T>(&mut self, session_handle: Option<Session>, f: F) -> T
+    where
+        // We only need to call f once, so it can be FnOnce
+        F: FnOnce(&mut Context) -> T,
+    {
+        self.execute_with_sessions((session_handle, None, None), f)
+    }
+
+    /// A special case of execute_with_sessions. Executes the closure without any sessions
+    pub fn execute_without_session<F, T>(&mut self, f: F) -> T
+    where
+        // We only need to call f once, so it can be FnOnce
+        F: FnOnce(&mut Context) -> T,
+    {
+        self.execute_with_sessions((None, None, None), f)
+    }
+
     /// Get current capability information about the TPM.
     pub fn get_capabilities(
         &mut self,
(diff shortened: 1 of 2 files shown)


tests/context_tests.rs
```diff
@@ -366,28 +366,26 @@ mod test_pcr_extend_reset {
...
-        context.clear_sessions();
         // Read PCR contents
         let pcr_selection_list = PcrSelectionListBuilder::new()
             .with_selection(HashingAlgorithm::Sha1, &[PcrSlot::Slot16])
             .with_selection(HashingAlgorithm::Sha256, &[PcrSlot::Slot16])
             .build();
-        let (_, _, pcr_data) = context.pcr_read(&pcr_selection_list).unwrap();
+        // pcr_read is NO_SESSIONS
+        let (_, _, pcr_data) =
+            context.execute_with_session(None, |ctx| ctx.pcr_read(&pcr_selection_list).unwrap());
```
ionut-arm, line 381: Given that some calls are `NO_SESSIONS`, does it make sense to have an `execute_with_no_sessions` method as well?

tests/context_tests.rs
```diff
@@ -366,28 +366,26 @@ mod test_pcr_extend_reset {
         // get reset by any D-RTPM events.
         // PCR (TCG PC Client Platform TPM Profile (PTP) for TPM 2.0 Version 1.05 Rev 14)
         let mut context = create_ctx_with_session();
-        let pcr_ses = context.sessions();
+        let pcr_ses = context.sessions().0;
 
         // We start by resetting. We do not place any expectations on the prior contents
-        context.pcr_reset(PcrHandle::Pcr16).unwrap();
+        context.execute_with_session(pcr_ses, |ctx| ctx.pcr_reset(PcrHandle::Pcr16).unwrap());
```
ionut-arm, line 372: Looks like you're only using this `pcr_ses` session now, wouldn't it be easier to simply set it as the main session on `Context` and use the new methods only where you need a different one?

ionut-arm (commented):
Thanks for the change 👌

tests/context_tests.rs
puiterwijk: I personally prefer to use the `execute_with_session` everywhere, to make it explicit which sessions are used with which calls.

tests/context_tests.rs
puiterwijk: Sure, I had exactly that in my own copy, will put that in here too.

tests/context_tests.rs
puiterwijk: If you prefer to use the `pcr_ses` by default, and only `execute_with_session` in the special case, I'll happily change. This is just my personal preference of explicit over implicit.

ionut-arm: I don't mind either way, it's only a test! It does help with checking that the thing works properly 😄 So leave as is!

src/context.rs
```diff
@@ -194,6 +194,45 @@ impl Context {
         self.sessions
     }
 
+    // Execute the closure in f with the specified set of sessions, and sets the original sessions back afterwards
+    // This is a convenience function
+    pub fn execute_with_sessions<F, T>(
```
ionut-arm, line 199: Could you make this proper Rustdoc comments?

ionut-arm (approved):
Looks good to me, just one comment below

Superhepper:
Nice! looks goof to me. They feel a lot like it would have come from the Rust std which is a good thing!

Superhepper (approved):

Issue #135, opened 2020-10-12
Add evict_control API to context


Superhepper:
We need to add support for the evict control api. I think I will do this while doing the interface types.

ionut-arm:
The Rust KeyLime team (of which @puiterwijk is a part) is looking at implementing wrappers for [a bunch of the API calls](https://github.com/keylime/rust-keylime/issues/89), you can sync with them if you plan to upstream this - `evict_control` is one of those functions

Superhepper:
I split this up. So that interface types were in their own PR. Now I work on the evict_control PR.

Issue #134, opened 2020-10-12
Implement interface types


Superhepper:
There is a need to implement the interface types(TPMI). These types are used in a lot of the API:s and are important if we want others to be able to develop new api:s

Interface types(TPMI) are used to indicate what kind of permanent handles that can be used in certain API calls. In the NV-functions there is a 'hack'(NvAuthorization) that needs to be removed. One of these interface types that are used a lot is the Hierarchy.

One major thing with this issue is to figure out how the interface types relates to the handle types that have already been implemented.

ionut-arm:
This sounds like a nice problem to tackle, I'll have a look at the spec for those types as well to get a feel on how they work and what they're used for in relation to the handles.

Superhepper:
Thanks. I really appricate that.

Superhepper:
One thing I have not been able to figure out is what DH stands for? TPMI_RH I assumed was 'resource handle'. Could TPMI_DH be 'data handle'?

ionut-arm:
It seems possible, but looking at a few of the `TPMI_DH` types it doesn't feel right:
> TPMI_DH_PARENT
The  TPMI_DH_PARENT  interface  type  is  a  handle  that  references  an  object  that  can  be  the  parent  of another  object.  The  handles  in  this  set  may  refer  to  either  transient  or  persistent  object  or  to Primary Seeds.

Wouldn't say that's a data handle. It seems the `RH` ones reference types that have either one or a small number of possible values - something like an enum. But what `R` means is beyond me. I hate the absolute obscurity around these naming conventions - none of them are explained anywhere. Presumably Part 1 of the spec contains abbreviations, but they only put in the effort to give the abbreviations for generic stuff like `I/O` - thanks, I'm sure you can't just figure that one out with a quick search.

Superhepper:
**TCG TPM2 r1p59 Part1 Architecture, Section 15.5**
>Fixed resource handles refer to Shielded Locations that are always associated with the same handle.
>These resources have handles with an MSO of 40 16 . Examples of these resources are Owner, Platform,
>and Endorsement hierarchy controls and the Lockout authorization value.

To me that statement seams to support my assumption that RH stands for resource handle.

Superhepper:
RH handles seam to refer to constants. These cannot be changed. They are not even implementation dependent!

DH refers to a set or sets of handles. Each DH 'type' one or several different sets of handles are valid. And these ranges are implementation dependent!!
[...]

PR #137, opened 2020-10-27
Add interface types

Superhepper:
Interface types are used in almost al API:s that uses TPM handles and are therefore essential to implement.

#134 #135


Superhepper:
I would like some comments on this. I think this is they I will do it. I just need add some more tests for evict_control API.

ionut-arm:
> I would like some comments on this. I think this is they I will do it. I just need add some more tests for evict_control API.

Yes, apologies, will have a look after lunch!

src/interface_types/resource_handles.rs
```diff
@@ -0,0 +1,371 @@
...
+impl TryFrom<TPM2_HANDLE> for Hierarchy {
+    type Error = Error;
+
+    fn try_from(handle: TPM2_HANDLE) -> Result<Self> {
+        match handle {
+            TPM2_RH_OWNER | ESYS_TR_RH_OWNER => Ok(Hierarchy::Owner),
+            TPM2_RH_PLATFORM | ESYS_TR_RH_PLATFORM => Ok(Hierarchy::Platform),
+            TPM2_RH_ENDORSEMENT | ESYS_TR_RH_ENDORSEMENT => Ok(Hierarchy::Endorsement),
+            TPM2_RH_NULL | ESYS_TR_RH_NULL => Ok(Hierarchy::Null),
+            _ => Err(Error::local_error(WrapperErrorKind::InconsistentParams)),
```
ionut-arm, line 60: This should probabily be `InvalidParams` (and similarly for the other `TryFrom`s)

src/context.rs
```diff
@@ -811,6 +817,41 @@ impl Context {
         }
     }
 
+    /// Evicts persistent objects or allows certain transient objects
+    /// to be made peristent.
+    pub fn evict_control(
```
ionut-arm, line 822: Any chance you could pull the `evict_control` part into its own PR? There's a lot of changes around the handles and this doesn't directly fit in

src/interface_types/resource_handles.rs
```diff
@@ -0,0 +1,371 @@
...
+        }
+    }
+}
+//////////////////////////////////////////////////////////////////////////////////
+/// Owner
+//////////////////////////////////////////////////////////////////////////////////
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Owner {
+    Owner,
+    Null, // TODO: Remove null and use Option instead?
```
ionut-arm, line 194: Regarding the TODO, I think keeping a `Null` variant is better, since it's named like that in the spec. Might be confusing to handle it differently here just because it's called `Null`

src/interface_types/resource_handles.rs
```diff
@@ -0,0 +1,371 @@
...
+        }
+    }
+}
+
+//////////////////////////////////////////////////////////////////////////////////
+// AttachedComponent
+//
+// TODO: Figure out how to implement this and how this should relate
+//       AttachedComponent enum value in crate::handles::tpm::TpmHandle.
+//////////////////////////////////////////////////////////////////////////////////
```
ionut-arm, line 365: The most sensible solution seems to be implementing `AC_FIRST:AC_LAST` as wrappers around a numeric value and checking that the value is in the correct range when constructing. It seems the type for `NV_INDEX_FIRST:NV_INDEX_LAST` (which I assume is `NvIndexHandle`) doesn't have checks for what values are valid or not.

ionut-arm (commented):
Wow, this is a big batch of changes! But I'm glad to see us drifting closer and closer towards a fully Rusty interface.

Left some comments below to discuss

src/interface_types/resource_handles.rs
Superhepper: No the reason for this is because those values have been commented out in the constants file:

```src/constants/tss.rs```
```rust
// the following constants cannot be instantiated since the TPM2_HT constants are u8
//pub const TPM2_HR_PCR               : TPM2_HC = TPM2_HT_PCR << TPM2_HR_SHIFT;
//pub const TPM2_HR_HMAC_SESSION      : TPM2_HC = TPM2_HT_HMAC_SESSION << TPM2_HR_SHIFT;
//pub const TPM2_HR_POLICY_SESSION    : TPM2_HC = TPM2_HT_POLICY_SESSION << TPM2_HR_SHIFT;
//pub const TPM2_HR_TRANSIENT         : TPM2_HC = TPM2_HT_TRANSIENT << TPM2_HR_SHIFT;
//pub const TPM2_HR_PERSISTENT        : TPM2_HC = TPM2_HT_PERSISTENT << TPM2_HR_SHIFT;
//pub const TPM2_HR_NV_INDEX          : TPM2_HC = TPM2_HT_NV_INDEX << TPM2_HR_SHIFT;
//pub const TPM2_HR_PERMANENT         : TPM2_HC = TPM2_HT_PERMANENT << TPM2_HR_SHIFT;
//pub const TPM2_PCR_FIRST            : TPM2_HC = TPM2_HR_PCR + 0; /* first PCR */
//pub const TPM2_PCR_LAST             : TPM2_HC = TPM2_PCR_FIRST + TPM2_MAX_PCRS - 1; /* last PCR */
//pub const TPM2_HMAC_SESSION_FIRST   : TPM2_HC = TPM2_HR_HMAC_SESSION + 0; /* first HMAC session */
//pub const TPM2_HMAC_SESSION_LAST    : TPM2_HC = TPM2_HMAC_SESSION_FIRST + 0x00fffffe; /* last HMAC session */
//pub const TPM2_LOADED_SESSION_FIRST : TPM2_HC = TPM2_HMAC_SESSION_FIRST; /* used in GetCapability */
//pub const TPM2_LOADED_SESSION_LAST  : TPM2_HC = TPM2_HMAC_SESSION_LAST; /* used in GetCapability */
//pub const TPM2_POLICY_SESSION_FIRST : TPM2_HC = TPM2_HR_POLICY_SESSION + 0; /* first policy session */
//pub const TPM2_POLICY_SESSION_LAST  : TPM2_HC = TPM2_POLICY_SESSION_FIRST + 0x00fffffe; /* last policy session */
//pub const TPM2_TRANSIENT_FIRST      : TPM2_HC = TPM2_HR_TRANSIENT + 0; /* first transient object */
//pub const TPM2_ACTIVE_SESSION_FIRST : TPM2_HC = TPM2_POLICY_SESSION_FIRST; /* used in GetCapability */
//pub const TPM2_ACTIVE_SESSION_LAST  : TPM2_HC = TPM2_POLICY_SESSION_LAST; /* used in GetCapability */
//pub const TPM2_TRANSIENT_LAST       : TPM2_HC = TPM2_TRANSIENT_FIRST + 0x00fffffe; /* last transient object */
//pub const TPM2_PERSISTENT_FIRST     : TPM2_HC = TPM2_HR_PERSISTENT + 0; /* first persistent object */
//pub const TPM2_PERSISTENT_LAST      : TPM2_HC = TPM2_PERSISTENT_FIRST + 0x00FFFFFF; /* last persistent object */
//pub const TPM2_PLATFORM_PERSISTENT  : TPM2_HC = TPM2_PERSISTENT_FIRST + 0x00800000; /* first platform persistent object */
//pub const TPM2_NV_INDEX_FIRST       : TPM2_HC = TPM2_HR_NV_INDEX + 0; /* first allowed NV Index */
//pub const TPM2_NV_INDEX_LAST        : TPM2_HC = TPM2_NV_INDEX_FIRST + 0x00FFFFFF; /* last allowed NV Index */
```

src/context.rs
Superhepper: yeah absolutly

src/interface_types/resource_handles.rs
Superhepper: At this point I agree with you (I am not sure I have made up my mind yet) but the reason I thought about it was because of how we sometimes handle ```ESYS_TR_NONE``` as an Option. So I thought the same thing might apply to these null values as well. But as you say the specification mentions the null value explicitly in this types so it might be confusing to remove them.

src/constants/tss.rs
```diff
@@ -516,32 +516,38 @@ pub const TPM2_HR_HANDLE_MASK: TPM2_HC = 0x00FFFFFF; /* to mask off the HR */
 pub const TPM2_HR_RANGE_MASK: TPM2_HC = 0xFF000000; /* to mask off the variable part */
 pub const TPM2_HR_SHIFT: TPM2_HC = 24;
 // the following constants cannot be instantiated since the TPM2_HT constants are u8
```
ionut-arm, line 518: This can be removed now! Cheers for fixing it 🥳

src/handles/handle.rs
```diff
@@ -263,6 +266,14 @@ pub mod auth {
...
     );
     impl_handle_conversion!(AuthHandle, ObjectHandle);
+    // The following constant handles can be used for authorization
+    // according to the TCG TPM2 r1p59 Structures specification.
+    add_constant_handle!(AuthHandle, OwnerHandle, ESYS_TR_RH_OWNER);
+    add_constant_handle!(AuthHandle, Lockout, ESYS_TR_RH_LOCKOUT);
+    add_constant_handle!(AuthHandle, EndorsementHandle, ESYS_TR_RH_ENDORSEMENT);
+    add_constant_handle!(AuthHandle, PlatformHandle, ESYS_TR_RH_PLATFORM);
+    // TODO: Figure out how to add AUTH_00 to AUTH_FF range
+    // TODO: Figure out how to add ACT_0 to ACT_F range
```
ionut-arm, line 276: Which ones are these? I can see the auth ones ( `#define ESYS_TR_RH_AUTH(x)     (ESYS_TR_RH_AUTH_FIRST +(ESYS_TR)(x))` ) but I don't get what the ACT range is.

ionut-arm (approved):
Looks good.

I have one question about the handles - what's the purpose/role of [`TpmConstantsHandle`](https://github.com/parallaxsecond/rust-tss-esapi/blob/d55f3fca59d1c2ac40d49ff371732440c8b1c2f8/src/handles/handle.rs#L218)? It seems to have variants that overlap with the other handle types and didn't see any particular type in the spec it would correspond to. (I'm aware it's not introduced or touched in this PR, I just stumbled upon it when reviewing)

src/handles/handle.rs
Superhepper: It comes from the very part I explicitly did not wanted to cover in this PR. The permanent handles (**TCG TPM2 r1p59 Part 2 Structures: Chapter 7.4 Permanent Handles**). I chose to just cover the range and leave it at that. But in the future I think they need to be specified further.

Superhepper: If you want to read more about the **Authenticated Countdown Timer** them self look at **TCG TPM2 r1p59 Part 3 Commands: Chapter 33 Authenticated Countdown Timer**.

Superhepper:
> 
> 
> Looks good.
> 
> I have one question about the handles - what's the purpose/role of [`TpmConstantsHandle`](https://github.com/parallaxsecond/rust-tss-esapi/blob/d55f3fca59d1c2ac40d49ff371732440c8b1c2f8/src/handles/handle.rs#L218)? It seems to have variants that overlap with the other handle types and didn't see any particular type in the spec it would correspond to. (I'm aware it's not introduced or touched in this PR, I just stumbled upon it when reviewing)

The TpmConstantsHandles in handles refer to the ESYS version of TPM handle. And I think I just put it there because I saw that the ESYS specification said something about "TPM Constants" or something so I made a type for it. Maybe it should be removed.

src/handles/handle.rs
ionut-arm: Cheers, that's what I was looking for, I'll have a read!

puiterwijk (approved):
This looks good to me.

PR #138, opened 2020-10-28
Add abstraction::nv::read_full to fully read an NV Index

puiterwijk:
It is not possible to read an NV Index in one request if the contents are larger than
the implementation's `TPM_PT_NV_BUFFER`.
This functions repeats over the length of the buffer, to ensure it is possible to read
everything.

### DIFF
--- a/src/abstraction/nv.rs
+++ b/src/abstraction/nv.rs
@@ -1,0 +1,39 @@
+// Copyright 2020 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{
+    constants::tags::PropertyTag,
+    handles::{AuthHandle, NvIndexHandle, NvIndexTpmHandle, TpmHandle},
+    Context, Result,
+};
+
+/// Allows reading an NV Index completely, regardless of the max TPM NV buffer size
+pub fn read_full(
+    context: &mut Context,
+    auth_handle: AuthHandle,
+    nv_index_handle: NvIndexTpmHandle,
+) -> Result<Vec<u8>> {
+    let maxsize = match context.get_tpm_property(PropertyTag::NvBufferMax)? {
+        Some(val) => val,
+        None => 512,
+    } as usize;
+
+    let nv_idx = TpmHandle::NvIndex(nv_index_handle);
+    let nv_idx = context.execute_without_session(|ctx| ctx.tr_from_tpm_public(nv_idx))?;
+    let nv_idx: NvIndexHandle = nv_idx.into();
+
+    let (nvpub, _) = context.execute_without_session(|ctx| ctx.nv_read_public(nv_idx))?;
+    let nvsize = nvpub.data_size();
+
+    let mut result = Vec::new();
+    result.reserve_exact(nvsize);
+
+    for offset in (0..nvsize).step_by(maxsize) {
+        let size: u16 = std::cmp::min(maxsize, nvsize - offset) as u16;
+
+        let res = context.nv_read(auth_handle, nv_idx, size, offset as u16)?;
+        result.extend_from_slice(&res);
+    }
+
+    Ok(result)
+}
--- a/src/constants/tags/property.rs
+++ b/src/constants/tags/property.rs
@@ -1,0 +1,101 @@
+// Copyright 2020 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+use crate::{constants::tss::*, tss2_esys::TPM2_PT, Error, Result, WrapperErrorKind};
+use log::error;
+use num_derive::{FromPrimitive, ToPrimitive};
+use num_traits::{FromPrimitive, ToPrimitive};
+use std::convert::{From, TryFrom};
+
+#[derive(FromPrimitive, ToPrimitive, Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[repr(u32)]
+pub enum PropertyTag {
+    None = TPM2_PT_NONE,
+    // Fixed
+    FamilyIndicator = TPM2_PT_FAMILY_INDICATOR,
+    Level = TPM2_PT_LEVEL,
+    Revision = TPM2_PT_REVISION,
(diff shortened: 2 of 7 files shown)


src/abstraction/nv.rs
```diff
@@ -0,0 +1,37 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{
+    handles::{AuthHandle, NvIndexHandle, NvIndexTpmHandle, TpmHandle},
+    Context, Result,
+};
+
+// Source: TCG PC Client Platform TPM Profile Specification for TPM 2.0
+// Version 1.05, Revision 14, section 4.1, table 1
+const TPM_PT_NV_BUFFER_MAX: usize = 512;
```
ionut-arm, line 11: Two things worth mentioning here:
* there is already a (wrong) definition of this constant [here](https://github.com/parallaxsecond/rust-tss-esapi/blob/master/src/constants/tss.rs#L424) where it follows the Trusted Platform Module Library Part 2: Structures, page 51. I think the value in that `tss.rs` file is actually the property code that can be used with `GetCapability` to actually obtain the true value of `TPM_PT_NV_BUFFER_MAX` from the TPM. I'll raise a bug for this, maybe we should change those `PT`s to an enum to indicate what they're intended for.
* this 512 that you used is for PC clients only (as you stated in the comment). I think there's a need to either mark the function below as only working with PC TPMs or to make it more flexible by using `GetCapability`

src/abstraction/nv.rs
```diff
@@ -0,0 +1,37 @@
+// Copyright 2019 Contributors to the Parsec project.
```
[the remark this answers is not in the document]
puiterwijk: That's what I get for just copy-pasting the copyright lines from another function without reading them.

src/abstraction/nv.rs
puiterwijk: I'd been working on actually abstracting the Capability type, so I may as well just finish that and use that here.

ionut-arm: That would be great, thanks!

puiterwijk: That is now in PR #140, and I'll rebase this PR on top of that once that gets merged.

puiterwijk: I rebase on top of the in-progress #140 and added automatic detection of the max size.

puiterwijk:
This is now based on top of #140

src/context.rs
```diff
@@ -236,10 +239,10 @@ impl Context {
     /// Get current capability information about the TPM.
     pub fn get_capabilities(
         &mut self,
-        capability: TPM2_CAP,
+        capability: CapabilityType,
```
Superhepper, line 242: Nice!

ionut-arm (approved):
Looks good to me, but I'll review again tomorrow

src/context.rs
```diff
@@ -263,16 +266,40 @@ impl Context {
...
+            let capabilitydata = unsafe { MBox::from_raw(outcapabilitydata) };
+            let capabilities = CapabilityData::try_from(*capabilitydata)?;
+            Ok((capabilities, moredata))
         } else {
             Err(ret)
         }
     }
 
+    /// Determine a TPM property, which is automatically cached if available
+    pub fn get_tpm_property(&mut self, property: u32) -> Result<Option<u32>> {
```
Superhepper, line 281: What is that ```property```?

If that is property tag then I would rather have a type or an enum for it like ```StructureTag``` (src/constants/tags/structure.rs) unless it would be an excessive amount of work? If it is to much work then at least do like make a ```PropertyTag ``` type that only wraps a u32 value and provides conversions to and from TPM_PT. And then make a ticket to make a proper implementation of ```PropertyTag```.

The reason I would like it is because this is a public api and we should try to make it hard to use them incorrectly

Superhepper (commented):
Thanks for the change! I just had one comment.

src/context.rs
puiterwijk: Will do.

src/constants/tags/property.rs
```diff
@@ -0,0 +1,101 @@
+// Copyright 2020 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+use crate::{constants::tss::*, tss2_esys::TPM2_PT, Error, Result, WrapperErrorKind};
+use log::error;
+use num_derive::{FromPrimitive, ToPrimitive};
+use num_traits::{FromPrimitive, ToPrimitive};
+use std::convert::{From, TryFrom};
+
+#[derive(FromPrimitive, ToPrimitive, Debug, Clone, Copy, PartialEq, Eq)]
```
Superhepper, line 9: Add Hash here:

I.e. #[derive(FromPrimitive, ToPrimitive, Debug, Clone, Copy, PartialEq, Eq, Hash)]

src/context.rs
```diff
@@ -67,6 +68,8 @@ pub struct Context {
     tcti_context: Option<MBox<TSS2_TCTI_CONTEXT>>,
     /// A set of currently open object handles that should be flushed before closing the context.
     open_handles: HashSet<ObjectHandle>,
+    /// A cache of determined TPM limits
+    cached_tpm_properties: HashMap<u32, u32>,
```
Superhepper, line 72: Change to HashMap<PropertyTag, u32>,

src/context.rs
```diff
@@ -274,6 +278,30 @@ impl Context {
         }
     }
 
+    /// Determine a TPM property, which is automatically cached if available
+    pub fn get_tpm_property(&mut self, property: PropertyTag) -> Result<Option<u32>> {
+        let property: u32 = property.into();
```
Superhepper, line 283: This line will then not be needed if the key in the HashMap is PropertyTag.

Superhepper (approved):

Superhepper:
Everything looks good to me!

ionut-arm (approved):
Sorry for the delay, looks good! Thanks for the change

puiterwijk:
Cool, thanks!

PR #139, opened 2020-10-28
Add abstraction::ek module

puiterwijk:
This adds the public functions `create_ek` and `retrieve_ek_pubcert`,
to create a handle to a created EK object and retrieve the endorsement key
public certificate from the NV index, respectively.

### DIFF
--- a/src/abstraction/ek.rs
+++ b/src/abstraction/ek.rs
@@ -1,0 +1,136 @@
+// Copyright 2020 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{
+    abstraction::nv,
+    constants::{algorithm::AsymmetricAlgorithm, tss::*},
+    handles::{AuthHandle, KeyHandle, NvIndexTpmHandle, TpmHandle},
+    tss2_esys::{
+        ESYS_TR_RH_ENDORSEMENT, TPM2B_ECC_PARAMETER, TPM2B_PUBLIC, TPM2B_PUBLIC_KEY_RSA,
+        TPMS_ECC_PARMS, TPMS_ECC_POINT, TPMS_RSA_PARMS, TPMS_SCHEME_HASH, TPMT_ECC_SCHEME,
+        TPMT_KDF_SCHEME, TPMT_RSA_SCHEME, TPMT_SYM_DEF_OBJECT, TPMU_ASYM_SCHEME, TPMU_SYM_KEY_BITS,
+        TPMU_SYM_MODE,
+    },
+    utils::{ObjectAttributes, PublicIdUnion, PublicParmsUnion, Tpm2BPublicBuilder},
+    Context, Result,
+};
+
+// Source: TCG EK Credential Profile for TPM Family 2.0; Level 0 Version 2.3 Revision 2
+// Section 2.2.1.4 (Low Range) for Windows compatibility
+const RSA_2048_EK_CERTIFICATE_NV_INDEX: u32 = 0x01c00002;
+const ECC_P256_EK_CERTIFICATE_NV_INDEX: u32 = 0x01c0000a;
+
+// Source: TCG EK Credential Profile for TPM Family 2.0; Level 0 Version 2.3 Revision 2
+// Appendix B.3.3 and B.3.4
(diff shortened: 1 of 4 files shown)


puiterwijk:
I am going to guess that the TPM in the test does not get issued with an EK cert. I'll look at adding that to the test server initialization.

src/abstraction/ek.rs
```diff
@@ -0,0 +1,171 @@
...
+
+    // TPM2_PolicySecret(TPM_RH_ENDORSEMENT)
+    // With 32 null-bytes attached, because of the type of with_auth_policy
+    let authpolicy: [u8; 64] = [
+        0x83, 0x71, 0x97, 0x67, 0x44, 0x84, 0xb3, 0xf8, 0x1a, 0x90, 0xcc, 0x8d, 0x46, 0xa5, 0xd7,
+        0x24, 0xfd, 0x52, 0xd7, 0x6e, 0x06, 0x52, 0x0b, 0x64, 0xf2, 0xa1, 0xda, 0x1b, 0x33, 0x14,
+        0x69, 0xaa, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+        0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+        0x00, 0x00, 0x00, 0x00,
+    ];
```
ionut-arm, line 72: How was this computed?

src/abstraction/ek.rs
```diff
@@ -0,0 +1,171 @@
...
+            })))
+            .build(),
+    }
+}
+
+/// Create the Endorsement Key object from the specification templates
+pub fn create_ek_object(context: &mut Context, alg: AsymmetricAlgorithm) -> Result<KeyHandle> {
+    let ek_public = create_ek_public_from_default_template(alg)?;
+
+    let ses = create_nullauth_session(context)?;
```
ionut-arm, line 141: I see in the function below you flush the context of the `nullauth` session - is there no need to do that here?

ionut-arm (commented):
Thanks! Looks good overall, left a couple of comments.

I'm guessing from your comments that this is intended specifically for TPM usage with Windows platforms?

puiterwijk:
> I'm guessing from your comments that this is intended specifically for TPM usage with Windows platforms?

No, it's just the case that most TPMs have their Endorsement Certs in the locations that Windows asks for.

src/abstraction/ek.rs
puiterwijk: That is the TPM policy digest value of a clear session where `TPM2_PolicySecret(TPM_RH_ENDORSEMENT)` was executed (theoretically).
This value comes straight from the TCG EK Credential Profile for TPM Family 2.0; Level 0 Version 2.3 Revision 2 spec.

src/abstraction/ek.rs
puiterwijk: Yep, that was needed. I just added the `execute_with_nullauth_session` to `Context`, since there may be more cases where you need to run something that's `TPM_ST_SESSIONS` without actually needing to do authentication.

src/context.rs
```diff
@@ -237,6 +238,36 @@ impl Context {
...
+    where
+        // We only need to call f once, so it can be FnOnce
+        F: FnOnce(&mut Context) -> Result<T>,
+    {
+        let session = match self.start_auth_session(
+            None,
+            None,
+            None,
+            SessionType::Hmac,
+            Cipher::aes_256_cfb(),
```
ionut-arm, line 252: Just FYI - in the future this might need to be made configurable (maybe storing a "default cipher" on `Context`), because not all TPMs support 256 bit AES. We've had problems with that for Parsec

ionut-arm (approved):
Looks good!

src/context.rs
puiterwijk: I might just set it to 128 bit AES then for now.
This is a nullauth session, so not used for any actual authentication anyway.

ionut-arm (approved):

Superhepper (approved):
Looks good to me

PR #141, opened 2020-11-03
Implement utils::get_tpm_vendor

puiterwijk:
This function will return the name of the TPM vendor as provided by the
TPM VendorString{1, 2, 3, 4} properties.

### DIFF
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -9,10 +9,11 @@
 //! type name. Unions are converted to Rust `enum`s by dropping the `TPMU` qualifier and appending
 //! `Union`.
 use crate::constants::algorithm::{Cipher, EllipticCurve, HashingAlgorithm};
+use crate::constants::tags::PropertyTag;
 use crate::constants::tss::*;
 use crate::structures::{Digest, PcrSlot};
 use crate::tss2_esys::*;
-use crate::{Error, Result, WrapperErrorKind};
+use crate::{Context, Error, Result, WrapperErrorKind};
 use bitfield::bitfield;
 use enumflags2::BitFlags;
 use log::error;
@@ -1515,4 +1516,37 @@ impl From<PcrData> for TPML_DIGEST {
         }
         tpml_digest
     }
+}
+
+fn tpm_int_to_string(num: u32) -> String {
+    num.to_be_bytes()
+        .iter()
+        .filter(|x| **x != 0)
+        .map(|x| char::from(*x))
+        .collect()
+}
+
+/// Get the TPM vendor name
+pub fn get_tpm_vendor(context: &mut Context) -> Result<String> {
+    // Retrieve the TPM property values
+    Ok([
+        PropertyTag::VendorString1,
+        PropertyTag::VendorString2,
+        PropertyTag::VendorString3,
+        PropertyTag::VendorString4,
+    ]
+    .iter()
+    // Retrieve property values
+    .map(|propid| context.get_tpm_property(*propid))
+    // Collect and return an error if we got one
+    .collect::<Result<Vec<Option<u32>>>>()?
+    .iter()
+    // Filter out the Option::None values
+    .filter_map(|x| *x)
+    // Filter out zero values
+    .filter(|x| *x != 0)
+    // Map through int_to_string
+    .map(tpm_int_to_string)
+    // Collect to a single string
+    .collect())
 }
--- a/tests/utils_tests.rs
+++ b/tests/utils_tests.rs
@@ -1,0 +1,20 @@
+// Copyright 2020 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::{env, str::FromStr};
+use tss_esapi::{utils, Context, Tcti};
+
+pub fn create_ctx_without_session() -> Context {
+    let tcti = match env::var("TEST_TCTI") {
+        Err(_) => Tcti::Mssim(Default::default()),
+        Ok(tctistr) => Tcti::from_str(&tctistr).expect("Error parsing TEST_TCTI"),
+    };
+    unsafe { Context::new(tcti).unwrap() }
+}
+
+#[test]
+fn get_tpm_vendor() {
+    let mut context = create_ctx_without_session();
+
+    utils::get_tpm_vendor(&mut context).unwrap();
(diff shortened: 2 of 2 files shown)


src/utils/mod.rs
```diff
@@ -1516,3 +1517,47 @@ impl From<PcrData> for TPML_DIGEST {
...
+        if num == 0 {
+            break;
+        }
+    }
+    result
+}
+
+/// Get the TPM vendor name
+pub fn get_tpm_vendor(context: &mut Context) -> Result<String> {
+    //let vendor_properties = [PropertyTag::VendorString1, PropertyTag::VendorString2, PropertyTag::VendorString3, PropertyTag::VendorString4];
```
ionut-arm, line 1541: Was this supposed to be left here?

src/utils/mod.rs
```diff
@@ -1516,3 +1517,47 @@ impl From<PcrData> for TPML_DIGEST {
...
+        if chr != 0 {
+            let chr = char::from(chr);
+            result.insert(0, chr);
+        }
+        if num == 0 {
+            break;
+        }
+    }
+    result
+}
```
ionut-arm, line 1537: I think this is covered by [`ToString`](https://doc.rust-lang.org/std/primitive.u32.html#impl-ToString) (?)

Superhepper: Are you sure?

**TCG TPM2 r1p59 Part 2 Structures**
>|Capability Name| Value | Comments|
>|-------------|-------------|-------------|
>|TPM_PT_VENDOR_STRING_1| PT_FIXED + 6| <p>the first four characters of the vendor ID string<br>NOTE: <br>When the vendor string is fewer than 16 octets, the additional property values do not have to be present. A vendor string of 4 octets can be represented in one 32-bit value and no null terminating character is required.</p>|
>|TPM_PT_VENDOR_STRING_2|PT_FIXED + 7|the second four characters of the vendor ID string|
>|TPM_PT_VENDOR_STRING_3|PT_FIXED + 8|the third four characters of the vendor ID string|
>|TPM_PT_VENDOR_STRING_4|PT_FIXED + 9|the fourth four characters of the vendor ID sting|

ionut-arm: Hmm, having a 2nd look the function I linked is definitely not what was intended, probably [this](https://doc.rust-lang.org/std/str/fn.from_utf8.html) is more appropriate. I think the section you quote is more relevant to the function below, that actually handles the full vendor string - this seems to be intended mostly as a `u32` -> `String` conversion, by treating the 4 octets as chars.

Superhepper: I do not like that loop. It should be ensured to only use the bytes in the value.

I would must more prefer it looking something like this (OBS psedo code)
```rust
let mut result = String::new();
for byte  in num.as_be_bytes() {
    if byte == 0 {
       break;
    }
    result.push(char::from(byte));
}
result
```

puiterwijk: Yep, I just wrote some code like that last version an hour ago, but was in calls so couldn't submit. Will do that shortly :).

Superhepper: > Hmm, having a 2nd look the function I linked is definitely not what was intended, probably [this](https://doc.rust-lang.org/std/str/fn.from_utf8.html) is more appropriate. I think the section you quote is more relevant to the function below, that actually handles the full vendor string - this seems to be intended mostly as a `u32` -> `String` conversion, by treating the 4 octets as chars.

Yes, the interesting part there was the note section. I am not sure the from_utf8 would work either because the specification says the string is in ASCII (but I might be wrong?).

ionut-arm: I think `from_utf8` should work - ASCII is a subset of UTF8. What I understand from that comment on `TPM_PT_VENDOR_STRING_1` is that the other vendor string values might be 0, in which case they shouldn't be included in the string, but those first 4 octets should be non-null. Maybe? I'm not sure

puiterwijk: Yeah, the null values are already prevented by 
```
    // Filter out the empty values
    .filter_map(|x| *x)
    // Filter out zero values
    .filter(|x| *x != 0)
```

puiterwijk: But I think I'm going with @Superhepper 's approach, given that interpreting it as UTF8 while it's really just individual ASCII characters seems not the best.

src/utils/mod.rs
puiterwijk: Nope, so I removed it.

src/utils/mod.rs
ionut-arm: I'm ok with that approach, though [`char::from(u8)`](https://doc.rust-lang.org/std/primitive.char.html#impl-From%3Cu8%3E) isn't really ASCII either, it's just compatible with.

tests/utils_tests.rs
```diff
@@ -0,0 +1,20 @@
...
+        Ok(tctistr) => Tcti::from_str(&tctistr).expect("Error parsing TEST_TCTI"),
+    };
+    unsafe { Context::new(tcti).unwrap() }
+}
+
+#[test]
+fn get_tpm_vendor() {
+    let mut context = create_ctx_without_session();
+
+    utils::get_tpm_vendor(&mut context).unwrap();
```
ionut-arm, line 19: Is the value obtained here constant (since it's using the same simulator every time)?

ionut-arm (approved):
LGTM

tests/utils_tests.rs
puiterwijk: If it uses the same simulator, yes. But when I run the tests on my personal system, it says something else, and I think it might not be ideal to fail the test just because I'm running it on my system instead of the CI.

src/utils/mod.rs
```diff
@@ -1516,3 +1517,37 @@ impl From<PcrData> for TPML_DIGEST {
         tpml_digest
     }
 }
+
+fn tpm_int_to_string(num: u32) -> String {
+    u32::from_be(num)
+        .to_ne_bytes()
```
Superhepper, line 1523: Sorry for being a little bit stupid when it comes to endians. But why couldn't you do ```num.to_be_bytes()```?

Superhepper (approved):
LGTM!

PR #142, opened 2020-11-05
Add support for SWTPM network TPM

puiterwijk:
This TCTI got added in tpm2-tss 3.0, but given that we just pass the TCTI config string as-is to tss2, this should build fine on older TSS versions, just not actually work.

Testing against that on `fedora:latest` results in all tests individually passing.
(When I run all the tests in one go, it is exhausting the session context space allotted.)

Right now, the link with more information about configuration is still pointing to the `Mssim_Init` function, because it take the same configuration, but the `Swtpm_Init` function is not yet documented online as far as I can find.

### DIFF
--- a/src/tcti.rs
+++ b/src/tcti.rs
@@ -10,6 +10,7 @@ use std::str::FromStr;
 
 const DEVICE: &str = "device";
 const MSSIM: &str = "mssim";
+const SWTPM: &str = "swtpm";
 const TABRMD: &str = "tabrmd";
 
 // Possible TCTI to use with the ESYS API.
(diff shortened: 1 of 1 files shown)


puiterwijk:
As discussed on Slack, it seems the session problem seems to be us leaking sessions. I'll be looking into debugging that, and then add a test based on swtpm/fedora.

ionut-arm (approved):
Looks good!

This kind of low-level conversion makes me a bit shivery, but hopefully it can't get exploited in any nasty way (I know I wrote it)

Superhepper (approved):
LGTM!

PR #145, opened 2020-11-06
Added support for tpm2-tss 3

Superhepper:
- Made custom build step provide the major version of the tpm2-tss lib
  as suggested by @puiterwijk [here](https://github.com/parallaxsecond/rust-tss-esapi/pull/145#issuecomment-725917955)
- Changed some Context methods to convert the Hierarchy to either
  a TpmHandle or ObjectHandle depending on the tpm2-tss major version.
- Added the permamnent handles that can be found in the specification to PermenentTpmHandle.
- Added more constant handles to ObjectHandle.
- Made it possible to convert Hierarchy, HierarchyAuth and Enables to and from ObjectHandle and Tpmhandle.

### DIFF
--- a/src/handles/handle.rs
+++ b/src/handles/handle.rs
@@ -1,9 +1,14 @@
 // Copyright 2020 Contributors to the Parsec project.
 // SPDX-License-Identifier: Apache-2.0
 
-///
+//! ESYS handle types
+//!
+//! The ESAPI specification has only one handle type
+//! ESYS_TR. That type is wrapped by ObjectHandle but there
+//! are also specific handle types that indicates what created
+//! them or how they are intended to be used.
+
 /// Macro for generating a basic handle implementation
-///
 macro_rules! impl_basic_handle {
     (
         $(#[$outer:meta])*
@@ -40,10 +45,8 @@ macro_rules! impl_basic_handle {
     };
 }
 
-///
 /// Macro for making a esys constant available
 /// for a handle type.
-///
 macro_rules! add_constant_handle {
     ($handle_type:ident, $constant_handle_name:ident, $constant_handle_value:ident) => {
         impl $handle_type {
@@ -55,33 +58,34 @@ macro_rules! add_constant_handle {
     };
 }
 
-///
-/// Object handle module
-///
+/// Module for the ObjectHandle
 pub mod object {
     use crate::tss2_esys::{
         ESYS_TR_NONE, ESYS_TR_PASSWORD, ESYS_TR_RH_ENDORSEMENT, ESYS_TR_RH_LOCKOUT,
         ESYS_TR_RH_NULL, ESYS_TR_RH_OWNER, ESYS_TR_RH_PLATFORM, ESYS_TR_RH_PLATFORM_NV,
     };
+
     impl_basic_handle!(
-        /// General handle type.
+        /// The ObjectHandle is the general handle type
+        /// and it wraps the ESYS_TR.
+        ///
+        /// All the other more specific handle types can be
+        /// converted into an ObjectHandle.
         ObjectHandle
     );
 
-    add_constant_handle!(ObjectHandle, PasswordHandle, ESYS_TR_PASSWORD);
-    add_constant_handle!(ObjectHandle, NoneHandle, ESYS_TR_NONE);
-    add_constant_handle!(ObjectHandle, OwnerHandle, ESYS_TR_RH_OWNER);
-    add_constant_handle!(ObjectHandle, LockoutHandle, ESYS_TR_RH_LOCKOUT);
-    add_constant_handle!(ObjectHandle, EndorsementHandle, ESYS_TR_RH_ENDORSEMENT);
-    add_constant_handle!(ObjectHandle, PlatformHandle, ESYS_TR_RH_PLATFORM);
-    add_constant_handle!(ObjectHandle, PlatformNvHandle, ESYS_TR_RH_PLATFORM_NV);
-    add_constant_handle!(ObjectHandle, NullHandle, ESYS_TR_RH_NULL);
+    add_constant_handle!(ObjectHandle, Password, ESYS_TR_PASSWORD);
+    add_constant_handle!(ObjectHandle, None, ESYS_TR_NONE);
+    add_constant_handle!(ObjectHandle, Owner, ESYS_TR_RH_OWNER);
+    add_constant_handle!(ObjectHandle, Lockout, ESYS_TR_RH_LOCKOUT);
+    add_constant_handle!(ObjectHandle, Endorsement, ESYS_TR_RH_ENDORSEMENT);
+    add_constant_handle!(ObjectHandle, Platform, ESYS_TR_RH_PLATFORM);
+    add_constant_handle!(ObjectHandle, PlatformNv, ESYS_TR_RH_PLATFORM_NV);
+    add_constant_handle!(ObjectHandle, Null, ESYS_TR_RH_NULL);
 }
 
-///
 /// Macro for creating ESYS_TR conversion for
 /// constant handle types
-///
 macro_rules! impl_basic_multiple_constant_values_handle {
     ($constant_handle_type:ident) => {
         impl From<$constant_handle_type> for ESYS_TR {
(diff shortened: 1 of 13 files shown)


Superhepper:
@ionut-arm @puiterwijk 

If Fedora is running version > 2.0 of tpm2-tss then maybe something like this would work?

ionut-arm:
I wonder if we could somehow use `pkg_config` or something like that to determine at runtime what library we're interacting with.

hug-dev:
Or still at buildtime but without needing an extra feature, you can emit a config flag from the build script that you use in the code depending on which version you detect with `pkg-config`.

ionut-arm:
> Or still at buildtime but without needing an extra feature, you can emit a config flag from the build script that you use in the code depending on which version you detect with `pkg-config`.

Doesn't that imply that you're always compiling just before using?

hug-dev:
Yes true then it would not work if you compile this crate for a machine with an old version and try it on a newer version one. I am not sure if this is common.

ionut-arm:
I'm trying to be conscious about this crate being used for a different application that is compiled, packaged and distributed in some centralised manner, so you end up running on systems where you weren't even compiled. Plus, the build script is only run when something in the code changed, I believe, so the env var gets set only the first time

Superhepper:
Well as code looks now the check is done at run-time. It should not be that difficult to get the values from pkg-config. Will have a look at it when I get the time.

ionut-arm:
> 
> 
> Well as code looks now the check is done at run-time. It should not be that difficult to get the values from pkg-config. Will have a look at it when I get the time.

Not sure if the check is done at runtime or not (the compiler might strip away the code from the "other" branch), but whether the feature is enabled or not is decided at compile time.

Superhepper:
> 
> 
> > Well as code looks now the check is done at run-time. It should not be that difficult to get the values from pkg-config. Will have a look at it when I get the time.
> 
> Not sure if the check is done at runtime or not (the compiler might strip away the code from the "other" branch), but whether the feature is enabled or not is decided at compile time.

Hmm, I only though it did that if you put the cfg check on like module or function level. If you but the cfg check in an if statement I just thought it converted it to a runtime check. I wonder where I got that idea from...

ionut-arm (approved):

puiterwijk:
I would personally prefer to just get this automatically detected based on the pkg-config version information instead.
That way we don't need to document this, and it should work automatically for everyone without having incorrect behaviour by forgetting to set the feature on build-time.

As I already said on Slack, I think that people that build it against v2 and then run it against v3 should not happen, and people should be recompiling it against the new version of the library.

puiterwijk:
@Superhepper Maybe something like https://github.com/puiterwijk/rust-tss-esapi/commit/af15343ba13e6b56c879faa5c41c97277cf8b296 ?
I just took your `context.rs` changes, but with an automatically detected TSS version.

Superhepper:
> @Superhepper Maybe something like [puiterwijk@af15343](https://github.com/puiterwijk/rust-tss-esapi/commit/af15343ba13e6b56c879faa5c41c97277cf8b296) ?
> I just took your `context.rs` changes, but with an automatically detected TSS version.

That looks really nice actually! 

The only thing that I would like to change is to remove esys_rh and make that a ```From``` conversion for it. And then use ```.into()``` in all places except where the ```tpm2_tss_2``` flag indicates that we need to use ```.rh()```.

puiterwijk:
> That looks really nice actually!
> 
> The only thing that I would like to change is to remove esys_rh and make that a `From` conversion for it. And then use `.into()` in all places except where the `tpm2_tss_2` flag indicates that we need to use `.rh()`.

Yep, fair enough. I just meant to show it as a suggestion for the automatic feature detection, I just took the rest of the code from your PR :).
Please feel free to take the auto-detection and build on it!

Superhepper:
> > That looks really nice actually!
> > The only thing that I would like to change is to remove esys_rh and make that a `From` conversion for it. And then use `.into()` in all places except where the `tpm2_tss_2` flag indicates that we need to use `.rh()`.
> 
> Yep, fair enough. I just meant to show it as a suggestion for the automatic feature detection, I just took the rest of the code from your PR :).
> Please feel free to take the auto-detection and build on it!

I will absolutely do that! Thanks a lot for testing this.

Superhepper:
So now the version is picked up from PKG-CONFIG and used in order to determine if the Hierarchy should be converted into a TpmHandle or EsysHandle.

I also added the Permanent handles that is specified in the specification in order to be able to convert Hierarchy into a TpmHandle. I also added handle constants to the ObjectHandle in order for hierarchy to be converted into an ObjectHandle.

I think it became rather elegant this way.

puiterwijk (changes requested):
I'm a bit worried about the mix of the tss3 support with more type reorganization in the same commit.
It makes it much harder to find exactly what caused a bug later on, and additionally makes reviewing harder.
Could you make the type changes a separate commit please?

Superhepper:
I have split the PR into two commits now.

src/handles/handle.rs
```diff
@@ -59,14 +59,23 @@ macro_rules! add_constant_handle {
...
     );
 
     add_constant_handle!(ObjectHandle, PasswordHandle, ESYS_TR_PASSWORD);
     add_constant_handle!(ObjectHandle, NoneHandle, ESYS_TR_NONE);
+    add_constant_handle!(ObjectHandle, OwnerHandle, ESYS_TR_RH_OWNER);
+    add_constant_handle!(ObjectHandle, LockoutHandle, ESYS_TR_RH_LOCKOUT);
+    add_constant_handle!(ObjectHandle, EndorsementHandle, ESYS_TR_RH_ENDORSEMENT);
+    add_constant_handle!(ObjectHandle, PlatformHandle, ESYS_TR_RH_PLATFORM);
+    add_constant_handle!(ObjectHandle, PlatformNvHandle, ESYS_TR_RH_PLATFORM_NV);
+    add_constant_handle!(ObjectHandle, NullHandle, ESYS_TR_RH_NULL);
```
ionut-arm, line 78: Should these be named like the other ones, i.e. `Password` or `PasswordSession`, dropping the `Handle`?

src/handles/handle.rs
```diff
@@ -59,14 +59,23 @@ macro_rules! add_constant_handle {
 /// Object handle module
 ///
 pub mod object {
-    use crate::tss2_esys::{ESYS_TR_NONE, ESYS_TR_PASSWORD};
+    use crate::tss2_esys::{
+        ESYS_TR_NONE, ESYS_TR_PASSWORD, ESYS_TR_RH_ENDORSEMENT, ESYS_TR_RH_LOCKOUT,
+        ESYS_TR_RH_NULL, ESYS_TR_RH_OWNER, ESYS_TR_RH_PLATFORM, ESYS_TR_RH_PLATFORM_NV,
+    };
     impl_basic_handle!(
         /// General handle type.
```
ionut-arm, line 67: Not necessarily for this PR, but maybe this should mention that it's a wrapper for `ESYS_TR` handles.

ionut-arm (commented):
Overall looks good to me! That's a lot of enums with the same subset of handles as variants 😔

src/handles/handle.rs
Superhepper: Yes, it should be mentioned. It is on my TODO list when it comes to improving the documentation.

src/handles/handle.rs
Superhepper: What other ones? ```const``` handles for both the different Esys handles and Tpm handles have the Handle-suffix.

ionut-arm: Indeed, I confused it with the types in `resource_handle.rs` - `HierarchyAuth`, `Enables` etc.

That being said, you choose which version you prefer - we've tended to exclude the suffix/prefix if it's already included in the type name.

Superhepper: Added a commit where I removed all the Handle suffixes.

Superhepper:
@puiterwijk Can you have another look and see if your requested changes have been addressed?

puiterwijk (approved):
Looks good to me

PR #146, opened 2020-11-10
Added evict_control context method

Superhepper:

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -9,9 +9,13 @@ use crate::{
         types::{capability::CapabilityType, session::SessionType},
     },
     handles::{
-        AuthHandle, KeyHandle, NvIndexHandle, ObjectHandle, PcrHandle, SessionHandle, TpmHandle,
+        AuthHandle, KeyHandle, NvIndexHandle, ObjectHandle, PcrHandle, PersistentTpmHandle,
+        SessionHandle, TpmHandle,
     },
-    interface_types::resource_handles::{Hierarchy, NvAuth},
+    interface_types::{
+        dynamic_handles::Persistent,
+        resource_handles::{Hierarchy, NvAuth, Provision},
+    },
     nv::storage::NvPublic,
     session::Session,
     structures::{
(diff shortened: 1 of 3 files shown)


src/context.rs
```diff
@@ -874,6 +878,41 @@ impl Context {
...
+    pub fn evict_control(
+        &mut self,
+        auth: Provision,
+        object_handle: ObjectHandle,
+        persistent: Persistent,
+    ) -> Result<ObjectHandle> {
+        // A session for authorization is required.
+        if self.sessions.0.is_none() {
+            error!("Missing session handle for authorization of auth.");
+            return Err(Error::local_error(ErrorKind::InvalidParam));
```
ionut-arm, line 892: I'm wondering if this deserves its own error kind, invalid param seems a bit far fetched and I'm sure there are other API methods where you need at least one session.

tests/context_tests.rs
```diff
@@ -1853,6 +1859,60 @@ mod test_ctx_load {
...
+            .expect("Failed to close persistant handle");
+
+        let retireved_persistant_handle = context
+            .tr_from_tpm_public(TpmHandle::Persistent(persistent_tpm_handle))
+            .expect("Failed to load the persistant handle");
+
+        assert_ne!(
+            retireved_persistant_handle,
+            ObjectHandle::from(ESYS_TR_NONE)
+        );
```
ionut-arm, line 1912: Could you also add a step to evict the persistent object and then try to load the TPM resource again?

ionut-arm (commented):
Thanks! Left couple of comments, I swear I thought we had this implemented already 😮

src/context.rs
puiterwijk: There are many of those, yeah. I think it might be an idea to make this a standard non-public `fn Context.require_a_session(&self) -> Result<()>` function.

Superhepper: Yeah, I will look into a new error type and see how to make it easy to use in context.

src/context.rs
```diff
@@ -874,6 +878,41 @@ impl Context {
...
+    /// Evicts persistent objects or allows certain transient objects
+    /// to be made peristent.
+    pub fn evict_control(
+        &mut self,
+        auth: Provision,
+        object_handle: ObjectHandle,
+        persistent: Persistent,
+    ) -> Result<ObjectHandle> {
+        // A session for authorization is required.
+        if self.sessions.0.is_none() {
```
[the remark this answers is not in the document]
Superhepper: Oh sorry forgot about that! I was thinking that I have to check the specification to see if there might be any cases where more then one session is required. I will look into it!

Superhepper:
There that looks good.

puiterwijk (approved):
LGTM

src/context.rs
```diff
@@ -1729,6 +1762,16 @@ impl Context {
...
+
+    /// Function that returns the required
+    /// session handle if it is available else
+    /// returns an error.
+    fn required_session_1(&self) -> Result<ESYS_TR> {
+        self.sessions.0.map(|v| v.handle().into()).ok_or_else(|| {
+            error!("Missing session handle for authorization (authSession1 = None)");
+            Error::local_error(ErrorKind::MissingAuthSession)
+        })
+    }
```
ionut-arm, line 1774: I wonder if we need to document the fact that we rely on the first session in that tuple whenever a session is _required_ - what do you think?

ionut-arm (approved):
👌

src/context.rs
Superhepper: Yes, I think that is needed. That is something we should do for set_sessions too. And refer to the ESAPI specification.

PR #147, opened 2020-11-11
Move Context functions to existing abstracted types

puiterwijk:
After this, the following types remain in `Context` in a `pub fn`:
- `TPM2B_PUBLIC` - We agreed this could be done later as it's very large and complex
- `TPM2B_SENSITIVE`
- `TPMT_SIG_SCHEME`
- `TPM2B_ATTEST`
- `TPMA_LOCALITY`
- `TPM2_CC`
I will try to send PRs for the ones that are not `TPM2B_PUBLIC` soon, but I wanted to get started on the ones where we have types already available.


src/structures/buffers.rs
```diff
@@ -92,6 +92,14 @@ pub mod data {
     buffer_type!(Data, 64, TPM2B_DATA);
 }
 
+pub mod sensitive_data {
+    buffer_type!(SensitiveData, 256, TPM2B_SENSITIVE_DATA);
```
Superhepper, line 96: Nice!

The reason that SensitiveData type was alone in ```src/algorithm/structures/sensitive/data.rs``` was because when I did a rather big refactoring a while back. I used the chapters in TPM Specification (Structures) as template for how to structure things, so because the sensitive data structure is specified in *Algorithm Parameters and Structures* chapter it ended up with the path it did (now that I look at it should actually have been 
```src/algorithm/structures/symmetric/sensitive/data.rs```).

But maybe that was not best way to do things ... perhaps it would have made more sense to do
 --> TPM2B - folder
 --> TPMI -folder
 --> TPML -folder

and so on ...


If the latter makes more sense then I think we should make an effort to structure things that way instead

**Review by Superhepper — COMMENTED** [review]
Looks good I only had one comment.

File: src/structures/buffers.rs (hunk 1 above)
Comment (puiterwijk) [reply src/structures/buffers.rs:96]: Right, though I was under the assumption that `buffers.rs` referred to `TPM2B_`-types, and `src/structures/lists/*` would be for `TPML_` structures.
So basically based it on the type prefix.
Honestly, if you prefer it in `src/algorithm/structures/symmetric/sensitive/data.rs`, that's fine, and I don't have an opinion about that, other than that I'd prefer `src/structures/algorithm/...` in that case, since it's still in the `Structures` part of the spec.
But I do thing that having it in `buffers.rs` makes sense as it's a `TPM2B_` type.

Comment (Superhepper) [reply src/structures/buffers.rs:96]: I have no strong opinion regarding this.

Right now I am leaning towards that it maybe would be better idea to structure things around the prefix instead.

Comment (puiterwijk) [reply src/structures/buffers.rs:96]: So, what would you prefer? Since I'd say that `buffers.rs` is the place for `TPM2B_` objects, or am I misunderstanding what you mean?

Comment (Superhepper) [reply src/structures/buffers.rs:96]: No, I think you are right. I think it is more logical to structure things around the prefix. So placing TPM2B in buffer.rs is a good idea. I just mean that we need to be consistent with this approach. Maybe do another refactoring soon in order to ensure this structure.

**Review by Superhepper — APPROVED** [review]

**Comment by ionut-arm** [comment]
Apologies for the delay, I'll review this over the weekend at some point. Thanks for moving so quickly!

File: src/context.rs [hunk 2]
```diff
@@ -652,7 +636,7 @@ impl Context {
         &mut self,
         key_handle: KeyHandle,
         message: PublicKeyRSA,
-        in_scheme: &TPMT_RSA_DECRYPT,
+        in_scheme: AsymSchemeUnion,
```
ionut-arm, line 639: We should rename this and get rid of the `Union` bit at some point

src/structures/buffers.rs
ionut-arm: I'm pro prefix-based placement as well - I noticed the weird placement of `SensitiveData` before but figured it was just a weird location in the spec

ionut-arm (approved):
Looks good, cheers!! 🎉

PR #150, opened 2020-11-17
Implement abstraction::nv::list

puiterwijk:

### DIFF
--- a/src/abstraction/nv.rs
+++ b/src/abstraction/nv.rs
@@ -2,9 +2,11 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::{
-    constants::tags::PropertyTag,
-    handles::{AuthHandle, NvIndexHandle, NvIndexTpmHandle, TpmHandle},
-    Context, Result,
+    constants::{tags::PropertyTag, tss::*, types::capability::CapabilityType},
+    handles::{AuthHandle, NvIndexHandle, NvIndexTpmHandle, ObjectHandle, TpmHandle},
+    nv::storage::NvPublic,
+    structures::{CapabilityData, Name},
+    Context, Error, Result, WrapperErrorKind,
 };
 
 /// Allows reading an NV Index completely, regardless of the max TPM NV buffer size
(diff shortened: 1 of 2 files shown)


src/abstraction/nv.rs
```diff
@@ -37,3 +39,25 @@ pub fn read_full(
 
     Ok(result)
 }
+
+pub fn list(context: &mut Context) -> Result<Vec<(NvPublic, Name)>> {
+    context.execute_without_session(|context| {
+        let (handles, _) = context.get_capabilities(
+            CapabilityType::Handles,
+            u32::to_be(TPM2_HT_NV_INDEX.into()),
```
Superhepper, line 47: Just a question, didn't you already create a ```PropertyTag``` type? So why is not used in ```get_capabilities```?

Superhepper (approved):
Looks good. Just had one question.

src/abstraction/nv.rs
puiterwijk: Because `PropertyTag` is tags for `CapabilityType::TPMProperties`, i.e. for TPM Properties (with thoughts of also including `CapabilityType::PCRProperties`). This is `CapabilityTypes::Handles`.

Superhepper: Ahh I see so that ```property``` variable is the 'Group' i.e. the type of handle (transient, persistent, permanent, nv_index ...).

puiterwijk: Almost. The first argument is `capability`, which is the group. Then the second one is `property`, which is which property to get (or start at), and then there's `propertyCount` to indicate how many.
e.g. `(TPM_Properties, VENDOR_STRING_1, 3)` would return `VENDOR_STRING_1`, `VENDOR_STRING_2, and `VENDOR_STRING_3`, but not `VENDOR_STRING_4`.

ionut-arm (approved):
Looks excellent - great having stuff like all those mappings abstracted. Could you also add a rustdoc description to this function?

src/abstraction/nv.rs
Superhepper: Yeah, that was what I meant. One thing though should it not be ```TPM2_NV_INDEX_FIRST``` for correctness? Not that it matters because in this case they have the same value.

But the specification says:
>The capability parameter determines the category of data returned. The property parameter selects the
>first value of the selected category to be returned. If there is no property that corresponds to the value of
>property, the next higher value is returned, if it exists.

puiterwijk: Yeah, fair enough. Updated that.

PR #151, opened 2020-11-17
Return all creation data

puiterwijk:
This adds rust types for CreationData and CreationTicket, and makes
sure that create_key and create_primary_key return all the information
returned by the TPM at creation time.


src/context.rs
```diff
@@ -402,15 +408,24 @@ impl Context {
...
             let primary_key_handle = KeyHandle::from(esys_prim_key_handle);
             let _ = self.open_handles.insert(primary_key_handle.into());
-            Ok(primary_key_handle)
+            Ok((
+                primary_key_handle,
+                *out_public,
+                creation_data,
+                creation_hash,
+                creation_ticket,
+            ))
```
Superhepper, line 428: Really like that all the data is returned it is something I have been missing!

Though I am not to found of how it is returned. I would much more have liked it if there were a struct that was returned.

src/context.rs
```diff
@@ -769,7 +792,7 @@ impl Context {
     }
 
     /// Read the public part of a key currently in the TPM and return it.
-    pub fn read_public(&mut self, key_handle: KeyHandle) -> Result<TPM2B_PUBLIC> {
+    pub fn read_public(&mut self, key_handle: KeyHandle) -> Result<(TPM2B_PUBLIC, Name, Name)> {
```
Superhepper, line 795: Same thing here. I would prefer a type instead of a tuple as return type.

Superhepper (commented):
Thanks looks pretty good, just had a few comments.

ionut-arm (approved):
Looks good to me, I'm ambivalent on whether `read_public` should return a struct - not sure what the "threshold" value should be, tbh.

Superhepper (approved):

Superhepper:
> 
> 
> Looks good to me, I'm ambivalent on whether `read_public` should return a struct - not sure what the "threshold" value should be, tbh.

Personally I am not a fan of public API:s returning tuples. But I as long as it is only two values in the tuple I can accept it.

PR #155, opened 2020-11-30
Improved handling of handles on drop.

### LINKED ISSUE #152: Persistent objects should not be flushed from context
I noticed that when dropping the `Context` after calling `tr_from_tpm_public()` with a `PersistentTpmHandle` instance, I would get these kinds of errors and warnings:
```
[2020-11-20T21:06:15Z INFO  tss_esapi::context] Closing context.
[2020-11-20T21:06:15Z INFO  tss_esapi::context] Flushing handle 4293479
WARNING:esys:src/tss2-esys/api/Esys_FlushContext.c:234:Esys_FlushContext_Finish() Received TPM Error 
ERROR:esys:src/tss2-esys/api/Esys_FlushContext.c:89:Esys_FlushContext() Esys Finish ErrorCode (0x000001c4) 
[2020-11-20T21:06:15Z ERROR tss_esapi::context] Error in flushing context: value is out of range or is not correct for the context (associated with parameter number 1).
[2020-11-20T21:06:15Z ERROR tss_esapi::context] Error when dropping the context: value is out of range or is not correct for the context (associated with parameter number 1).
[2020-11-20T21:06:15Z INFO  tss_esapi::context] Context closed.
```

It seems that `tr_from_tpm_public()` adds the object handle it returns to the `open_handles` set, to automatically flush it from TPM memory upon dropping the `Context` instance. The TPM 2.0 spec says this about the `FlushContext` command:
> This command causes all context associated with a loaded object, sequence object, or session to be
removed from TPM memory.
>
> This command may not be used to remove a persistent object from the TPM.

So perhaps `tr_from_tpm_public()` should not add the object handle to the `open_handles` set if the input handle type is `Persistent`? I'm not sure if filtering out just the `Persistent` handles would cover all object types that shouldn't be flushed. Probably not, considering all the variants of the `TpmHandle` enum (shown below), but I don't know exactly which variants map to the "persistent objects" referenced in the spec.
```
pub enum TpmHandle {
    Pcr(PcrTpmHandle),
    NvIndex(NvIndexTpmHandle),
    HmacSession(HmacSessionTpmHandle),
    LoadedSession(LoadedSessionTpmHandle),
    PolicySession(PolicySessionTpmHandle),
    SavedSession(SavedSessionTpmHandle),
    Permanent(PermanentTpmHandle),
    Transient(TransientTpmHandle),
    Persistent(PersistentTpmHandle),
    AttachedComponent(AttachedComponentTpmHandle),
}
```

Superhepper:
- Added a manager that holds the states of each handle
  and the actions that needs to be performed on the handle
  when the context is dropped.

- Made Context methods that opens, flushes or closes handles
  set the correct state for the handle they operate on in the 
  HandleManager. This should fix #152.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -1,5 +1,6 @@
 // Copyright 2020 Contributors to the Parsec project.
 // SPDX-License-Identifier: Apache-2.0
+mod handle_manager;
 use crate::{
     constants::{
         algorithm::{Cipher, HashingAlgorithm},
@@ -8,8 +9,8 @@ use crate::{
         types::{capability::CapabilityType, session::SessionType},
     },
     handles::{
-        AuthHandle, KeyHandle, NvIndexHandle, ObjectHandle, PcrHandle, PersistentTpmHandle,
-        SessionHandle, TpmHandle,
+        handle_conversion::TryIntoNotNone, AuthHandle, KeyHandle, NvIndexHandle, ObjectHandle,
+        PcrHandle, PersistentTpmHandle, SessionHandle, TpmHandle,
     },
     interface_types::{
         dynamic_handles::Persistent,
@@ -31,9 +32,10 @@ use crate::{
     },
     Error, Result, WrapperErrorKind as ErrorKind,
 };
+use handle_manager::{HandleDropAction, HandleManager};
 use log::{error, info};
 use mbox::MBox;
-use std::collections::{HashMap, HashSet};
+use std::collections::HashMap;
 use std::convert::{TryFrom, TryInto};
 use std::ffi::CString;
 use std::ptr::{null, null_mut};
@@ -73,8 +75,9 @@ pub struct Context {
     /// TCTI context handle associated with the ESYS context.
     /// As with the ESYS context, an optional Mbox wrapper allows the context to be deallocated.
     tcti_context: Option<MBox<TSS2_TCTI_CONTEXT>>,
-    /// A set of currently open object handles that should be flushed before closing the context.
-    open_handles: HashSet<ObjectHandle>,
+    /// Handle manager that keep tracks of the state of the handles and how they are to be
+    /// disposed.
+    handle_manager: HandleManager,
     /// A cache of determined TPM limits
     cached_tpm_properties: HashMap<PropertyTag, u32>,
 }
@@ -116,7 +119,7 @@ impl Context {
                 esys_context,
                 sessions: (None, None, None),
                 tcti_context,
-                open_handles: HashSet::new(),
+                handle_manager: HandleManager::new(),
                 cached_tpm_properties: HashMap::new(),
             };
             Ok(context)
@@ -176,9 +179,10 @@ impl Context {
 
         let ret = Error::from_tss_rc(ret);
         if ret.is_success() {
-            let _ = self
-                .open_handles
-                .insert(ObjectHandle::from(esys_session_handle));
+            self.handle_manager.add_handle(
+                ObjectHandle::from(esys_session_handle),
+                HandleDropAction::Flush,
+            )?;
             Ok(Session::create(
                 session_type,
                 SessionHandle::from(esys_session_handle),
@@ -413,7 +417,8 @@ impl Context {
             let creation_ticket = CreationTicket::try_from(*creation_ticket)?;
 
             let primary_key_handle = KeyHandle::from(esys_prim_key_handle);
-            let _ = self.open_handles.insert(primary_key_handle.into());
+            self.handle_manager
+                .add_handle(primary_key_handle.into(), HandleDropAction::Flush)?;
             Ok(CreatePrimaryKeyResult {
                 key_handle: primary_key_handle,
                 out_public: *out_public,
@@ -560,7 +565,8 @@ impl Context {
         let ret = Error::from_tss_rc(ret);
         if ret.is_success() {
             let key_handle = KeyHandle::from(esys_key_handle);
-            let _ = self.open_handles.insert(key_handle.into());
+            self.handle_manager
+                .add_handle(key_handle.into(), HandleDropAction::Flush)?;
             Ok(key_handle)
         } else {
             error!("Error in loading: {}.", ret);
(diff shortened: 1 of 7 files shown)


src/context.rs
```diff
@@ -913,7 +935,18 @@ impl Context {
...
+            let new_object_handle = ObjectHandle::from(esys_object_handle);
+            // If you look at the specification and see that it says ESYS_TR_NULL
+            // then that is an error in the spec. ESYS_TR_NULL was renamed to
+            // ESYS_TR NONE.
+            if new_object_handle != ObjectHandle::None {
+                self.handle_manager
+                    .add_handle(new_object_handle, HandleDropAction::Close)?;
+            } else {
+                self.handle_manager.set_as_closed(object_handle)?;
+            }
```
ionut-arm, line 947: I'm guessing that before you can call `evict_control` you need to get an `ESYS_TR` handle, so you can't just call it without `tr_from_tpm_public`?

src/context.rs
```diff
@@ -1789,19 +1836,43 @@ impl Context {
...
+            info!("Closing handle {}", ESYS_TR::from(*handle));
+            if let Err(e) = self.tr_close(handle) {
+                error!("Error when dropping context: {}.", e);
+            }
+        }
+
+        // Check if all handles have been cleaned up proeprly.
+        if self.handle_manager.has_open_handles() {
+            error!("Not all handles have had its rersources successfully released");
+        }
```
ionut-arm, line 1875: I guess it's a sanity check, but is this necessary? We have a mutable reference to `self`, so nothing could've been added between those calls above which together cover everything in the `HashMap`

ionut-arm (approved):
Thanks a lot, it's really slick with the handle manager!! Left a couple of nitpicks below, but overall I'm ok with the change

src/context.rs
```diff
@@ -1789,19 +1836,43 @@ impl Context {
...
+    /// Function that returns the handle as an ESYS_TR if it
+    /// was valid or an error if it was None.
+    fn ensure_not_none(handle: &ObjectHandle, fn_name: &str) -> Result<ESYS_TR> {
+        if *handle != ObjectHandle::None {
+            Ok((*handle).into())
+        } else {
+            error!("ObjectHandle::None cannot be used in call to {}", fn_name);
+            Err(Error::local_error(ErrorKind::InvalidParam))
+        }
+    }
```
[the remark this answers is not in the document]
Superhepper: No it got placed there because I was doing stuff in Context and needed the function. But you are right it is not really tied to the Context struct so it should probably not be there.

src/context.rs
Superhepper: Nah you could have a handle while running that you have evicted to make persistent and then you can evict the same handle again to make it transient again and when you do then the handles referring to that object becomes invalid. 

But you need to have an open and valid handle in order to call evict control when you want to make something transient.

src/context.rs
Superhepper: It is a "This should never happen" kind of check.

src/context.rs
```diff
@@ -1789,19 +1836,43 @@ impl Context {
...
+        for handle in self.handle_manager.handles_to_close().iter_mut() {
+            info!("Closing handle {}", ESYS_TR::from(*handle));
+            if let Err(e) = self.tr_close(handle) {
+                error!("Error when dropping context: {}.", e);
+            }
+        }
+
+        // Check if all handles have been cleaned up proeprly.
+        if self.handle_manager.has_open_handles() {
+            error!("Not all handles have had its rersources successfully released");
```
puiterwijk, line 1874:
```suggestion
            error!("Not all handles have had their resources successfully released");
```

src/context/handle_manager.rs
```diff
@@ -0,0 +1,133 @@
...
+                        "Closing handle({}) that should have been flushed",
+                        ESYS_TR::from(handle)
+                    );
+                    Err(Error::local_error(WrapperErrorKind::InvalidHandleState))
+                }
+            })
+    }
+
+    /// Retrieves all handles that needs to be flushed.
+    pub(super) fn handles_to_flush(&self) -> Vec<ObjectHandle> {
```
puiterwijk, line 100: I'm wondering whether we really need to build a new vector. Given that the only use for this is to iterate over it, maybe instead just return an `Iter<ObjectHandle>`?

src/handles/tpm.rs
```diff
@@ -37,6 +37,19 @@ pub enum TpmHandle {
     AttachedComponent(attached_component::AttachedComponentTpmHandle),
 }
 
+impl TpmHandle {
+    /// Method that indicates if the flushing the
+    /// context of the handle is a valid action.
+    pub fn may_be_flushed(&self) -> bool {
```
puiterwijk, line 43: Maybe we want this to be `pub(crate)` instead of fully `pub`? I don't think we want this as the public API per se?

src/context.rs
```diff
@@ -913,7 +935,18 @@ impl Context {
         };
         let ret = Error::from_tss_rc(ret);
         if ret.is_success() {
-            Ok(ObjectHandle::from(esys_object_handle))
+            let new_object_handle = ObjectHandle::from(esys_object_handle);
+            // If you look at the specification and see that it says ESYS_TR_NULL
+            // then that is an error in the spec. ESYS_TR_NULL was renamed to
+            // ESYS_TR NONE.
+            if new_object_handle != ObjectHandle::None {
```
puiterwijk, line 942: Maybe instead of this and the other check that also checks against `Null`, just add an
```
impl ObjectHandle {
    pub(crate) is_none() -> bool ...
```
?

puiterwijk (approved):
Some nitpicks, but the general idea looks great to me.

src/context/handle_manager.rs
Superhepper: It was just the easiest way to not hold a reference to the manager while calling close/flush because those functions also interacts with the manager. I will try to improve on it.

puiterwijk: Eh, I don't care that much (it's only upon drop, and not that invasive), just figured it might be nicer.

PR #157, opened 2020-12-07
Improve logging

ionut-arm:
This commit adds or improves the logging in the crate.

### DIFF
--- a/src/abstraction/transient.rs
+++ b/src/abstraction/transient.rs
@@ -659,7 +659,7 @@ impl TransientKeyContextBuilder {
             .find(|sz| **sz == self.root_key_size)
             .is_none()
         {
-            error!("The reference implementation only supports key sizes of 1,024 and 2,048 bits.");
+            error!("The reference implementation only supports key sizes of 1,024 and 2,048 bits");
             return Err(Error::local_error(ErrorKind::WrongParamSize));
         }
         let mut context = Context::new(self.tcti)?;
--- a/src/context.rs
+++ b/src/context.rs
@@ -101,7 +101,7 @@ impl Context {
         let ret = Tss2_TctiLdr_Initialize(tcti_name_conf.as_ptr(), &mut tcti_context);
         let ret = Error::from_tss_rc(ret);
         if !ret.is_success() {
-            error!("Error when creating a TCTI context: {}.", ret);
+            error!("Error when creating a TCTI context: {}", ret);
             return Err(ret);
         }
         let mut tcti_context = Some(MBox::from_raw(tcti_context));
@@ -124,7 +124,7 @@ impl Context {
             };
             Ok(context)
         } else {
-            error!("Error when creating a new context: {}.", ret);
+            error!("Error when creating a new context: {}", ret);
             Err(ret)
         }
     }
@@ -189,7 +189,7 @@ impl Context {
                 auth_hash,
             ))
         } else {
-            error!("Error when creating a session: {}.", ret);
+            error!("Error when creating a session: {}", ret);
             Err(ret)
         }
     }
@@ -315,6 +315,7 @@ impl Context {
             let capabilities = CapabilityData::try_from(*capabilitydata)?;
             Ok((capabilities, moredata))
         } else {
+            error!("Error when getting capabilities: {}", ret);
             Err(ret)
         }
     }
@@ -427,7 +428,7 @@ impl Context {
                 creation_ticket,
             })
         } else {
-            error!("Error in creating primary key: {}.", ret);
+            error!("Error in creating primary key: {}", ret);
             Err(ret)
         }
     }
@@ -512,7 +513,7 @@ impl Context {
                 creation_ticket,
             })
         } else {
-            error!("Error in creating derived key: {}.", ret);
+            error!("Error in creating derived key: {}", ret);
             Err(ret)
         }
     }
@@ -537,7 +538,7 @@ impl Context {
             let out_data = unsafe { MBox::from_raw(out_data) };
             Ok(SensitiveData::try_from(*out_data)?)
         } else {
-            error!("Error in unsealing: {}.", ret);
+            error!("Error in unsealing: {}", ret);
             Err(ret)
         }
     }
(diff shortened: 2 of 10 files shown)


ionut-arm:
Feel free to suggest a different wording on the log messages, particularly for the context policy methods.

Superhepper (approved):

puiterwijk:
So, I'm personally absolutely not a fan of libraries printing errors on their own, and we already have the ESAPI do that by itself.
This means that any errors that occur are printed, even if my using application is just trying to see if it works, and then has a fallback if it fails.

I think that if we want to add more information, that should get encoded in the `Error` type (like the [failure::ResultExt::with_context](https://docs.rs/failure/0.1.8/failure/trait.ResultExt.html#tymethod.with_context) approach).

Superhepper:
> So, I'm personally absolutely not a fan of libraries printing errors on their own, and we already have the ESAPI do that by itself.
> This means that any errors that occur are printed, even if my using application is just trying to see if it works, and then has a fallback if it fails.
> 
> I think that if we want to add more information, that should get encoded in the `Error` type (like the [failure::ResultExt::with_context](https://docs.rs/failure/0.1.8/failure/trait.ResultExt.html#tymethod.with_context) approach).
I actually agree with @puiterwijk here it is not so nice to throw a lot of errors when they perhaps are handled by the application that uses the library.
So my suggestion is to have look at using anyhow or thiserror crates in order to just provide some kind of context for the error so that it can be logged later by the application.

hug-dev:
> So, I'm personally absolutely not a fan of libraries printing errors on their own, and we already have the ESAPI do that by itself.

With the `log` macros (like `error`), the binary application that uses the crate can customise the log level and filter logs I believe? I agree it is bad if it's non-configurable but in the case of `error`, it seems it's only a bonus to add it?

I also agree that not having to use logging and have each `Error` instance actually contain why it failed would always be the best, would be nice to see what the options are

ionut-arm:
> So my suggestion is to have look at using anyhow or thiserror crates in order to just provide some kind of context for the error so that it can be logged later by the application.

I don't think there's a need to use those crates unless you want to surface a bunch of different errors coming from sources with their own types (in our case, it's just the errors from the crate and the ones from the TPM) - and they're an abstraction over `std::error::Error` which we implement, anyway. So it would be quite easy to use e.g. `anyhow` above this crate; from their README:
> Within the function, use ? to easily propagate any error that implements the std::error::Error trait.

For where we should log - I'd like to believe what Hugues was saying about disabling logs for a dependency, that was my understanding as well, but couldn't verify it easily. And my point for that was since you can control that sort of thing, you can rely on the crate to do the logging for you (instead of always logging for each call you make).

puiterwijk:
> For where we should log - I'd like to believe what Hugues was saying about disabling logs for a dependency, that was my understanding as well, but couldn't verify it easily. And my point for that was since you can control that sort of thing, you can rely on the crate to do the logging for you (instead of always logging for each call you make).

So, this depends on which actual `log` implementation you use.
The `log` crate is just a facade, and then you can use e.g. [env_logger](https://docs.rs/env_logger/0.8.2/env_logger/) or any other crate to configure it.

I guess it could be possible to say the app has to configure the logging framework, but I really think that this should just get all encoded in the `Error` type, and up to the application to display.

ionut-arm:
> 
> So, this depends on which actual `log` implementation you use.
> The `log` crate is just a facade, and then you can use e.g. [env_logger](https://docs.rs/env_logger/0.8.2/env_logger/) or any other crate to configure it.
> 
> I guess it could be possible to say the app has to configure the logging framework, but I really think that this should just get all encoded in the `Error` type, and up to the application to display.

What if I add a new feature `logging` and hide log messages behind it? That sounds like a good compromise - if you want logging from the crate, you enable the feature.

puiterwijk:
> What if I add a new feature `logging` and hide log messages behind it? That sounds like a good compromise - if you want logging from the crate, you enable the feature.

That sounds like a reasonable compromise, yeah.

hug-dev:
What is the essential problem with the `error`? By themselves they won't print anything and as @puiterwijk said you need a logger to really print.
We don't dictate any logger from this crate, so it the problem that with some loggers that people use it is impossible to filter the logs from this crate? I would have thought that the use of `log` is quite idiomatic and widespread over the Rust ecosystem, without the need of a specific feature to enable it or not in crates.

puiterwijk:
> What is the essential problem with the `error`? By themselves they won't print anything and as @puiterwijk said you need a logger to really print.
> We don't dictate any logger from this crate, so it the problem that with some loggers that people use it is impossible to filter the logs from this crate? I would have thought that the use of `log` is quite idiomatic and widespread over the Rust ecosystem, without the need of a specific feature to enable it or not in crates.

My problem initially was that I kept seeing error messages come by, but after just testing, I realize that that's purely from the tss-esapi, and that indeed `log::error` doesn't print until you've set an error handler.
In which case, I'm much more fine with it, and I will go look at the actual contents of the PR now for a full review.

puiterwijk (approved):
This looks pretty decent.
The one thing I'd like to see changed is the removal of the full stops after the various messages, but that's my preference.
If you like it like this, feel free to merge.

ionut-arm:
> but after just testing, I realize that that's purely from the tss-esapi, and that indeed `log::error` doesn't print until you've set an error handler.

I'm curious now if we can somehow disable the TSS2 logs, let me check

ionut-arm:
Seems there is: https://github.com/tpm2-software/tpm2-tss/blob/master/doc/logging.md#runtime-log-level

I'll add a mention on the Readme and in the docs about this.

> the removal of the full stops after the various messages

To be honest, I didn't even notice the full stop. Will remove

hug-dev:
> Seems there is: https://github.com/tpm2-software/tpm2-tss/blob/master/doc/logging.md#runtime-log-level

We could maybe set that ourselves in this crate? Either disable them completely or enable it through a method?

ionut-arm:
> 
> We could maybe set that ourselves in this crate? Either disable them completely or enable it through a method?

I don't think we should be making decisions for the end user, we're just a dependency. We can tell them how to control that, but not control it for them

puiterwijk:
Yeah, I agree with @ionut-arm here: rust-tss-esapi is just a wrapper, and shouldn't make decisions.
I wished esapi had a different default, but that's their decision.

puiterwijk (approved):
:+1: thanks!

hug-dev (approved):
🥇

PR #159, opened 2020-12-09
Update bindgen

ionut-arm:
This uses the commit in #154 and removes the dependency on 1.38.0 and nightly.

If there is any other requirement that I forgot about, let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,15 @@ case:
 $ export PKG_CONFIG_PATH=/usr/local/lib/pkgconfig
 ```
 
+## Minimum Supported Rust Version (MSRV)
+
+At the moment we test (via CI) and support the following Rust compiler versions:
+
+* On Ubuntu we test with the latest stable compiler version, as accessible through `rustup`.
+* On Fedora we test with the compiler version included with the Fedora 33 release.
+
+If you need support for other versions of the compiler, get in touch with us to see what we can do!
+
 ## Community channel
 
 Come and talk to us in [our Slack channel](https://app.slack.com/client/T0JK1PCN6/CPMQ9D4H1)!
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ zeroize = { version = "1.1.0", features = ["zeroize_derive"] }
 env_logger = "0.7.1"
 
 [build-dependencies]
-bindgen = "0.54.0"
+bindgen = "0.56.0"
 pkg-config = "0.3.18"
 
 [features]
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -33,6 +33,15 @@ case:
 $ export PKG_CONFIG_PATH=/usr/local/lib/pkgconfig
 ```
 
+## Minimum Supported Rust Version (MSRV)
+
+At the moment we use test and support the following Rust compiler versions:
+
+* On Ubuntu we test with the latest stable compiler version, as accessible through `rustup`.
+* On Fedora we test with the compiler version included with the Fedora 33 release.
```
hug-dev, line 41: Do we test with the _latest_ compiler version available in Fedora 33?

README.md
```diff
@@ -33,6 +33,15 @@ case:
...
+
+If you need support for other versions of the compiler, get in touch with us to see what we can do!
```
hug-dev, line 43: Maybe mention somewhere that for those supproted versions we test on the CI that every PR works with them.

hug-dev (commented):
Nice!! We should keep that in mind for Parsec projects as well 👍

README.md
ionut-arm: I think we test with the one that comes with the OS release? CC @puiterwijk

README.md
ionut-arm: I made a mistake on the line before the list above, will include it there.

Superhepper: I was under the impression that we would use the packaged versions for the different distributions so in the case of ubuntu one would need to do the following: ```apt update && apt -y upgrade && apt install rustc && apt install cargo```. So that we could say that we support the latest version that the distribution provides using its update channels.

README.md
puiterwijk: Yes, we test with the one that's included in Fedora 33.

README.md
ionut-arm: In the Dockerfile we actually build with the pipe-to-shell command from the [Rust website](https://www.rust-lang.org/tools/install) - but that's a good point, maybe we should try out the one included in Ubuntu.

ionut-arm: It seems the latest supported is `1.43.0` (for `rustc`), but the weird thing is that `cargo` has version `0.44.1` 🤔

hug-dev: > I was under the impression that we would use the packaged versions for the different distributions

I think we decided that as well for all the distributions that we want to support (like Fedora) but I think we should also definitely test with the latest stable Rust got from `rustup`. If Ubuntu is also one of the distribution we want to support then we should do the three:
* Ubuntu using the latest version of `cargo` from `apt`
* Fedora using the latest version of `cargo` from `dnf`
* *any distrib* using the latest version from `rustup`

My point being that for the case when we test on `rustup`, the distribution should not really matter.

Superhepper (approved):

PR #162, opened 2020-12-10
Implement changeauth operations

puiterwijk:
This implements both ObjectChangeAuth and HierarchyChangeAuth.

### DIFF
--- a/tests/context_tests.rs
+++ b/tests/context_tests.rs
@@ -1257,6 +1257,69 @@ mod test_policy_pcr {
     }
 }
 
+mod test_change_auth {
+    use super::*;
+
+    #[test]
+    fn test_object_change_auth() {
+        let mut context = create_ctx_with_session();
+
+        let prim_key_handle = context
+            .create_primary_key(
+                Hierarchy::Owner,
+                &decryption_key_pub(),
+                None,
+                None,
+                None,
+                PcrSelectionListBuilder::new().build(),
+            )
+            .unwrap()
+            .key_handle;
+        let keyresult = context
+            .create_key(
+                prim_key_handle,
+                &decryption_key_pub(),
+                None,
+                None,
+                None,
+                PcrSelectionListBuilder::new().build(),
+            )
+            .unwrap();
+        let loaded_key = context
+            .load(prim_key_handle, keyresult.out_private, keyresult.out_public)
+            .unwrap();
+
+        let random_digest = context.get_random(16).unwrap();
+        let new_key_auth = Auth::try_from(random_digest.value().to_vec()).unwrap();
+
+        let new_private = context
+            .object_change_auth(loaded_key.into(), prim_key_handle.into(), new_key_auth)
+            .unwrap();
+        context
+            .load(prim_key_handle, new_private, keyresult.out_public)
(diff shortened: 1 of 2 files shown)


tests/context_tests.rs
```diff
@@ -1257,6 +1257,69 @@ mod test_policy_pcr {
...
+    fn test_hierarchy_change_auth() {
+        let mut context = create_ctx_with_session();
+
+        let random_digest = context.get_random(16).unwrap();
+        let new_auth = Auth::try_from(random_digest.value().to_vec()).unwrap();
+
+        // NOTE: If this test failed on your system, you are probably running it against a
+        //  real (hardware) TPM or one that is provisioned. This hierarchy is supposed to be
+        //  used by the platform. It's used in this test because if we fail at restoring auth,
+        //  it should not be a big deal on a software TPM, and it won't impact the other tests.
```
ionut-arm, line 1313: I do hope people don't decide to run all these tests against a real TPM 😬

tests/context_tests.rs
```diff
@@ -1257,6 +1257,69 @@ mod test_policy_pcr {
...
+        let mut context = create_ctx_with_session();
+
+        let prim_key_handle = context
+            .create_primary_key(
+                Hierarchy::Owner,
+                &decryption_key_pub(),
+                None,
+                None,
+                None,
+                PcrSelectionListBuilder::new().build(),
```
ionut-arm, line 1274: Seeing this provided in each test for each `create..`, I'm in two minds about potentially changing this to an `Option` too, so we can just pass `None`

ionut-arm (approved):
Looks good!

tests/context_tests.rs
puiterwijk: I know I do sometimes, just out of laziness! This note was here primarily for myself 😀

tests/context_tests.rs
puiterwijk: Yep, makes sense. I can create a PR for that.

tests/context_tests.rs
ionut-arm: Does it not affect the TPM? I'd be afraid of ending up with it ☠️

tests/context_tests.rs
ionut-arm: No need to do that now, I'll raise an issue

tests/context_tests.rs
puiterwijk: On my system it's been doing pretty decently. And I've found a couple of "out of resource" conditions there, that made me rework some code!
If my TPM breaks, I'll just call the vendor and get the motherboard of my system replaced  😀

puiterwijk: Honestly, the only things that will actually impact the TPM's lifespan are writes to the NV, and changing the hierarchy authentication. All the rest is not stored inside the TPM's persistent NVRAM.  (which is really the only part that will wear out)

tests/context_tests.rs
puiterwijk: :+1:

Superhepper (approved):
Looks good to me

Issue #163, opened 2020-12-10
Make PCR selection optional for key creation


ionut-arm:
Most params for `create_key` and `create_primary_key` are `Option<...>`, however `PcrSelectionList` isn't even though (at least for our use cases) it's not that common. Worth considering if that should be changed to an `Option` too.

puiterwijk:
This is fixed with #163.

PR #164, opened 2020-12-10
Make PCR Selection optional for create_{primary_,}key

puiterwijk:
Issue #163 is marked as a question, but I figured I'd draft a PR for it if we do want it.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -370,7 +370,7 @@ impl Context {
         auth_value: Option<&Auth>,
         initial_data: Option<&SensitiveData>,
         outside_info: Option<&Data>,
-        creation_pcrs: PcrSelectionList,
+        creation_pcrs: Option<PcrSelectionList>,
     ) -> Result<CreatePrimaryKeyResult> {
         let sensitive_create = TPM2B_SENSITIVE_CREATE {
             size: std::mem::size_of::<TPMS_SENSITIVE_CREATE>()
@@ -381,6 +381,7 @@ impl Context {
                 data: TPM2B_SENSITIVE_DATA::try_from(initial_data.cloned().unwrap_or_default())?,
             },
         };
+        let creation_pcrs = PcrSelectionList::list_from_option(creation_pcrs);
 
         let mut outpublic = null_mut();
         let mut creation_data = null_mut();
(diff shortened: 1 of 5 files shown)


src/context.rs
```diff
@@ -381,6 +381,11 @@ impl Context {
             },
         };
 
+        let creation_pcrs = match creation_pcrs {
+            Some(val) => val,
+            None => PcrSelectionListBuilder::new().build(),
+        };
```
ionut-arm, line 387:
```suggestion
        let creation_pcrs = creation_pcrs.unwrap_or(PcrSelectionListBuilder::new().build());
```

ionut-arm (approved):
🚀

src/context.rs
```diff
@@ -381,6 +381,11 @@ impl Context {
             },
         };
 
+        let creation_pcrs = match creation_pcrs {
+            Some(val) => val,
```
Superhepper, line 385: I had this same problem when I did Session where an Option<Session> that was none was supposed to be transformed into an ESYS_TR_NONE. I tried to make some attempts at using simple conversions but failed because I neither ESYS_TR nor Option<T> where my types. So I ended up declaring a static function belonging to the Session struct that makes these kind of conversions. 

I wonder if something similar should be made for PcrSelectionList?

ionut-arm: I think it depends on how many places there are that use this conversion - for `Session` it's used in lots of places, right? Don't know how common `PcrSelectionList` is

Superhepper: I think it would be nice because it would make the approach consistent to have None is converted into TSS type.

src/context.rs
puiterwijk: Thanks for reminding me of all the alternative functions that avoid match'es that I keep forgetting to use!

ionut-arm (approved):

src/context.rs
ionut-arm: I'm ok with either approach, I'll let you decide

puiterwijk: I added `PcrSelectionList::list_from_option(pcr_list: Option<PcrSelectionList>) -> PcrSelectionList`.

Superhepper (approved):
Perfect!

PR #165, opened 2020-12-11
Implement tpm2_clear{,_control}

puiterwijk:


Superhepper (approved):
Thanks was thinking about doing this my self but got stuck in documentation.

LGTM!

puiterwijk:
> Thanks was thinking about doing this my self but got stuck in documentation.

That's a pretty good place to get stuck at... I'll admit I've not been doing the best at adding documentation everywhere!

ionut-arm (approved):
👍🏻

PR #168, opened 2020-12-18
Added documentation to Context methods.

Superhepper:
- Added documentation to the Context methods so that
  they at least have some small documentation.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -144,15 +144,34 @@ impl Context {
     /// If the returned session handle from ESYS api is ESYS_TR_NONE then
     /// the value of the option in the result will be None.
     ///
-    /// The caller nonce is passed as a slice and converted by the method in a TSS digest
-    /// structure.
+    /// # Example
     ///
-    /// # Constraints
-    /// * nonce must be at most 64 elements long
+    /// ```rust, no_run
+    /// use tss_esapi::{Context, Tcti,
+    ///     constants::{
+    ///         algorithm::{Cipher, HashingAlgorithm},
+    ///         types::session::SessionType,
+    ///     },
+    /// };
+    /// // Create context that uses Device TCTI.
+    /// let mut context = unsafe {
+    ///     Context::new(Tcti::Device(Default::default())).expect("Failed to create Context")
+    /// };
     ///
-    /// # Errors
-    /// * if the `nonce` is larger than allowed, a `WrongSizeParam` wrapper error is returned
-    // TODO: Fix when compacting the arguments into a struct
+    /// // Create auth session without key_handle, bind_handle
+    /// // and Nonce
+    /// let session = context
+    ///     .start_auth_session(
+    ///         None,
+    ///         None,
+    ///         None,
+    ///         SessionType::Hmac,
+    ///         Cipher::aes_256_cfb(),
+    ///         HashingAlgorithm::Sha256,
+    ///     )
+    ///     .expect("Failed to create session")
+    ///     .expect("Recived invalid handle");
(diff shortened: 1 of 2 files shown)


ionut-arm (approved):
LGTM, are you intending to add references to the spec?

Superhepper:
I am not sure. I might. But most of all I am going to add some examples before this is done.

src/context.rs
```diff
@@ -346,7 +353,42 @@ impl Context {
...
+    /// // Create context test TCTI used for internal testing.
+    /// let tcti = env::var("TEST_TCTI")
+    ///     .map_or_else(
+    ///         |_| {
+    ///             Tcti::Mssim(Default::default())
+    ///         },
+    ///         |v| {
+    ///             Tcti::from_str(&v).expect("Error parsing TEST_TCTI")
+    ///     });
+    /// let mut context = unsafe { Context::new(tcti).expect("Failed to create Context") };
```
ionut-arm, line 385: I would probably leave this out of the example shown in the docs ([with `#`](https://doc.rust-lang.org/rustdoc/documentation-tests.html#hiding-portions-of-the-example)) - it makes it too verbose otherwise. I'm not sure for the other examples how much you should leave in the visible example - e.g. does the session creation have something relevant in that sense for it to be included?

src/context.rs
```diff
@@ -2092,9 +2270,11 @@ impl Context {
         }
     }
 
-    /// NV Write
-    ///
     /// Writes data to an nv index.
+    ///
+    /// # Details
+    /// This method is used to write a value from
+    /// the nv memory in the TPM.
```
ionut-arm, line 2277:
```suggestion
    /// This method is used to write a value to
    /// the nv memory in the TPM.
```

src/context.rs
Superhepper: The only reason it is there is because I had not found the flag to turn off testing the examples in the docmentation. So the CI would fail because I used the wrong TCTI. I really did not want it there. Now I have found the flag I was looking for which only makes sure that the examples are compiled but do not run them.

Superhepper: This have been fixed now.

Superhepper:
Do you think the links to the specifications are a good thing?

puiterwijk:
> Do you think the links to the specifications are a good thing?

I don't think that that makes a lot of sense. The `Context` layer is pretty hard to use for the non-initiated, and the functions are named close enough for the initiated to find them I think.
The `abstraction` layer should be such that you don't need the specifications, and then sending people there would be detrimental to adoption, I think, as it would mean that they get sent to highly complicated documents to use simple things.

puiterwijk:
I think that we could do with a higher level crates that are more of the `abstraction` type of code, but I think that that should be separate endeavours.
I've already done that at some places, like e.g. https://github.com/fedora-iot/rust-tpm2-policy for policy abstraction (though for now it's still lacking a lot of policies, and documentation).

ionut-arm (commented):
I'm ok with it either way, whether the references to the spec are there or not, I'm just glad we have code examples now

puiterwijk:
> I'm ok with it either way, whether the references to the spec are there or not, I'm just glad we have code examples now

Oh, absolutely! Thanks @Superhepper for the code examples and documentation!

Superhepper:
I think this PR is finished now. Even though the documentation of Context is not finished it contains a couple a good examples so it is easy for others to continue. And I do not want the PR to be to big.

puiterwijk (approved):
Looks good to me! Thanks!

PR #170, opened 2020-12-22
Improved session attributes

Superhepper:
- Renamed TpmaSession to SessionAttributes.
- Removed raw tss constants from SessionAttributes api.
- Added new type SessionAttributesMask.
- Added a new SessAttributesBuilder.


src/context.rs
```diff
@@ -278,11 +274,15 @@ impl Context {
...
+        // let (session_attributes, session_attributes_mask) = TpmaSessionBuilder::new()
+        //     .with_flag(TPMA_SESSION_DECRYPT)
+        //     .with_flag(TPMA_SESSION_ENCRYPT)
+        //     .build();
+        let (session_attributes, session_attributes_mask) = SessionAttributesBuilder::new()
+            .with_decrypt(true)
+            .with_encrypt(true)
             .build();
-        self.tr_sess_set_attributes(session, session_attr)?;
+        self.tr_sess_set_attributes(session, session_attributes, session_attributes_mask)?;
```
ionut-arm, line 285: Just realised something - do we want this to be done here? Or, should we at least document that we're adding these attributes to the session (or make them a parameter to the method)?

ionut-arm (commented):
Left one comment but overall it looks good, thanks for the changes! Also, I think there was a place where some code was left commented out, I think just above my comment

src/context.rs
ionut-arm: Hmm, I'm struggling to identify what changes you've made here 😕

Superhepper: I wrote a reply for this and pressed resolve and thought it would add the reply before resolving but that is not the case.

Any way. It is not a problem that the attributes are set to the session inside the method because it is a session that is local to that method and it gets flushed when the at the end of the method. If some one would like other attributes they would have to create a session them self and use ```execute_with_session``` instead. 

But you are right it should be added to the documentation of the method that the closure will get executed using a session with these attributes.

Superhepper: I have now fixed the documentation.

ionut-arm: I think just having the doc is fine, my worry was related to something we noticed when testing with some random version of TSS on a RaspberryPi, where an operation failed because `decrypt` was set to true for the session even though the operation did not allow decrypt/did not have something to decrypt in the response (I believe it was a call to `GetRandom`). Might be confusing with these attributes not documented

ionut-arm (approved):
👍🏻

puiterwijk (approved):
👍

PR #171, opened 2021-01-05
Added missing size check when converting native buffer into TSS buffer.

Superhepper:
For this size check there are two ways to deal with it and I chose to be pedantic and check the size even though it should never be problem.

The other way to deal with it is to change the conversion into  ```From``` conversion.

### DIFF
--- a/src/nv/storage/public.rs
+++ b/src/nv/storage/public.rs
@@ -75,7 +75,7 @@ impl TryFrom<NvPublic> for TPM2B_NV_PUBLIC {
                 nvIndex: nv_public.nv_index.into(),
                 nameAlg: nv_public.name_algorithm.into(),
                 attributes: nv_public.attributes.try_into()?,
-                authPolicy: nv_public.authorization_policy.try_into()?,
+                authPolicy: nv_public.authorization_policy.into(),
                 dataSize: nv_public.data_size as u16,
             },
         })
--- a/src/structures/buffers.rs
+++ b/src/structures/buffers.rs
@@ -71,16 +71,15 @@ macro_rules! named_field_buffer_type {
             }
         }
 
-        impl TryFrom<$native_type> for $tss_type {
-            type Error = Error;
-
-            fn try_from(native: $native_type) -> Result<Self> {
+        impl From<$native_type> for $tss_type {
+            fn from(native: $native_type) -> Self {
                 let mut buffer = $tss_type {
                     size: native.0.len() as u16,
                     ..Default::default()
                 };
+                buffer.size = native.0.len() as u16;
                 buffer.$buffer_field_name[..native.0.len()].copy_from_slice(&native.0);
-                Ok(buffer)
+                buffer
             }
         }
     };
--- a/src/context.rs
+++ b/src/context.rs
@@ -183,7 +183,7 @@ impl Context {
         auth_hash: HashingAlgorithm,
     ) -> Result<Option<Session>> {
         let nonce_ptr: *const TPM2B_NONCE = match nonce {
-            Some(val) => &TPM2B_NONCE::try_from(val.clone())?,
+            Some(val) => &val.clone().into(),
             None => null(),
         };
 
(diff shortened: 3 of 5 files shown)


src/structures/buffers.rs
```diff
@@ -76,6 +76,10 @@ macro_rules! named_field_buffer_type {
 
             fn try_from(native: $native_type) -> Result<Self> {
                 let mut buffer: $tss_type = Default::default();
+                if native.0.len() > $native_type::MAX_SIZE {
+                    error!("Error: Invalid buffer size(> {})", $native_type::MAX_SIZE);
+                    return Err(Error::local_error(WrapperErrorKind::WrongParamSize));
+                }
```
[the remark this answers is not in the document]
Superhepper: To me it was kind of a code smell to have a ```TryFrom``` that could never fail. So this sanity check made the code more consistent which was nice.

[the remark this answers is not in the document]
ionut-arm: Actually, that's true, do we *need* it to be `TryFrom`? I don't think it's necessarily a smell, but we could also consider replacing the returned error with [`Infallible`](https://doc.rust-lang.org/std/convert/enum.Infallible.html)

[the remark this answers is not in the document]
Superhepper: We do not need it to be a ```TryFrom``` because the size of the buffer is validated as the native type is created. Changing the error type to be Infallible would have removed the sense of code smell because that would have shown intent.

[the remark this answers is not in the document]
puiterwijk: So is the plan now to change the `Error` type to `Infallible` or keep it as is? The comments seem to imply we agree to use `Infallible` or just `From<T>`, but I don't see a change?
(I'd be fine with either, just want to make sure we're agreed)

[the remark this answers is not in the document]
Superhepper: Sorry I have not been able to work on it that is why you have not seen a change. I thought it looked it nice and consistent to have a size check for all conversions in the macro even though in this conversion it would only have been a sanity check. But I do not really care that much. When I find the time I will look if it can be a ```From<T>``` conversion otherwise I will change the error to ```Infallible```.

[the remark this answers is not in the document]
Superhepper: It has now been changed to a ```From``` conversion.

src/nv/storage/public.rs
```diff
@@ -10,7 +10,7 @@ use crate::{
     Error, Result, WrapperErrorKind,
 };
 use log::error;
-use std::convert::{TryFrom, TryInto};
+use std::convert::{From, TryFrom, TryInto};
```
ionut-arm, line 13: I don't think you need to import `From`, it gets included in the [prelude](https://doc.rust-lang.org/std/prelude/index.html)

src/structures/buffers.rs
```diff
@@ -71,16 +71,23 @@ macro_rules! named_field_buffer_type {
...
 
-        impl TryFrom<$native_type> for $tss_type {
-            type Error = Error;
+        // impl TryFrom<$native_type> for $tss_type {
+        //     type Error = Infallible;
+
+        //     fn try_from(native: $native_type) -> Result<Self> {
+        //         Ok($tss_type::from(native))
+        //     }
+        // }
```
ionut-arm, line 80: Keeping this in case we need to revert?

ionut-arm (commented):
Thank you for the update!

src/structures/buffers.rs
Superhepper: No forgot to remove it.

Superhepper: Fixed

src/nv/storage/public.rs
Superhepper: Fixed

puiterwijk (commented):
This looks fine to me. Once you push a new version that removes the `From` import, I'll approve.

Superhepper:
> 
> 
> This looks fine to me. Once you push a new version that removes the `From` import, I'll approve.

Done

puiterwijk (approved):
👍

src/structures/buffers.rs
```diff
@@ -71,16 +71,15 @@ macro_rules! named_field_buffer_type {
...
-            type Error = Error;
-
-            fn try_from(native: $native_type) -> Result<Self> {
+        impl From<$native_type> for $tss_type {
+            fn from(native: $native_type) -> Self {
                 let mut buffer = $tss_type {
                     size: native.0.len() as u16,
                     ..Default::default()
                 };
+                buffer.size = native.0.len() as u16;
```
ionut-arm, line 80: Isn't the `size` field assignment duplicated here?

Superhepper: Yup, that is not how it is supposed to be. I'll send up PR in a sec.

PR #172, opened 2021-01-05
Potential improvement to get_tpm_properties

Superhepper:
Potential speed improvement of the get_tpm_properties function. By using the extend function on the HashMap containing the cached values.


ionut-arm (commented):
I'm ok with the change and it does the same as before, I don't know if it's more or less readable, though 😕

Superhepper:
I have left this as draft because I am note sure my self that I like the change even though it could be a very slight performance improvement it could also as you say for some hurt the readability.

puiterwijk:
I'm okay with the change, although I prefer the original, but I'm heavily biased obviously :).

PR #174, opened 2021-01-11
Added proper documentation for pcr_* methods.

Superhepper:
Added proper documentation to the pcr_* context methods. This includes detailed description, argument information, examples and links to relevant types.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -1274,9 +1274,86 @@ impl Context {
 
     /// Extends a PCR with the specified digests.
     ///
+    /// # Arguments
+    /// * `pcr_handle`- A [PcrHandle] to the PCR slot that is to be extended.
+    /// * `digests` - The [DigestValues] with which the slot shall be extended.
+    ///
     /// # Details
     /// This method is used to cause an update to the indicated PCR. The digests param
     /// contains the digests for specific algorithms that are to be used.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # use tss_esapi::{
+    /// #     Context, Tcti,
+    /// #     constants::{
+    /// #         algorithm::Cipher,
+    /// #         types::session::SessionType,
+    /// #     },
+    /// #     session::SessionAttributesBuilder,
+    /// #     handles::PcrHandle,
+    /// #     structures::Digest,
+    /// # };
+    /// # use std::{env, str::FromStr};
+    /// # // Create TCTI that uses the correct software for testing
+    /// # let tcti = env::var("TEST_TCTI")
+    /// #   .map(|env_tcti_str| Tcti::from_str(&env_tcti_str).expect("Failed to parse TCTI string."))
+    /// #   .unwrap_or_else(|_| Tcti::Mssim(Default::default()));
+    /// # // Create context
+    /// # let mut context = unsafe {
+    /// #     Context::new(tcti).expect("Failed to create Context")
+    /// # };
+    /// # // Create session for a pcr
+    /// # let pcr_session = context
+    /// #     .start_auth_session(
+    /// #         None,
+    /// #         None,
+    /// #         None,
+    /// #         SessionType::Hmac,
+    /// #         Cipher::aes_256_cfb(),
+    /// #         tss_esapi::constants::algorithm::HashingAlgorithm::Sha256,
+    /// #     )
+    /// #     .expect("Failed to create session")
+    /// #     .expect("Recived invalid handle");
+    /// # let (session_attributes, session_attributes_mask) = SessionAttributesBuilder::new()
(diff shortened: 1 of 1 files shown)


src/context.rs
```diff
@@ -1274,9 +1274,75 @@ impl Context {
...
+    ///         HashingAlgorithm::Sha256,
+    ///     )
+    ///     .expect("Failed to create session")
+    ///     .expect("Recived invalid handle");
+    /// let (session_attributes, session_attributes_mask) = SessionAttributesBuilder::new()
+    ///     .with_decrypt(true)
+    ///     .with_encrypt(true)
+    ///     .build();
+    /// context.tr_sess_set_attributes(pcr_session, session_attributes, session_attributes_mask)
+    ///     .expect("Failed to set attributes on session");
```
ionut-arm, line 1322: I think these sections could be made invisible in the documentation, as they're not directly relevant to how to use the method

Superhepper: I am not sure how to do that. Because I like the idea that the documentation needs to be able to build otherwise we get an error.

ionut-arm: You just need to prefix the lines with `#`, as described [here](https://doc.rust-lang.org/rustdoc/documentation-tests.html#hiding-portions-of-the-example). Something like:
```rust
///#use tss_esapi::{
///#     Context, Tcti,
///#     constants::{
...
```

Superhepper: That is really neat. I wonder if we should actually make the code runnable with tests as well and then just show the good parts? Because in that case would could move all the "basic" context tests into the documentation and only show the good parts?

Superhepper: Have look at the ```pcr_extend``` method. If you think that looks ok. Then I will change the rest of the methods to look like that. I must say that it feels like I am lying when I do not show everything that needs to be done in order to get things to work hehe.

ionut-arm: > That is really neat. I wonder if we should actually make the code runnable with tests as well and then just show the good parts? Because in that case would could move all the "basic" context tests into the documentation and only show the good parts?

That might be ok, though I do remember facing some issues when trying to do something like that because the doc tests were run in a weird way. Also, I think moving all Context tests might be complete overkill - maybe just the ones that test each method works well?

src/context.rs
```diff
@@ -1274,9 +1274,83 @@ impl Context {
...
+    ///     constants::algorithm::HashingAlgorithm,
+    /// };
+    /// // Extend both sha256 and sha1
+    /// let mut vals = DigestValues::new();
+    /// vals.set(
+    ///     HashingAlgorithm::Sha1,
+    ///     Digest::try_from(vec![
+    ///         1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20,
+    ///     ])
+    ///     .expect("Failed to create Digest from data"),
```
ionut-arm, line 1337: I'm wondering if this could be hidden into a variable as well, and just used as `digest` - kind of like `pcr_session`

src/context.rs
Superhepper: Nah not remove everything from context test. Just like the basic. Like when test that calling the function doesn't result in an error.

src/context.rs
Superhepper: Fixed

src/context.rs
```diff
@@ -1298,12 +1375,63 @@ impl Context {
         }
     }
 
-    /// Resets the value in a PCR register.
+    /// Resets the value in a PCR.
+    ///
+    /// # Arguments
+    /// *`pcr_handle` -  A [PcrHandle] to the PCR slot that is to be resetted.
```
ionut-arm, line 1381: Is the `[PcrHandle]` part supposed to be a link? Or does it work like a link?

ionut-arm (approved):
I like the examples, though they stretch over so many lines X_X would be nice if there was some way of supporting a "common setup" for them, per file

src/context.rs
Superhepper: In the documentation it becomes a link to that type.

Superhepper:
> I like the examples, though they stretch over so many lines X_X would be nice if there was some way of supporting a "common setup" for them, per file

I tried to find a way to avoid. But the only way to that would be to expose test functions in the library and I did not want to do that. I had hoped that they some how could get hold of the common module in tests but it doesn't work like that. Each individual doc compiles to its own crate and can only make use of ```std``` and library it self like.

puiterwijk (approved):
Looks good to me.

src/context.rs
ionut-arm: Damn, that's useful, would've been good to know about it some while ago

PR #175, opened 2021-01-16
Added NvIndexAttributesBuilder.

Superhepper:
Added a builder for the NvIndexAttributes.

The advantage with this is that the builder can check
for missing or conflicting attributes and it does not
need to be mutable when creating the attributes.

### DIFF
--- a/src/nv/storage/index.rs
+++ b/src/nv/storage/index.rs
@@ -48,7 +48,7 @@ impl TryFrom<TPM2_NT> for NvIndexType {
             TPM2_NT_PIN_FAIL => Ok(NvIndexType::PinFail),
             TPM2_NT_PIN_PASS => Ok(NvIndexType::PinPass),
             _ => {
-                error!("Error: Found invalid value when trying to parse Nv Index Type");
+                error!("Found invalid value when trying to parse Nv Index Type");
                 Err(Error::local_error(WrapperErrorKind::InvalidParam))
             }
         }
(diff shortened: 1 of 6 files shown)


src/nv/storage/index.rs
```diff
@@ -61,31 +61,52 @@ bitfield! {
     pub struct NvIndexAttributes(TPMA_NV);
     impl Debug;
     // NV Index Attributes
-    pub pp_write, set_pp_write: 0;
-    pub owner_write, set_owner_write: 1;
-    pub auth_write, set_auth_write: 2;
-    pub policy_write, set_policy_write: 3;
+    pub pp_write, _: 0;
+    _, set_pp_write: 0;
```
ionut-arm, line 65: What was the problem with having the setter and the getter on the same line?

src/nv/storage/index.rs
```diff
@@ -105,3 +126,286 @@ impl From<NvIndexAttributes> for TPMA_NV {
...
+        self.nv_index_attributes.set_read_stclear(set);
+        self
+    }
+
+    /// Builds the nv index attributes.
+    ///
+    /// # Errors
+    /// Returns an error if some attributes are missing
+    /// or are in conflict with eachother.
+    pub fn build(self) -> Result<NvIndexAttributes> {
```
ionut-arm, line 368: Would it be useful to also provide this as a method on `NvIndexAttributes`, something like `fn check(&self) -> Result<()>`

tests/nv_storage_nv_index_attributes_tests.rs
```diff
@@ -60,78 +63,289 @@ mod test_nv_storage_nv_index_attributes {
...
+                .with_policy_write(true)
+                .build()
+        );
+    }
 
-        attributes.set_auth_write(true);
-        assert_eq!(0b0000_0000_0000_0000_0000_0000_0000_0111u32, attributes.0);
+    #[test]
+    fn test_nv_index_attributes_builder_missing_write_attribute_failure() {
+        // Test missing read error
```
ionut-arm, line 101: Complete nit: 
```suggestion
        // Test missing write error
```

tests/nv_storage_nv_index_attributes_tests.rs
```diff
@@ -60,78 +63,289 @@ mod test_nv_storage_nv_index_attributes {
...
+                .build()
+        );
+    }
 
-        // PinPass = 1001 (7,4)
-        attributes.set_index_type(NvIndexType::PinPass);
-        assert_eq!(0b0000_0000_0000_0000_0000_0000_1001_1111u32, attributes.0);
-        assert_eq!(NvIndexType::PinPass, attributes.index_type().unwrap());
+    #[test]
+    fn test_attributes_builder() {
```
ionut-arm, line 134: I'm debating the usefulness of this test - I guess it's testing the builder more than the code in `bitfield` (which would've been a bit weird). But I'm happy to keep it 😄

ionut-arm (commented):
Overall looks good, but I've left a few comments and questions below!

src/nv/storage/index.rs
Superhepper: For some weird reason I thought that the visibility specifier effected the line .. I will fix that.

src/nv/storage/index.rs
Superhepper: Maybe, I have not really thought about it. What I have been thinking about is to be able to initialize the builder using attributes. That would also give a way to verify that they are ok by then initializing a builder with the attributes you want to check and then call build on it. Though not so intuitive and user friendly perhaps.

```NvIndexAttributesBuilder::create_from_attributes(nv_index_attributes)```

File: src/nv/storage/index.rs (hunk 1 above)
Comment (ionut-arm) [reply src/nv/storage/index.rs:65]: Makes sense, thanks!

File: src/nv/storage/index.rs (hunk 2 above)
Comment (ionut-arm) [reply src/nv/storage/index.rs:368]: Hmm, I think it's ok as it is now, but maybe it's a good idea to change the `From<TPMA_NV>` above to a `TryFrom` that also does the checks? Don't know if by using `create_from_attributes` or in some other way, up to you!

File: tests/nv_storage_nv_index_attributes_tests.rs (hunk 4 above)
Comment (Superhepper) [reply tests/nv_storage_nv_index_attributes_tests.rs:134]: It sort of test the entire chain. From builder to attributes into raw value. So that the correct bit is set in the TSS value when you call corresponding function in builder with true as argument.

File: tests/nv_storage_nv_index_attributes_tests.rs (hunk 3 above)
Comment (Superhepper) [reply tests/nv_storage_nv_index_attributes_tests.rs:101]: Done

File: src/nv/storage/index.rs (hunk 2 above)
Comment (Superhepper) [reply src/nv/storage/index.rs:413]: I did all of it

File: src/nv/storage/index.rs [hunk 5]
```diff
@@ -61,47 +61,362 @@ bitfield! {
...
         NvIndexType::try_from(self.tss_index_type())
     }
+
+    /// Performs checks on the `NvIndexAttributes` and returns
+    /// `Ok(self)` if it passes all checks.
+    ///
+    /// # Errors
+    /// Returns an error if some attributes are missing
+    /// or are in conflict with eachother.
+    pub fn as_validated(&self) -> Result<NvIndexAttributes> {
```
ionut-arm, line 124: Is it possible to keep this as `validate(&self) -> Result<()>` and in the builder deconstruct it to return the `NvIndexAttributes` already found inside it?

ionut-arm (approved):
I'm ok with the changes and ambivalent about the question below

src/nv/storage/index.rs
Superhepper: fixed t

Superhepper: I think it is the way to go. Because as_validated would have been nice if ```&mut self``` had been required to perform the checks or if it was not possible to copy ```self``` trivially.

ionut-arm (approved):
:100: 

Thanks for the changes!

puiterwijk (approved):
Looks good to me

PR #176, opened 2021-01-26
Fixed specifying size twice.

Superhepper:


hug-dev (approved):
The PRs we want 🥇 

btw I don't know if that is the case in the `tss-esapi` crate, but if you implement `From<std::num::TryFromIntError>` on the error type then you can use `try_into` everywhere instead of casting, which is a bit more safe 😃

ionut-arm (approved):

PR #177, opened 2021-01-26
Split repo into two crates

ionut-arm:
This commit splits the repository into two crates: `tss-esapi` for the
Rust-native, idiomatic interface, and `tss-esapi-sys` for the FFI
bindings.

This is the first step in the process and probably multiple other PRs
will be needed to clean up.


tss-esapi-sys/build.rs
```diff
@@ -0,0 +1,118 @@
...
+    const TSS_INCLUDE_PATH: &str = "vendor/include/tss2";
+
+    use super::MINIMUM_VERSION;
+    use std::path::PathBuf;
+
+    // Only on a specific feature
+    pub fn generate_bindings() {
+        let out_path = std::path::PathBuf::from(std::env::var("OUT_DIR").unwrap());
+        let esys_path = out_path.join("tss_esapi_bindings.rs");
+        generate_from_system(esys_path.clone())
```
puiterwijk, line 47: I'd prefer it if `generate_from_submodule` was a separate feature, instead of the fallback if generating from the system fails.
Because on Fedora, we'd always want (need) to do the system generation, and if that fails, we'd want to know that that failed and why.

README.md
```diff
@@ -44,8 +18,8 @@ If you need support for other versions of the compiler, get in touch with us to
 
 ## Community channel
 
-Come and talk to us in [our Slack channel](https://app.slack.com/client/T0JK1PCN6/CPMQ9D4H1)!
-[Here](http://dockr.ly/slack) is how to join the workspace.
+Come and talk to us in [our Slack channel](https://cloud-native.slack.com/archives/C01EARH2ZB3)!
```
hug-dev, line 21: That's why nobody came 😞

tss-esapi-sys/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "tss-esapi-sys"
+version = "0.1.0"
+authors = ["Ionut Mihalcea <ionut.mihalcea@arm.com>",
+           "Hugues de Valon <hugues.devalon@arm.com>"]
```
hug-dev, line 5: Yes agree that it would be nice to put the same mention that in our copyright files, "Contributors to the Parsec project" or something similar. We should, by the way, do that everywhere!

tss-esapi-sys/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+edition = "2018"
+description = "FFI wrapper around TSS 2.0 Enhanced System API"
+readme = "README.md"
+keywords = ["tpm", "tss", "esys", "esapi"]
+categories = ["api-bindings", "external-ffi-bindings", "cryptography"]
+license = "Apache-2.0"
+repository = "https://github.com/parallaxsecond/rust-tss-esapi"
+documentation = "https://docs.rs/crate/tss-esapi-sys"
+
```
hug-dev, line 14: I think you need to add a `links` field, see [here](https://doc.rust-lang.org/cargo/reference/manifest.html#the-links-field)

tss-esapi-sys/README.md
```diff
@@ -0,0 +1,39 @@
...
+might happen that `pkg-config` can not find them. Run the following command if that is the
+case:
+```bash
+$ export PKG_CONFIG_PATH=/usr/local/lib/pkgconfig
+```
+
+The FFI bindings presented by this crate can be either those commited in the
+crate under `src/bindings` or generated on the fly from the library headers
+found on the system, at build time. For generating the bindings at build time
+please enable the `generate-bindings` feature, as it is not enabled by default.
```
hug-dev, line 33: Perfect 👌 !

tss-esapi-sys/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+NOTE: Only a limited set of bindings are committed and their target triplet
+is included in the name of the file - if the triplet you require is not
+available, feel free to raise a Pull Request to add it or to use build-time
+generation of bindings. All the committed bindings **MUST** be generated from
+the library version found under the `vendor` submodule.
```
hug-dev, line 39: Is it needed to generate one set of bindings per target for the TSS API? I believe it could be the same accross all targets it if uses platform-independant types and such

tss-esapi/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+hostname-validator = "1.0.0"
+regex = "1.3.9"
+zeroize = { version = "1.1.0", features = ["zeroize_derive"] }
+tss-esapi-sys = { path = "../tss-esapi-sys" }
+
+[dev-dependencies]
+env_logger = "0.7.1"
+
+[build-dependencies]
+bindgen = "0.56.0"
```
hug-dev, line 32: `bindgen` is optional on `generate-bindings`, right?

tss-esapi-sys/build.rs
hug-dev: I was wondering actually, should we have the `generate_from_submodule` fallback? I believe it is only used for the headers and not for compilation? How would someone then link to its TSS libraries when using the `generate_from_submodule` fallback?

If we give the constraint that `pkg-config` must be present on the system that that it must locate successfully the TSS libraries then we can simplify:
* by default, the crate will use its pre-generated bindings
* when `generate-bindings` is activated, we use `pkg-config` to find the headers and generate them

We would not need the `vendor` submodule then.

In the future, we could think about other modes:
1. The libraries are compiled by the crate itself to be linked locally and the bindings generated (or not)
2. The path to a custom location of the header files and libraries is given to generate the bindings (or not) and link (through environment variables or dynamically load at runtime)

But I think that, at least for now, the first two options should be enough in most use-cases?

hug-dev (commented):
Amazing, that looks very clear 👌 A few comments/questions

tss-esapi-sys/Cargo.toml
ionut-arm: Uhm, and what happens if I need to `link` to multiple libraries? I think we're doing it wrong and should have one crate per library that we link with.......

tss-esapi-sys/README.md
ionut-arm: I've not checked, tbh, but I do remember things being different when building on the RaspberryPi

tss-esapi/Cargo.toml
ionut-arm: `bindgen` is no longer needed in the `tss-esapi` crate*

tss-esapi-sys/build.rs
ionut-arm: Yeah, make sense, I've removed the submodule - keeping just generation from the system headers

tss-esapi-sys/Cargo.toml
hug-dev: > I think we're doing it wrong and should have one crate per library that we link with.......
Good point, I don't know if ppl would make use of a `sys` crate + abstraction of those other libraries we use. But we can still do it in the future.

tss-esapi-sys/README.md
hug-dev: ah ok, better safe than sorry

ionut-arm: I think there are changes every now and then on the headers, don't know if even on the same version of the API or if they release new versions when they make a change

hug-dev (approved):
Thanks!

puiterwijk (approved):
👍  Awesome, thanks!

tss-esapi/src/structures/lists/digest_values.rs
```diff
@@ -24,14 +24,40 @@ impl DigestValues {
     }
 }
 
+fn digest_to_ha(digest: (HashingAlgorithm, Digest)) -> Result<TPMT_HA> {
```
Superhepper, line 27: What is the reason for this move from hash/agile?

ionut-arm: Yes, I forgot to mention that - had to change it because `TPMT_HA` is now defined in a different crate, and because there was a different generic implementation of the `TryFrom`, that I wasn't sure how to solve apart from switching to this

Superhepper: I wonder if it would have been better to create a specific type instead of the tuple or if that would just have given rise to more problems.

Superhepper: I am ok with things as they are. But it is a little bit inconsistent to have a function doing the conversion instead of a try_from.

Superhepper (approved):

tss-esapi/src/structures/lists/digest_values.rs
ionut-arm: Done! I've created a new structure, `HashAgile` in the place of the previous `TryFrom`

PR #179, opened 2021-01-27
Split the Context into files per the TPM spec categorization

puiterwijk:


src/abstraction/transient.rs
```diff
@@ -700,7 +700,7 @@ impl TransientKeyContextBuilder {
         context.set_sessions((Some(session), None, None));
 
         let root_key_handle = context
-            .create_primary_key(
+            .create_primary(
                 self.hierarchy,
```
Superhepper, line 704: Wonderful to see the names being what you expect them to be!

puiterwijk:
So, the one new thing I did in this was the `// Missing function: ` comments. Other than that, all of the code is literally yanked+put from `context.rs`

Superhepper (commented):
This all looks great! The only thing that bugs me slightly is the name ```context_func```. To me it would have been more appropriate to either:
1. Append all files with _apis and put them directly in the context folder.
2. Or rename the ```context_func``` dir to apis.

puiterwijk:
Sure, I'm fine with `apis`. This was just the first I came up with. In the end, it doesn't matter at all 😄

puiterwijk:
Or maybe `crate::context::tpm_commands`? Since it's not all the APIs on `Context`.

src/abstraction/transient.rs
puiterwijk: Agreed, this has always bugged me!

ionut-arm (approved):
I'm ok with whichever folder names you decide to go with!

Superhepper (approved):
Sure tpm_commands works better.

PR #180, opened 2021-01-28
Make doc tests run and use env TCTI

puiterwijk:

### DIFF
--- a/tss-esapi/src/tcti.rs
+++ b/tss-esapi/src/tcti.rs
@@ -3,6 +3,7 @@
 use crate::{Error, WrapperErrorKind};
 use regex::Regex;
 use std::convert::TryFrom;
+use std::env;
 use std::ffi::CString;
 use std::net::IpAddr;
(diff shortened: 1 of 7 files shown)


tss-esapi/src/tcti.rs
```diff
@@ -36,6 +37,30 @@ pub enum Tcti {
     Tabrmd(TabrmdConfig),
 }
 
+impl Tcti {
+    /// Gets a TCTI from the following environment variables, in order:
+    /// - TPM2TOOLS_TCTI
```
Superhepper, line 42: What is the reasoning behind this order?

I wonder if we should prefix our own environment variables with something to perhaps avoid name collision?

puiterwijk: The order was random, based on what I thought would be okayish. The primary reason for this function is the doctests.

Superhepper (approved):

ionut-arm (approved):

PR #189, opened 2021-02-09
Added the policy_template context method.

Superhepper:

### DIFF
--- a/tss-esapi/src/context/tpm_commands/enhanced_authorization_ea_commands.rs
+++ b/tss-esapi/src/context/tpm_commands/enhanced_authorization_ea_commands.rs
@@ -454,6 +454,71 @@ impl Context {
         }
     }
 
-    // Missing function: PolicyTemplate
+    /// Bind policy to a specific creation template.
+    ///
+    /// # Arguments
+    /// * `policy_session` - The policy session being extended.
+    /// * `template_hash` - The digest to be added to the policy.
+    ///
+    /// # Example
+    /// ```rust
+    /// # use tss_esapi::{
+    /// #   constants::{SessionType, algorithm::{Cipher, HashingAlgorithm}},
+    /// #   Context, Tcti, structures::Digest,
+    /// # };
+    /// # use std::convert::TryFrom;
+    /// // Create context that uses Device TCTI.
+    /// # let mut context = unsafe {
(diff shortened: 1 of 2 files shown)


tss-esapi/src/context/tpm_commands/enhanced_authorization_ea_commands.rs
```diff
@@ -454,6 +454,71 @@ impl Context {
...
+                self.optional_session_3(),
+                &template_hash.clone().into(),
+            )
+        };
+        let ret = Error::from_tss_rc(ret);
+        if ret.is_success() {
+            Ok(())
+        } else {
+            error!(
+                "Failed to bind template has to a specific creation template: {}",
```
ionut-arm, line 517:
```suggestion
                "Failed to bind template to a specific creation template: {}",
```
?

tss-esapi/src/context/tpm_commands/enhanced_authorization_ea_commands.rs
```diff
@@ -454,6 +454,71 @@ impl Context {
...
+    /// #     )
+    /// #     .expect("Failed to create session")
+    /// #     .expect("Recived invalid handle");
+    /// # let template_hash = Digest::try_from(vec![
+    /// #     1, 2, 3, 4, 5, 6, 7, 8, 9, 10,
+    /// #     11, 12, 13, 14, 15, 16, 17, 18, 19, 20
+    /// # ]).expect("Failed to create template hash");
+    /// // Result should be checked because the TPM might not support this.
+    /// let _ = context.policy_template(trial_session, &template_hash)
+    ///     .expect("Call to policy_template failed");
```
ionut-arm, line 495: Hmm, is this example adding much to the documentation? I think if the documentation for the parameters is clear enough (which it seems to be!) and the setup doesn't require complicated steps then we can just skip the example

Superhepper: Nah I think you are right. The only interesting in this example is the comment. But that should be added to the detailed section instead.

puiterwijk (approved):
LGTM

ionut-arm (approved):
Thank you!

PR #190, opened 2021-02-14
Interface types improvement

Superhepper:
This improves the algorithm and ecc interface types.

The specification lists a complete set of existing algorithms that all have an id assigned to them. Each algorithm interface type then specifies a subset of these algorithms. Previously there were types specified in ```constants::algorithm``` that corresponded "well enough" to the interface types in the specification. 

This changes so that in constants lists all available algorithms in the enum ```AlgorithmIdentifer``` and the interface types are moved into the ```interface_types::algorithm``` module. 
The same thing applies to the ECC as well.

Further more this change also moves Cipher into the abstraction module and removes the use of Cipher from the context methods interfaces. New Rust types have been created for the types that context methods needs in their interface.


Superhepper:
The build error on ubuntu is due to in version 2.3.X of tpm2-tss lib some of the names are wrong and in this particular case it is the interface face type for Camellia key bits that is not called ```TPMI_CAMELLIA_KEY_BITS``` but rather ```TPMI_TPM2_CAMELLIA_KEY_BITS```. I am looking into fixing this some how.

tss-esapi/src/abstraction/cipher.rs
```diff
@@ -0,0 +1,241 @@
...
+/// Structure useful for handling an abstract representation of ciphers. Ciphers are
+/// defined foremost through their symmetric algorithm and, depending on the type of that
+/// algorithm, on a set of other values.
+#[derive(Copy, Clone, Debug)]
+pub struct Cipher {
+    algorithm: SymmetricAlgorithm,
+    mode: Option<SymmetricMode>,
+    key_bits: Option<u16>,
+    hash: Option<HashingAlgorithm>,
+}
```
ionut-arm, line 24: Does it make any sense to keep this if we have the `SymmetricDefinition` and `SymmetricDefinitionObject` structs?

tss-esapi/tests/context_tests/tpm_commands/symmetric_primitives_tests.rs
```diff
@@ -61,9 +57,11 @@ mod test_hmac {
         let key_pub = Tpm2BPublicBuilder::new()
             .with_type(TPM2_ALG_KEYEDHASH)
             .with_name_alg(TPM2_ALG_SHA256)
-            .with_parms(PublicParmsUnion::KeyedHashDetail(KeyedHashParms::HMAC {
-                hash_alg: HashingAlgorithm::Sha256,
-            }))
+            .with_parms(PublicParmsUnion::KeyedHashDetail(KeyedHashParameters::new(
+                KeyedHashScheme::Hmac {
+                    hmac_scheme: HmacScheme::new(HashingAlgorithm::Sha256),
+                },
```
ionut-arm, line 63: Do we use this `KeyedHashScheme` a lot, or do we expect it to be used a lot? I can see a `const` implemented on it (like `AES_256_CFB` for ciphers), something like `HMAC_SHA256`

tss-esapi/src/structures/tagged/schemes.rs
```diff
@@ -0,0 +1,57 @@
...
+/// Enum representing the keyed hash scheme.
+///
+/// # Details
+/// This corresponds to TPMT_SCHEME_KEYEDHASH.
+#[derive(Clone, Copy, Debug)]
+pub enum KeyedHashScheme {
+    Xor { xor_scheme: XorScheme },
+    Hmac { hmac_scheme: HmacScheme },
+    Null,
+}
```
ionut-arm, line 19: Nice 👌🏻  I like these new types!

tss-esapi/src/interface_types/algorithm.rs
```diff
@@ -0,0 +1,466 @@
...
+}
+
+impl TryFrom<AlgorithmIdentifier> for KeyedHashSchemeAlgorithm {
+    type Error = Error;
+    fn try_from(algorithm_identifier: AlgorithmIdentifier) -> Result<Self> {
+        match algorithm_identifier {
+            AlgorithmIdentifier::Hmac => Ok(KeyedHashSchemeAlgorithm::Hmac),
+            AlgorithmIdentifier::Xor => Ok(KeyedHashSchemeAlgorithm::Xor),
+            AlgorithmIdentifier::Null => Ok(KeyedHashSchemeAlgorithm::Null),
+            _ => Err(Error::local_error(WrapperErrorKind::InvalidParam)),
```
ionut-arm, line 106: I'm wondering if at some point we should make these error codes more expressive. Right now, whether the first conversion from `TPMI_ALG_KEYEDHASH_SCHEME` fails or this one fails, we get the same `InvalidParam`, right? But this isn't really important for this PR.

It's really nice to have the conversion implemented like this, though, so we don't have duplicate stuff 💯

tss-esapi/src/interface_types/algorithm.rs
```diff
@@ -0,0 +1,466 @@
...
+            SignatureScheme::RsaPss => Ok(AsymmetricAlgorithm::Rsa),
+            SignatureScheme::EcDsa => Ok(AsymmetricAlgorithm::Ecc),
+            SignatureScheme::EcDaa => Ok(AsymmetricAlgorithm::Ecc),
+            SignatureScheme::Sm2 => Ok(AsymmetricAlgorithm::Ecc),
+            SignatureScheme::EcSchnorr => Ok(AsymmetricAlgorithm::Ecc),
+            _ => {
+                // HMAC is for symmetric algorithms
+                //
+                // Null could be converted into AsymmetricAlgorithm::Null
+                // but I do not know if that is usefull.
```
ionut-arm, line 408: Leave it as is - can change if we get a request

ionut-arm (approved):
Thanks! 🥇 I left a few comments below but I'm ok with the changes overall.

For future PRs, could you try keeping them somewhat smaller? One of the reasons why I struggle to review is because some PRs end up quite large and it takes a lot of time in one go

tss-esapi/src/abstraction/cipher.rs
Superhepper: I think yes. Because it is one type that can generate them both so that you avoid specify the same thing twice which can be nice.

Superhepper:
> 
> 
> Thanks! 🥇 I left a few comments below but I'm ok with the changes overall.
> 
> For future PRs, could you try keeping them somewhat smaller? One of the reasons why I struggle to review is because some PRs end up quite large and it takes a lot of time in one go

Yes I am sorry about that. I should not have added the Lincense headers here. Even without it changing interface types will affect a lot of files.

ionut-arm:
> 
> Yes I am sorry about that. I should not have added the Lincense headers here. Even without it changing interface types will affect a lot of files.

That's alright, I didn't even realise they weren't there 😱 thanks for noticing and fixing it!

tss-esapi/src/interface_types/algorithm.rs
Superhepper: Yes, I liked the the double conversions it makes it more clear what goes wrong. The same goes for AlgorithmIdentifier.

puiterwijk (approved):
Looks good to me, thanks!

tss-esapi/tests/context_tests/tpm_commands/symmetric_primitives_tests.rs
Superhepper: Not really sure but does not hurt to add it.

PR #191, opened 2021-02-17
Enable running of hmac doctest

puiterwijk:
The PR adding hmac coincided with the one enabling the doctests, so
let's make sure the hmac one also runs.


Superhepper (approved):
I think it looks good. I would just like to avoid the unwrap and use expect instead to explain what went wrong.

ionut-arm (approved):

PR #193, opened 2021-02-26
ak: Ephemeral key / Protect against replay attacks

baloo:
To give credit where credit is due:
 > https://safeboot.dev/attestation/#why-is-generating-a-quote-so-slow
 > By creating an ephemeral AK (with the stclear bit set in the
 > attributes), the TPM will not allow it to be persisted and will refuse
 > to reload it when the reboot counter increments.

The downside is that it's slow as one need to generate a key every time.

### DIFF
--- a/tss-esapi/src/abstraction/ak.rs
+++ b/tss-esapi/src/abstraction/ak.rs
@@ -32,6 +32,9 @@ fn create_ak_public(
         .with_fixed_tpm(true)
         .with_fixed_parent(true)
         .with_sensitive_data_origin(true)
+        // Prevent replay attacks.
+        // Ensure the key can't be reused after a TPM reset/reboot.
+        .with_st_clear(true)
         .build()?;
 
     match key_alg {


baloo:
I can provide an alternative implementation to give the user a parameter to opt-in the `stclear`. but I don't see a use-case where `stclear` is not required, so I'm pushing it as a default.

ionut-arm (approved):
It sounds like a sensible change to me, but @puiterwijk is the owner here so I'll let him express relevant opinions. 

Alternatively, if you're looking to create an abstraction for a specific use case that doesn't work with what's already here, you can create a new file in `abstraction` that matches your needs.

puiterwijk:
So, the problem I have with this is that I explicitly took the set of attributes from the [tpm2-tools](https://github.com/tpm2-software/tpm2-tools/blob/master/tools/tpm2_createak.c#L136), to maintain compatibility with keys generated that way.
If people want to add other attributes, they can easily do this as well. Or we can add an argument that does this, that's fine with me if it's documented accurately.

baloo:
Well, that makes sense and I don't want to break that.

What would you think about something like:
```
diff --git a/tss-esapi/src/abstraction/ak.rs b/tss-esapi/src/abstraction/ak.rs
index 9efb1b5..6c17a68 100644
--- a/tss-esapi/src/abstraction/ak.rs
+++ b/tss-esapi/src/abstraction/ak.rs
@@ -19,22 +19,39 @@ use crate::{
     Context, Error, Result, WrapperErrorKind,
 };

-fn create_ak_public(
+pub trait KeyCustomization {
+    fn attributes(&self, attributes_builder: ObjectAttributesBuilder) -> ObjectAttributesBuilder {
+        attributes_builder
+    }
+    fn template(&self, template_builder: Tpm2BPublicBuilder) -> Tpm2BPublicBuilder {
+        template_builder
+    }
+}
+
+fn create_ak_public<K: KeyCustomization>(
     key_alg: AsymmetricAlgorithm,
     hash_alg: HashingAlgorithm,
     sign_alg: SignatureScheme,
+    key_customization: Option<K>,
 ) -> Result<TPM2B_PUBLIC> {
-    let obj_attrs = ObjectAttributesBuilder::new()
+    let obj_attrs_builder = ObjectAttributesBuilder::new()
         .with_restricted(true)
         .with_user_with_auth(true)
         .with_sign_encrypt(true)
         .with_decrypt(false)
         .with_fixed_tpm(true)
         .with_fixed_parent(true)
-        .with_sensitive_data_origin(true)
-        .build()?;
+        .with_sensitive_data_origin(true);

-    match key_alg {
+    let obj_attrs_builder = if let Some(ref k) = key_customization {
+        k.attributes(obj_attrs_builder)
+    } else {
+        obj_attrs_builder
+    };
+
+    let obj_attrs = obj_attrs_builder.build()?;
+
+    let key_builder = match key_alg {
         AsymmetricAlgorithm::Rsa => Tpm2BPublicBuilder::new()
             .with_type(TPM2_ALG_RSA)
             .with_name_alg(hash_alg.into())
@@ -82,8 +99,15 @@ fn create_ak_public(
                 },
             }))
             .with_unique(PublicIdUnion::Ecc(Box::new(Default::default()))),
-    }
-    .build()
+    };
+
+    let key_builder = if let Some(ref k) = key_customization {
+        k.template(key_builder)
+    } else {
+        key_builder
+    };
+
+    key_builder.build()
 }

 /// This loads an Attestation Key previously generated under the Endorsement hierarchy
@@ -135,16 +159,17 @@ pub fn load_ak(
 }

 /// This creates an Attestation Key in the Endorsement hierarchy
-pub fn create_ak(
+pub fn create_ak<K: KeyCustomization>(
     context: &mut Context,
     parent: KeyHandle,
     hash_alg: HashingAlgorithm,
     sign_alg: SignatureScheme,
     ak_auth_value: Option<&Auth>,
+    key_customization: Option<K>,
 ) -> Result<CreateKeyResult> {
     let key_alg = sign_alg.get_key_alg();

-    let ak_pub = create_ak_public(key_alg, hash_alg, sign_alg)?;
+    let ak_pub = create_ak_public(key_alg, hash_alg, sign_alg, key_customization)?;

     let session = match context.start_auth_session(
         None,
```

I care mostly about the AK, but I could generalize that throughout (`create_restricted_decryption_rsa_public` and others  as well as EK)

puiterwijk:
That looks good to me.

baloo:
Closed in favor of #197

Issue #186, opened 2021-02-03
Investigate if context methods are using incorrect types.


Superhepper:
As I saw when I was trying to write the documentation for the NV context methods. They were actually using the wrong interface types for the auth_handle values. And I suspect this is quite common throughout the context methods. Once upon a time before the interface types existed the different handle types were created and they do not contain the same restrictions as the interface types does.

So I think this requires some investigation to see which methods that should be using interface types instead of what they are currently taking as argument.

Superhepper:
One thing I found directly is that ```policy_signed``` should not take ```auth_object: ObjectHandle``` it should instead take ```auth_object: interface_types::dynamic_handles::Object```.

PR #195, opened 2021-02-26
Changed Session into an interface type.

Superhepper:
This change moves the Session type into interface types and renames
it to AuthSession to bring it closer to the specification(see #186).
It also adds two new session types HmacSession and PolicySession.

The current session type used in the context policy methods can contain
a type of session that is not valid to use with those methods. According to the
specification the interface type to use in those methods should be the PolicySession.
Creating and using a correct interface type for use with these methods makes it
harder to provide incorrect arguments to the methods.(#186).


tss-esapi/src/abstraction/transient.rs
```diff
@@ -733,7 +733,7 @@ impl TransientKeyContextBuilder {
             })?;
         if let (Some(old_session), _, _) = context.sessions() {
             context.set_sessions((Some(new_session), None, None));
-            context.flush_context(old_session.handle().into())?;
```
ionut-arm, line 736: How come the new session types can't have a `handle()` method?

tss-esapi/tests/common/mod.rs
```diff
@@ -172,7 +173,9 @@ pub fn get_pcr_policy_digest(
         .unwrap();
 
     if do_trial {
-        let pcr_session = context
+        // Create a trial policy session to use for calls the policy
+        // context methods.
```
ionut-arm, line 177:
```suggestion
        // Create a trial policy session to use for calls in the policy
        // context methods.
```
?

ionut-arm: I think there's one more comment like this below - I'm guessing that was what you intended to write?

tss-esapi/src/abstraction/transient.rs
Superhepper: Of course they can I just thought it was nicer to be able to convert them into one. Even though I am not so found of the double conversion.

tss-esapi/src/interface_types/session_handles.rs
```diff
@@ -0,0 +1,204 @@
...
+/// Enum representing an policy session interface type
+///
+/// # Details
+/// This corresponds to TPMI_SH_POLICY but provides more
+/// information regarding the parameters used when the policy session
+/// was created.
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub enum PolicySession {
+    PolicySession(PolicySessionData),
+}
```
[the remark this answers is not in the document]
Superhepper: The reason for the enum is to be consistent with the rest of the interface types. I think almost all of them are enums even the single value ones.

tss-esapi/src/abstraction/transient.rs
Superhepper: Maybe it would be better to add a conversion directly to an ObjectHandle?

In the long term I hope to be able to get rid of all the ESYS handle types because they are not really correct. The specification says that ```flush_context``` should take a TPMI_DH_CONTEXT but I am not sure if the CONTEXT, OBJECT and ENTITY interface types makes much sense in ESAPI layer.

tss-esapi/src/interface_types/session_handles.rs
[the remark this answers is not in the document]
Superhepper: Oh now I understand what you mean. I have fixed that.

tss-esapi/src/interface_types/session_handles.rs
```diff
@@ -0,0 +1,155 @@
...
+
+/// Enum representing an policy session interface type
+///
+/// # Details
+/// This corresponds to TPMI_SH_POLICY but provides more
+/// information regarding the parameters used when the policy session
+/// was created.
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub enum PolicySession {
+    PolicySession {
```
ionut-arm, line 17: Still think it's a bit superfluous, but I'm ok with this as the final form

ionut-arm (approved):

tss-esapi/src/interface_types/session_handles.rs
Superhepper: I actually agree with you and I was going back and forth with different approaches. This approach had to me two advantages:
1. It makes the interface types consistent (they are all enums)
2. And it makes them resemble how they are defined in the specification.

ionut-arm:
@puiterwijk - Would you like to have a look? We're thinking of including this in the next version release (along with the parallel support for 2.4.x and 2.3.3)

tss-esapi/tests/context_tests/general_esys_tr_tests.rs
```diff
@@ -20,7 +20,6 @@ mod test_tr_from_tpm_public {
     fn remove_nv_index_handle_from_tpm(nv_index_tpm_handle: NvIndexTpmHandle, nv_auth: Provision) {
         let mut context = create_ctx_without_session();
         let mut property = TPM2_NV_INDEX_FIRST;
-        println!("KALLEE");
```
hug-dev, line 23: 🤣

hug-dev (approved):
I mostly trust @ionut-arm 's review but the changes seem ok for me! More precise types.

tss-esapi/tests/context_tests/general_esys_tr_tests.rs
Superhepper: Hahaha

PR #197, opened 2021-02-27
ak, ek: allows use to pass options to key creation

baloo:
This introduces a KeyCustomization arguments to create_ek and create_ak,
this allows a consumer of this API to alter the way a key would be
created.

### DIFF
--- a/tss-esapi/src/abstraction/mod.rs
+++ b/tss-esapi/src/abstraction/mod.rs
@@ -6,3 +6,24 @@ pub mod cipher;
 pub mod ek;
 pub mod nv;
 pub mod transient;
+
+use crate::attributes::ObjectAttributesBuilder;
+use crate::utils::Tpm2BPublicBuilder;
+
+/// KeyCustomizaion allows to adjust how a key is going to be created
+pub trait KeyCustomization {
+    /// Alter the attributes used on key creation
+    fn attributes(&self, attributes_builder: ObjectAttributesBuilder) -> ObjectAttributesBuilder {
+        attributes_builder
+    }
+
+    /// Alter the key template used on key creation
+    fn template(&self, template_builder: Tpm2BPublicBuilder) -> Tpm2BPublicBuilder {
+        template_builder
+    }
+}
+
+/// DefaultKey provides a blanket implementation of KeyCustomization
+#[derive(Copy, Clone, Debug)]
+pub struct DefaultKey;
+impl KeyCustomization for DefaultKey {}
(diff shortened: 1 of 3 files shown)


tss-esapi/src/abstraction/mod.rs
```diff
@@ -6,3 +6,32 @@ pub mod cipher;
...
+// ```
+// ak::create_ak(ctx, handle, Sha256, SignatureScheme, auth, None)
+// ```
+// as rust type checker will complain that it can't infer type for None.
+// To mitigate for that, provide a default implementation for (). This is
+// meant to be used like:
+// ```
+// ak::create_ak::<()>(ctx, handle, Sha256, SignatureScheme, auth, None)
+// ```
+impl KeyCustomization for () {}
```
Superhepper, line 37: I am not so sure I like that the default or 'I do not care about customization' name is ```()``` couldn't it be called ```DefaultKeyCustomization``` or something like that?

Superhepper (commented):
Could you add a test where the customization is used so we can see what it looks like.

tss-esapi/src/abstraction/mod.rs
baloo: I'm not entirely happy about that solution either.
I went with `()` as it's easy to type in. My concern was to not introduce too much burden on the user, and thought `create_ak::<DefaultKeyCustomization>(_,_,_,_,_, None)` would undermine the usability.

baloo:
I'm not too sure how to implement a test for this one yet. I would need to create an ak/ek and tests the attributes afterward I guess.

tss-esapi/src/abstraction/mod.rs
baloo: Here is another proposal. Relying on two methods instead. This avoids headaches with type inference.

baloo: Another alternative would be:
https://play.rust-lang.org/?version=nightly&mode=release&edition=2018&gist=17fc37fc6d2ad309b0c52e4208bbf233

Superhepper: Is it not possible to define it with a default implementation?
```pub fn foo<K: KeyCustomization = DefaultKeyCustomization>(.............).```

baloo: No :( it has been deprecated :(
https://github.com/rust-lang/rust/issues/36887

Superhepper:
@baloo should this one be closed?

baloo:
well, this one and #199 are two API  proposal. I think both valid, just a matter of opinion. I prefer the #199 one but it's very much up to you!

Superhepper:
> 
> 
> well, this one and #199 are two API proposal. I think both valid, just a matter of opinion. I prefer the #199 one but it's very much up to you!

Hehe I preferred the other one as well.

PR #199, opened 2021-02-28
ak, ek: allows use to pass options to key creation

baloo:
This introduces a KeyCustomization argument to create_ek and create_ak,
this allows a consumer of this API to alter the way a key would be
created.

An alternative implementation to #197

### DIFF
--- a/tss-esapi/src/abstraction/ak.rs
+++ b/tss-esapi/src/abstraction/ak.rs
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::{
-    abstraction::cipher::Cipher,
+    abstraction::{cipher::Cipher, IntoKeyCustomization, KeyCustomization},
     attributes::{ObjectAttributesBuilder, SessionAttributesBuilder},
     constants::{tss::*, SessionType},
     handles::{AuthHandle, KeyHandle},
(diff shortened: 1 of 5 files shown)


baloo:
Example usage: https://github.com/baloo/reproducibility-lab/blob/d2b7905932fb1ed3ccc15a4431dc58753be6c16e/pkgs/pcr-eventlog-attestation/src/client/mod.rs#L202-L208

Superhepper (approved):
Thanks for the changes.

Yes, this is much better.

tss-esapi/src/abstraction/ak.rs
```diff
@@ -19,22 +19,31 @@ use crate::{
 use log::error;
 use std::convert::{TryFrom, TryInto};
 
-fn create_ak_public(
+fn create_ak_public<IKC: IntoKeyCustomization>(
     key_alg: AsymmetricAlgorithm,
     hash_alg: HashingAlgorithm,
     sign_alg: SignatureScheme,
+    key_customization: IKC,
```
ionut-arm, line 26: I've not participated in the previous discussion, so apologies if these things were already discussed.

So from what I understand, `KeyCustomization` is essentially used to allow the caller to pass two functions that take in the `ObjectAttributesBuilder` and the `Tpm2BPublicBuilder` and do stuff with them (i.e. set or reset parameters). It feels more natural to me (though this is obviously subjective) to pass those two functions as `Option`s of [closures](https://doc.rust-lang.org/rust-by-example/fn/closures/input_parameters.html). Then the caller doesn't need to create a new type and implement a trait, just need to create a `Some(|builder| { ... })` when they call this function (and you can pass just one, without the other). 

An alternative would be to create some type that the builders accept as a method (something like `.with_customization(Option<...>)`) and which you can pass here, though that might be difficult to pin down (I assume the reason you didn't go down this route is because for the TPM2B Public builder, the number of options and customizations is... kinda annoying :) ).

What do you think? I'm ok with the current changes, they just seem a bit over-engineered.

ionut-arm (commented):
Sorry for the huge delay, not the lightest of periods lately!

tss-esapi/src/abstraction/ak.rs
ionut-arm: @Superhepper @puiterwijk thoughts?

Superhepper: I like both ideas. I like being able to create a specific key type that implements the key customization trait because it makes it clear what you are trying to achieve. But I also like the simplicity in having it as closure options.

ionut-arm (approved):
Ok, let's go with it like this

Issue #200, opened 2021-03-01
Problems with sessions in tests


Superhepper:
I was doing some testing on why it was not possible to run some tests twice. 

And I found that there are several 'potential' problems with running the tests on a non CI vm(i.e. locally).
1. On my machine and I found that some calls to ESAPI required the session to be flushed and recreated between calls. I found that one such bug that required to the session to be flushed was fixed in 2.3.3 and 2.4.0 ([](https://github.com/tpm2-software/tpm2-tss/issues/1574). But this one did not quite match what tests are doing.

What I have found causing this bug is to first call ```create_ctx_with_session``` then afterwards create a new session. What happens then is that the first created session becomes an optional session handle in the call to create the next session. When we run integration tests this does not seem to be a problem and I have no idea why.

I have been searching through the specification but not really found any conclusive answer.

PR #201, opened 2021-03-01
This is an attempt to fix tests that fails locally.

Superhepper:
This change tries to address the issue described in #200 

Sometimes when trying to run tests locally they fail with
some cryptic message that a new auth session could not be
created due to authorization failure.

The cause behind this is that one session is being used
as an optional session for creating another session
whether this should be allowed or not is not quite clear.
Until this have been fully investigated this change remove
this behaviour.

### DIFF
--- a/tss-esapi/tests/context_tests/tpm_commands/object_commands_tests.rs
+++ b/tss-esapi/tests/context_tests/tpm_commands/object_commands_tests.rs
@@ -205,38 +205,36 @@ mod test_activate_credential {
         let (session_attributes, session_attributes_mask) = SessionAttributesBuilder::new().build();
 
         let session_1 = context
-            .start_auth_session(
-                None,
-                None,
-                None,
-                SessionType::Hmac,
-                SymmetricDefinition::AES_256_CFB,
-                HashingAlgorithm::Sha256,
-            )
-            .unwrap();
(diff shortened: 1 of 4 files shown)


tss-esapi/tests/context_tests/tpm_commands/object_commands_tests.rs
```diff
@@ -251,23 +249,26 @@ mod test_activate_credential {
...
+        context.set_sessions((Some(session_1), Some(session_2), None));
 
         let decrypted = context
             .activate_credential(key_handle, key_handle, credential_blob, secret)
-            .unwrap();
+            .expect("Call to active_credential failed");
 
-        let expected = Digest::try_from(vec![1, 2, 3, 4, 5]).unwrap();
+        let expected =
+            Digest::try_from(vec![1, 2, 3, 4, 5]).expect("Failed to create digest for expected");
```
ionut-arm, line 271: Not looking for you to change this, but I think the `expect` instead of `unwrap` here doesn't add much because we know it won't fail - I'd say the cases where failing is unlikely/impossible can stay `unwrap` in general

ionut-arm (approved):
Changes look fine to me!

puiterwijk (approved):
Looks good to me!

PR #206, opened 2021-03-11
Add cross-compilation example to nightly

### LINKED ISSUE #204: Support and test cross-compilation
That is, documenting and testing the necessary steps so that: `cargo build --target aarch64-unknown-linux-gnu` works.
I think there are multiple things:
1. generating the bindings for the set of target that we want to support (I think it's good to start with `aarch64-unknown-linux-gnu` and `armv7-unknown-linux-gnueabihf`)
2. when the TSS libs are also cross-compiled and installed somewhere in the system, `pkg-config` (both the crate and the executable) need to be configured correctly to find the correct version (the cross-compiled and not the host one)

For 2., see the relevent links:

* https://docs.rs/pkg-config/0.3.19/src/pkg_config/lib.rs.html#330-353

```rust
        // pkg-config may not be aware of cross-compilation, and require
        // a wrapper script that sets up platform-specific prefixes.
        match self.targetted_env_var("PKG_CONFIG_ALLOW_CROSS") {
            // don't use pkg-config if explicitly disabled
            Some(ref val) if val == "0" => false,
            Some(_) => true,
            None => {
                // if not disabled, and pkg-config is customized,
                // then assume it's prepared for cross-compilation
                self.targetted_env_var("PKG_CONFIG").is_some()
                    || self.targetted_env_var("PKG_CONFIG_SYSROOT_DIR").is_some()
            }
        }
```

* https://docs.rs/pkg-config/0.3.19/pkg_config/

> PKG_CONFIG_ALLOW_CROSS - The pkg-config command usually doesn't support cross-compilation, and this crate prevents it from selecting incompatible versions of libraries. Setting PKG_CONFIG_ALLOW_CROSS=1 disables this protection, which is likely to cause linking errors, unless pkg-config has been configured to use appropriate sysroot and search paths for the target platform.

* https://autotools.io/pkgconfig/cross-compiling.html: describing the wrapper script to use

We should at least document the steps to do to perform cross-compilation, in the easiest way possible, and maybe check0in the wrapper script (not fully sure about that).

ionut-arm:
This commit adds cross-compilation to armv7 and aarch64 to the nightly
build. This involves modifications to the way `pkg-config` is used
as described in #204.

Fixes #204

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -41,13 +41,15 @@ jobs:
         run: cargo audit
 
   coverage:
-    name: Calculate code coverage
+    name: Calculate code coverage and cross compile
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
       - name: Build the container
         run: docker build -t ubuntucontainer tss-esapi/tests/ --file tss-esapi/tests/Dockerfile-ubuntu
-      - name: Run the container
+      - name: Run the code coverage script
         run: docker run -v $(pwd):/tmp/rust-tss-esapi -w /tmp/rust-tss-esapi/tss-esapi --security-opt seccomp=unconfined ubuntucontainer /tmp/rust-tss-esapi/tss-esapi/tests/coverage.sh
       - name: Collect coverage results
         run: bash <(curl -s https://codecov.io/bash)
+      - name: Run the cross-compilation script
+        run: docker run -v $(pwd):/tmp/rust-tss-esapi -w /tmp/rust-tss-esapi/tss-esapi ubuntucontainer /tmp/rust-tss-esapi/tss-esapi/tests/cross-compile.sh
--- a/tss-esapi-sys/README.md
+++ b/tss-esapi-sys/README.md
@@ -39,3 +39,21 @@ is included in the name of the file - if the triplet you require is not
 available, feel free to raise a Pull Request to add it or to use build-time
 generation of bindings. All the committed bindings **MUST** be generated from
 the library version found under the `vendor` submodule.
+
+## Cross compiling
+
+Cross-compilation can be done as long as you have on your build system the TSS 
+libraries compiled for your target system of choice. We rely on `pkg-config` to
+identify the libraries which we link against. Installing `tpm2-tss` does yield
+`.pc` files which can be used for this purpose, but depending on the exact build
+environment setup, the configuration and compilation of `tpm2-tss` could require
+some special tailoring.
+
+We include cross-compilation builds as a nightly check in Github Actions - you
+can find them [here](../tss-esapi/tests/cross-compile.sh) as an example of the
+steps needed. You can find more information on using `pkg-config` when cross-compiling
+[here](https://github.com/parallaxsecond/rust-tss-esapi/issues/204). Our wrapper
+script around `pkg-config` can be seen [here](../tss-esapi/tests/pkg-config).
+
+Be advised that in some cases the linker used might need to be set manually in
+`.cargo/config`.
--- a/tss-esapi/tests/cross-compile.sh
+++ b/tss-esapi/tests/cross-compile.sh
@@ -1,0 +1,83 @@
+#!/usr/bin/env bash
+
+# Copyright 2021 Contributors to the Parsec project.
+# SPDX-License-Identifier: Apache-2.0
+
+# Cross compile the `tss-esapi` crate (and its dependencies) for Armv7 and Aarch64
+# In order to cross-compile the TSS library we need to also cross-compile OpenSSL
+
+set -euf -o pipefail
+
+OPENSSL_VERSION="OpenSSL_1_1_1j"
+
+cross-compile-openssl() {
+    # Prepare directory for cross-compiled OpenSSL files
+    mkdir -p /tmp/openssl-$1
+    export INSTALL_DIR=/tmp/openssl-$1
+
+    pushd /tmp/openssl
+    # Compile and copy files over
+    ./Configure $2 shared --prefix=$INSTALL_DIR --openssldir=$INSTALL_DIR/openssl --cross-compile-prefix=$1-
+    make clean
+    make depend
+    make -j$(nproc)
+    make install
+    popd
+
+    export INSTALL_DIR=
+}
+
+cross-compile-tpm2-tss() {
+    # Prepare directory for cross-compiled TSS lib
+    # `DESTDIR` is used in `make install` below to set the root of the installation paths.
+    # The `./configure` script accepts a `--prefix` input variable which sets the same root,
+    # but also adds it to the paths in `.pc` files used by `pkg-config`. This prevents the 
+    # use of `PKG_CONFIG_SYSROOT_DIR`.
+    mkdir -p /tmp/tpm2-tss-$1
+    export DESTDIR=/tmp/tpm2-tss-$1
+    # Set sysroot to be used by the `pkg-config` wrapper
+    export SYSROOT=/tmp/tpm2-tss-$1
+
+    pushd /tpm2-tss
+    # Compile and copy files over
+    ./configure --build=x86_64-pc-linux-gnu --host=$1 --target=$1 CC=$1-gcc \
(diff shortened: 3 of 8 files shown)


ionut-arm:
NOTE: I've updated the `x86_64-unknown-linux-gnu` bindings, as the ones generated previously were not based on v2.3.3 of the tss libs 😬

.github/workflows/nightly.yml
```diff
@@ -41,13 +41,15 @@ jobs:
...
       - uses: actions/checkout@v2
       - name: Build the container
         run: docker build -t ubuntucontainer tss-esapi/tests/ --file tss-esapi/tests/Dockerfile-ubuntu
-      - name: Run the container
+      - name: Run the code coverage script
         run: docker run -v $(pwd):/tmp/rust-tss-esapi -w /tmp/rust-tss-esapi/tss-esapi --security-opt seccomp=unconfined ubuntucontainer /tmp/rust-tss-esapi/tss-esapi/tests/coverage.sh
       - name: Collect coverage results
         run: bash <(curl -s https://codecov.io/bash)
+      - name: Run the cross-compilation script
+        run: docker run -v $(pwd):/tmp/rust-tss-esapi -w /tmp/rust-tss-esapi/tss-esapi ubuntucontainer /tmp/rust-tss-esapi/tss-esapi/tests/cross-compile.sh
```
hug-dev, line 55: Is it useful to still use the usual container? Since we have to cross-compile what we would use in that container, could we use a slimer container instead?
Although that way we also replicate an environment where both TSS lib for host + target are installed which is nice to test.

And it is very valuable to have a self-contained script which shows how to do it as well. So let's keep it this way!

tss-esapi-sys/.cargo/config
```diff
@@ -0,0 +1,7 @@
+# Cargo seems to expect the -gcc variant instead of -ld
```
hug-dev, line 1: Is this file needed? I only put it in Parsec because `cargo` was using the system linker (`cc`) and it was failing but for rust-psa-crypto and parsec-se-driver, cross compilation worked fine without.

tss-esapi-sys/.cargo/config
```diff
@@ -0,0 +1,7 @@
...
+
+[target.armv7-unknown-linux-gnueabi]
+linker = "arm-linux-gnueabihf-gcc"
```
hug-dev, line 4: Maybe it should use `arm-linux-gnueabi-gcc`, without the `hf`, if that exists.

tss-esapi-sys/README.md
```diff
@@ -38,4 +38,22 @@ NOTE: Only a limited set of bindings are committed and their target triplet
 is included in the name of the file - if the triplet you require is not
 available, feel free to raise a Pull Request to add it or to use build-time
 generation of bindings. All the committed bindings **MUST** be generated from
-the library version found under the `vendor` submodule.
\ No newline at end of file
+the library version found under the `vendor` submodule.
+
+## Cross compiling
```
hug-dev, line 43: Great section 👌

tss-esapi/tests/cross-compile.sh
```diff
@@ -0,0 +1,81 @@
...
+export PKG_CONFIG_ALLOW_CROSS=1
+# Make the `pkg-config` crate use our wrapper
+export PKG_CONFIG=$(pwd)/tests/pkg-config
+
+# Compile OpenSSL and TSS stack for aarch64-unknown-linux-gnu
+cross-compile-openssl aarch64-linux-gnu linux-generic64
+cross-compile-tpm2-tss aarch64-linux-gnu
+
+rustup target add aarch64-unknown-linux-gnu
+cargo build --features generate-bindings --target aarch64-unknown-linux-gnu
```
hug-dev, line 74: Should we remove the `--features generate-bindings` to check that the committed bindings work?

hug-dev (commented):
Thanks! Very nice work! The README.md and script will be very useful for users.

tss-esapi-sys/.cargo/config
ionut-arm: True, I'll check if it works without

tss-esapi-sys/src/bindings/x86_64-unknown-linux-gnu.rs
```diff
@@ -3306,13 +3282,13 @@ impl Default for TPMS_AUTH_RESPONSE {
 }
 pub type TPMI_AES_KEY_BITS = TPM2_KEY_BITS;
 pub type TPMI_SM4_KEY_BITS = TPM2_KEY_BITS;
-pub type TPMI_CAMELLIA_KEY_BITS = TPM2_KEY_BITS;
+pub type TPMI_TPM2_CAMELLIA_KEY_BITS = TPM2_KEY_BITS;
```
Superhepper, line 3285: I know we are supporting version 2.3.3 and unfortunately we then have to deal with these kind of bugs. I wonder if it would be possible to bump the minimum supported tpm2-tss version soon.

Superhepper (approved):
My knowledge of cross compilation is a bit limited with that in mind I have really nothing against how it looks now. One thing that seemed odd to me was how the cross compilation of the tpm2-tss lib was dependent on OpenSSL. Should it not be the other way around?

Another thing is that the 2.3 versions of tpm2-tss lib have a few naming errors that will be a little bit annoying to deal with (like it better when it was generated from 2.4.X and we didn't tell any one hehehe). Not a big deal everything is U32 in the end any way.

tss-esapi-sys/src/bindings/x86_64-unknown-linux-gnu.rs
ionut-arm: Hmm, yeah, we're going to tread carefully on that on the Parsec side - I'll try and think of a way of bumping the minimum supported version in the crate without burning down support for Parsec. 

Potentially keeping the 2.3.3 version supported in a separate branch and move on with `master`? (it wouldn't involve porting every new feature to it, only stuff that we in Parsec need, and important bugfixes)

puiterwijk: I don't think that forking the crate is going to be great for anyone.

ionut-arm: @puiterwijk - is there some standard way of making something like this work, that you know of? Or are we going to end up in hell either way we try

@Superhepper - what version did you have in mind? >=2.4.0 or straight up 3.0.0?

puiterwijk: Well, we could go back to the idea of having a feature for which version of the TSS you're using, and then change implementation details based on that (maybe just have both of them in a side-module).

Superhepper: No, straight up 3.0 would not be advisable because I think there still a lot of people running 2.x. But some 2.4 release would be nice.

puiterwijk: (And then, based on the version during install, put in a fake version of the new constant, so that we can just assume the lateest version exists, but maybe it comes from bindgen, maybe from our own version)

ionut-arm: I've raised #207 - will start looking at it soon

ionut-arm:
> 
> 
> My knowledge of cross compilation is a bit limited with that in mind I have really nothing against how it looks now. One thing that seemed odd to me was how the cross compilation of the tpm2-tss lib was dependent on OpenSSL. Should it not be the other way around?

Sorry, forgot to reply to this - it does seem a bit weird that `tpm2-tss` depends on OpenSSL, but the reason behind that is the need for in-memory crypto functionality in TSS. For example, encryption of arguments needs some crypto primitives - they've chosen not to re-implement them, but pull them from OSSL.

tss-esapi-sys/.cargo/config
ionut-arm: Removed it completely.

hug-dev (approved):
Looks good for me!

PR #208, opened 2021-03-18
Prepare the new crates versions

hug-dev:
This is to upload `tss-esapi` to `4.0.11-alpha.1` and `tss-esapi-sys` to
`0.1.0`.

### DIFF
--- a/tss-esapi/Cargo.toml
+++ b/tss-esapi/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "tss-esapi"
-version = "4.0.11-alpha.1"
+version = "5.0.0"
 authors = ["Parsec Project Contributors"]
 edition = "2018"
 description = "Rust-native wrapper around TSS 2.0 Enhanced System API"
--- a/tss-esapi/Cargo.toml
+++ b/tss-esapi/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "tss-esapi"
-version = "4.0.10-alpha.2"
+version = "4.0.11-alpha.1"
 authors = ["Parsec Project Contributors"]
 edition = "2018"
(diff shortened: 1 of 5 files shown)


tss-esapi/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "tss-esapi"
-version = "4.0.10-alpha.2"
+version = "4.0.11-alpha.1"
```
ionut-arm, line 3:
```suggestion
version = "4.0.10-alpha.3"
```

hug-dev: is that for breaking changes? Not just bug fixes?

Superhepper (approved):

tss-esapi/Cargo.toml
Superhepper: Well moving stuff around breaks others code. 

I must say it is slightly odd that we increase minor version on something that is yet to be released. Would make more sense to just 4.0.0-alpha.x all the way until we are happy with the alpha.

ionut-arm: Yeah, we should've been increasing the alpha version from the beginning 😓

hug-dev: As I said on the channel, would it be that bad to release a `5.0.0`?? If we are not making releases that often now. Is there a big difference if our real "stable" is `54.0.0` instead of `5.0.0`?

puiterwijk (approved):

hug-dev:
All the maintainers [agreed on Slack](https://cloud-native.slack.com/archives/C01EARH2ZB3/p1616087764010900) that `5.0.0` was fine, changing to that version and removing the `alpha` prefix.

ionut-arm (approved):
Let's DO this 🚀

PR #209, opened 2021-03-22
Fix cross-compile permissions

ionut-arm:
`cross-compile.sh` did not have the execute permissions set, so the nightly run failed because of it.


hug-dev (approved):
😲

Superhepper (approved):

PR #213, opened 2021-03-25
Prepare patch release to fix docs.rs build

hug-dev:


Superhepper (approved):

ionut-arm (approved):

ionut-arm:
The changes worked for cross-compilation at least, hopefully docs.rs is sorted too.

hug-dev:
It worked 🥳 
https://docs.rs/tss-esapi-sys/0.1.1/tss_esapi_sys/

ionut-arm:
Actually - I'm a bit worried/curious. That's the binding for `x86-unknown-linux-gnu` but there's clearly a difference between that and what a user running on Aarch64 Linux would see. Isn't that common for FFI crates in docs.rs? I don't remember if there was anything about it in that blogpost about doing FFI in Rust.

hug-dev:
On docs.rs you can use the "Platform" option at the top to change what platform the docs have been generated for. We can customize that for the targets we want. [It's very easy](https://github.com/rust-embedded/cortex-m/commit/51dfa4d25f54705b2a7efb7fca08062b00bb1891)

ionut-arm:
Oh wow, that's quite nice! Maybe we should do that for our crates

PR #217, opened 2021-03-31
Bump bindgen to 0.57.0

eclipseo:


ionut-arm (approved):
👍🏻 Would you like another release that includes this change?

@hug-dev - we'll need to update the bindgen used for `psa-crypto` 😁 `cryptoki` is already using this version

hug-dev (approved):

eclipseo:
> 👍🏻 Would you like another release that includes this change?
> 
Not needed for me, Ihave patched it locally for now so it can wait when you feel ready wrt your planned changes.

PR #218, opened 2021-04-01
Fixes

ionut-arm:
I think the best course of action would be to merge non-breaking changes into the `5.x.y` branch and keep `main` syncrhonised by merging any breaking changes into it, but also continuously rebasing from `5.x.y`. Does that sound sensible?

### DIFF
--- a/tss-esapi/src/context/tpm_commands/capability_commands.rs
+++ b/tss-esapi/src/context/tpm_commands/capability_commands.rs
@@ -4,7 +4,7 @@ use crate::{
     constants::CapabilityType, structures::CapabilityData, tss2_esys::*, utils::PublicParmsUnion,
     Context, Error, Result, WrapperErrorKind as ErrorKind,
 };
-use log::error;
+use log::{error, warn};
 use mbox::MBox;
 use std::convert::{TryFrom, TryInto};
 use std::ptr::null_mut;
@@ -76,7 +76,7 @@ impl Context {
         if ret.is_success() {
             Ok(())
         } else {
-            error!("Error while testing parameters: {}", ret);
+            warn!("Parameters under test could not be unmarshalled: {}", ret);
             Err(ret)
         }
     }
--- a/tss-esapi/src/lib.rs
+++ b/tss-esapi/src/lib.rs
@@ -28,6 +28,7 @@
     unused_results,
     missing_copy_implementations
 )]
+#![allow(clippy::upper_case_acronyms)]
 //! # TSS 2.0 Rust Wrapper over Enhanced System API
 //! This crate exposes the functionality of the TCG Software Stack Enhanced System API to
 //! Rust developers, both directly through FFI bindings and through more Rust-tailored interfaces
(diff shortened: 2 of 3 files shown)


tss-esapi/src/context/tpm_commands/capability_commands.rs
```diff
@@ -76,7 +76,7 @@ impl Context {
         if ret.is_success() {
             Ok(())
         } else {
-            error!("Error while testing parameters: {}", ret);
+            warn!("Parameters under test could not be unmarshalled: {}", ret);
```
hug-dev, line 79: 👌 Perfect!

tss-esapi/src/lib.rs
```diff
@@ -28,6 +28,7 @@
     unused_results,
     missing_copy_implementations
 )]
+#![allow(clippy::upper_case_acronyms)]
```
hug-dev, line 31: Could we add a separate commit to fix the acronyms?

hug-dev (commented):
I think it's better if we just target all PRs at `main` and then when we want to make a `5.x.y` release we cherry pick from `main` the non-breaking commits to the `5.x.y` branch. It might happen that we are not going to do any release in a while so might never have to do that!

ionut-arm:
>  when we want to make a `5.x.y` release we cherry pick from `main` the non-breaking commits to the `5.x.y` branch.

That doesn't work very well if you end up doing breaking and non-breaking changes alternatively, because the cherry-picking will fail. And even just from the long-term perspective of the main branch commits, it makes sense to have the "post-`5.x.y`" commits stacked on top of the ones for `5.x.y`

tss-esapi/src/lib.rs
ionut-arm: Done

hug-dev:
> if you end up doing breaking and non-breaking changes alternatively, because the cherry-picking will fail. 

Why? If `main` contain well separated commits I think you will be able to cherry-pick them to the release branch without too many issues.

> And even just from the long-term perspective of the main branch commits, it makes sense to have the "post-5.x.y" commits stacked on top of the ones for 5.x.y

From my point of view, I think it's fine if breaking and non-breaking commits are done alternatively.

My main points is just that it seems easier for all contributors to make their PR against `main` and then us cherry-picking from `main` to `5.x.y` when we need to do a non-breaking release. Otherwise are we going to ask contributors to choose the correct branch? That seems complicated, especially if we are not going to make a new `5.x.y` release often.

tss-esapi/src/lib.rs
hug-dev: I was thinking about actually replacing the acronyms with snake case but that's good for now!

ionut-arm: I think this "separate branch for 5.x.y" releases is going to be painful. I mean, if I replace the acronyms, that change needs to go into `main`, not in `5.x.y`. I still need to allow the lint on the `5.x.y` branch.

Also this is an example of why having alternative commits won't work - if I fix the acronyms in the `main` branch and then do a "non-breaking" commit where I use one of the acronyms somewhere, if I make that commit on top of `main` I'll be using the new version of the acronyms, whereas if I cherry-pick in `5.x.y` the commit I need it to be using the current version

Superhepper (approved):

hug-dev (approved):

PR #226, opened 2021-04-23
Link to tss2-sys as well

ionut-arm:
This commit adds tss2-sys to the list of libraries we explicitly link
to. The reason is that Parsec was having issues linking correctly for
cross-compilation, and explicitly linking helped.


hug-dev (approved):
👍

Superhepper (approved):

PR #227, opened 2021-05-06
Add tss2-sys when generating bindings too

ionut-arm:
The tss2-sys is only being linked to explicitly when the bindings are
not generated, after #226. This fixes that discrepancy.

### DIFF
--- a/tss-esapi-sys/build.rs
+++ b/tss-esapi-sys/build.rs
@@ -56,6 +56,10 @@ fn main() {
 
 #[cfg(feature = "generate-bindings")]
 pub fn generate_from_system(esapi_out: PathBuf) {
+    pkg_config::Config::new()
+        .atleast_version(MINIMUM_VERSION)
+        .probe("tss2-sys")
+        .expect("Failed to find tss2-sys library.");
     let tss2_esys = pkg_config::Config::new()
         .atleast_version(MINIMUM_VERSION)
         .probe("tss2-esys")


hug-dev (approved):
In [here](https://github.com/parallaxsecond/rust-tss-esapi/blob/01391d9d8a086c890c79678a337c3e0158a9adee/tss-esapi-sys/Cargo.toml#L13) maybe we should indicate every lib that we link to

ionut-arm:
> 
> 
> In [here](https://github.com/parallaxsecond/rust-tss-esapi/blob/01391d9d8a086c890c79678a337c3e0158a9adee/tss-esapi-sys/Cargo.toml#L13) maybe we should indicate every lib that we link to

It's a bit problematic, because that field is made for _one_ value, see [doc](https://doc.rust-lang.org/cargo/reference/manifest.html#the-links-field). They say [here](https://doc.rust-lang.org/cargo/reference/build-scripts.html#the-links-manifest-key) there should be at most one package per `links` value, but I think it's the other way around as well.

hug-dev:
Ah true, I think actually we discussed that in the post and said that maybe ideally there should be one `-sys` crate per lib. I guess this is fine to leave that for now.

Superhepper:
Why is it only sys? Why is it not needed to link against each individual library in the software stack?

ionut-arm:
> 
> 
> Why is it only sys? Why is it not needed to link against each individual library in the software stack?

I'm not entirely sure, the problem only appeared when I tried to cross-compile Parsec - for some reason doing the same thing that we do in the cross-compilation tests for this crate doesn't work, it complains that the symbols in `tss2-sys` can't be found. I worry that if I ignore it and just link "by hand" to the library in the Parsec builds, every user of this crate might have to do the same for their binaries.

Superhepper (approved):

PR #234, opened 2021-07-02
Add a wrapper above the TCTI Loader library

hug-dev:
Adds wrappers around the TCTI_CONTEXT and TCTI_INFO structures to safely
create and free them.

This is a wrapper around the `libtss2-tctildr.so`. Ideally this would be in its own crates with `-sys` variant but since those libs seem to be shipped together it makes sense to group them. I just put everything in one module here.

This is currently not breaking but I think the goal would be to remove the existing `tcti` module and replace it with that.

The context for this is to be able to check if the TCTI Loading is working, with the `get_info` call, before having to create a transient context. See https://github.com/parallaxsecond/parsec/issues/401

### DIFF
--- a/tss-esapi/src/tcti_ldr.rs
+++ b/tss-esapi/src/tcti_ldr.rs
@@ -1,0 +1,613 @@
+// Copyright 2021 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Wrapper around the TCTI Loader Library interface.
+//! See section 3.5 of the TCG TSS 2.0 TPM Command Transmission Interface(TCTI) API
+//! Specification.
+
+use crate::{Error, Result, WrapperErrorKind};
+use log::error;
+use regex::Regex;
+use std::convert::TryFrom;
+use std::env;
+use std::ffi::CStr;
+use std::ffi::CString;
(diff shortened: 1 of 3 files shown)


tss-esapi/src/tcti_ldr.rs
```diff
@@ -0,0 +1,617 @@
...
+
+impl Drop for TctiInfo {
+    fn drop(&mut self) {
+        unsafe {
+            tss_esapi_sys::Tss2_TctiLdr_FreeInfo(&mut self.tcti_info);
+        }
+    }
+}
+
+/// Placeholder TCTI types that can be used when initialising a `Context` to determine which
```
hug-dev, line 115: From down below everything is copied but changing `Tcti` to `TctiNameConf` to match the spec.

ionut-arm (approved):
Looks good, should we write it down somewhere to not forget to update `Context` to use the new `TctiContext` in `new`?

hug-dev:
> should we write it down somewhere to not forget to update Context to use the new TctiContext in new?

Yes, as soon as this is merged I will create a new issue for us to do that.

puiterwijk (approved):
This looks decent to me.

PR #237, opened 2021-07-14
Abstract execute_With_nullauth_session return type

puiterwijk:
This makes it possible to call any function from
execute_with_nullauth_session, as long as the resulting Error type can
be created from tss_esapi::Error.
This allows an application to call a function that returns myapp::Error,
as long as an implementation From<tss_esapi::Error> exists for that
type.

### DIFF
--- a/tss-esapi/src/context.rs
+++ b/tss-esapi/src/context.rs
@@ -272,10 +272,11 @@ impl Context {
     /// The session attributes for the generated empty session that
     /// is used to execute closure will have the attributes decrypt
     /// and encrypt set.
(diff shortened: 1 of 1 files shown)


ionut-arm (approved):
Does this technically break API compatibility? I think anything that worked before will still work, right?

puiterwijk:
Nope, it doesn't break any compatibility.
Everything that worked before still works: previously it would always hve been `tss_esapi::Error`, and there is definitely an `impl From<tss_esapi::Error> for tss_esapi::Error` 😄

Superhepper (approved):
Yeah this is nice.

PR #238, opened 2021-07-15
Ensure PcrData retains order

puiterwijk:
PcrData would (possibly) reorder and merge the different TPM2B_DIGEST
entries in a TPML_DIGEST.
This is not valid in the case where for example the TPML_DIGEST is the
result of a pcr_read call: in that case, the order of the
TPML_PCR_SELECTION entries and the TPML_DIGEST match, and must remain
matching.
This is important when a digest of all the digests is passed along, like
in the case of a Quote operation.

### DIFF
--- a/tss-esapi/src/utils/mod.rs
+++ b/tss-esapi/src/utils/mod.rs
@@ -21,7 +21,7 @@ use log::error;
 use zeroize::Zeroize;
 
 use serde::{Deserialize, Serialize};
-use std::collections::{BTreeMap, HashMap};
+use std::collections::BTreeMap;
 use std::convert::{TryFrom, TryInto};
 /// Helper for building `TPM2B_PUBLIC` values out of its subcomponents.
 ///
@@ -1196,7 +1196,7 @@ impl<'a> IntoIterator for &'a PcrBank {
 /// hashing algorithm
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct PcrData {
-    data: HashMap<HashingAlgorithm, PcrBank>,
+    data: Vec<(HashingAlgorithm, PcrBank)>,
 }
 
 impl PcrData {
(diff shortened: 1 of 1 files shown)


puiterwijk:
This patch has one small semantical change: `PcrData.pcr_bank` will now only return the first entry if you have for example a `TPML_DIGEST` with `[sha256:10, sha1:10, sha256:16]` instead of the combined `sha256:10,16` when we merged things.
This should not commonly happen, but now this is a more direct match to `TPML_DIGEST`: garbage in, garbage out.

puiterwijk:
I'm also not fully convinced that this is a massive problem, since the attestation server will need to reorder the PCRs anyway, in case more than 8 values were selected.
So as I was submitting this I wondered if it's really needed, but I figured it would be at least getting us closer to the actual `TPML_DIGEST`, so might still be useful.

tss-esapi/src/utils/mod.rs
```diff
@@ -1262,27 +1260,23 @@ impl PcrData {
...
-            // Add the parsed pcr bank for the parsed hashing algorithm.
-            if parsed_pcr_data
-                .data
-                .insert(parsed_hash_algorithm, parsed_pcr_bank)
-                .is_some()
-            {
-                error!("Error trying to insert data into a PcrBank where data have already been inserted");
-                return Err(Error::local_error(WrapperErrorKind::InconsistentParams));
-            }
+            data.push((parsed_hash_algorithm, parsed_pcr_bank));
```
ionut-arm, line 1263: Is it possible this loses track of some invalid cases that the previous version was handling, i.e. having multiple identical hash algorithms?

puiterwijk: No, because this just adds the additional instance, and then re-builds it.
It's totally possible to have `data: [(sha1, _), (sha256, _), (sha1, _)]`, and this will still allow that.

Superhepper:
This looks fine and absolutely correct. I have more and more realised that the order is very often important. I made PcrData in order to get a convenient way to handle the return values from PcrRead but naively thought that order was unimportant.

Superhepper (approved):

ionut-arm (approved):

PR #239, opened 2021-07-23
Remove old Tcti feature

### LINKED ISSUE #229: How to call `Context::new` safely?
The docs for the method `Context::new` states:
     
            * the client is responsible for ensuring that the context can be initialized safely, threading-wise
      
How do we make sure the call is safe? If we have a multithreaded application, should we make sure that only one `Context` is created at a time? 

The way I understand the [ESAPI specs](https://trustedcomputinggroup.org/wp-content/uploads/TSS_TSS-2.0-Enhanced-System-API_V0.9_R03_Public-Review-1.pdf) (see below), `esys_initialize` should always be safe to call, but the same `ESYS_CONTEXT` object must not be shared between threads. 

> State information is stored in contexts provided to the application. These ESAPI contexts areaccessible simultaneously. No two threads are allowed the same ESAPI context simultaneously.

### LINKED ISSUE #235: Use the new TctiContext in the TSS Context
This is to import the new TctiContext (see #234) in the other TSS contexts. After this the `tcti.rs` file could be deleted. This will be a breaking change!

hug-dev:
The code now uses TctiContext and TctiNameConf to initialize contexts.
Also remove the `unsafe` keyword when creating a context using the
TABRMD.

Fix #235 
Fix #229 cc @gaetanww Would that be ok for you?


tss-esapi/src/context.rs
```diff
@@ -119,6 +110,17 @@ impl Context {
         }
     }
 
+    /// Create a new ESYS context based on the TAB Ressource Manager Daemon.
+    /// The TABRMD will make sure that multiple users can use the TPM safely.
+    ///
+    /// # Errors
+    /// * if either `Tss2_TctiLdr_Initiialize` or `Esys_Initialize` fail, a corresponding
+    /// Tss2ResponseCode will be returned
+    pub fn new_tabrmd(tabrmd_conf: TabrmdConfig) -> Result<Self> {
```
hug-dev, line 119: @gaetanww specifically this part, this method forces TABRMD usage but is now safe

Superhepper (approved):

tss-esapi/src/context.rs
gaetanww: Good to have that method, thanks!
Should it be named something like: “with_tabrmd” or “new_with_tabrmd” or “initalize_with_tabrmd” to conform with rust naming guidelines? 
see https://rust-lang.github.io/api-guidelines/predictability.html#c-ctor

ionut-arm (approved):
👍🏻

tss-esapi/src/context.rs
hug-dev: I chose `new_with_tabrmd` 🚀

PR #240, opened 2021-07-26
Remove the unsafe when creating contexts

### LINKED ISSUE #229: How to call `Context::new` safely?

hug-dev:
It was decided in #229 that the unsafe keyword was not used correctly in
order to prevent multiple applications to use TCTI that can't handle
them. Removes the unsafe and replace the Safety notice with a Warning
one.

Fix #229


hug-dev:
@gaetanww This one should do it 😃

ionut-arm (approved):

MattDavis00 (approved):

PR #241, opened 2021-07-28
Creates native rust type for TPM2B_PUBLIC.

Superhepper:
- Removes all use of TPM2B_PUBLIC outside of the Context methods.
- Adds new builders and structures in order to be able to Creates
  the new Public object.

### DIFF
--- a/tss-esapi/src/abstraction/transient.rs
+++ b/tss-esapi/src/abstraction/transient.rs
@@ -13,50 +13,35 @@
 //!
 //! Object contexts thus act as an opaque handle that can, however, be used by the client to seralize
 //! and persist the underlying data.
-use crate::abstraction::cipher::Cipher;
-use crate::attributes::{ObjectAttributesBuilder, SessionAttributesBuilder};
-use crate::constants::tss::*;
-use crate::constants::SessionType;
-use crate::handles::{KeyHandle, SessionHandle};
-use crate::interface_types::{
-    algorithm::HashingAlgorithm, ecc::EccCurve, resource_handles::Hierarchy,
+use crate::{
+    abstraction::cipher::Cipher,
+    attributes::{ObjectAttributesBuilder, SessionAttributesBuilder},
+    constants::{tss::*, SessionType},
+    handles::{KeyHandle, SessionHandle},
+    interface_types::{
+        algorithm::{HashingAlgorithm, PublicAlgorithm, RsaSchemeAlgorithm},
+        ecc::EccCurve,
+        key_bits::RsaKeyBits,
+        resource_handles::Hierarchy,
+    },
+    structures::{
+        Auth, CreateKeyResult, Data, Digest, EccScheme, Public, PublicBuilder, PublicKeyRsa,
+        PublicRsaParametersBuilder, RsaDecryptionScheme, RsaExponent, RsaScheme, Signature,
+        VerifiedTicket,
+    },
+    tcti_ldr::TctiNameConf,
+    tss2_esys::*,
+    utils::{
+        create_restricted_decryption_rsa_public,
+        create_unrestricted_encryption_decryption_rsa_public,
+        create_unrestricted_signing_ecc_public, create_unrestricted_signing_rsa_public,
+        create_unrestricted_signing_rsa_public_with_unique, PublicKey, TpmsContext,
+    },
+    Context, Error, Result, WrapperErrorKind as ErrorKind,
 };
-use crate::structures::{Auth, CreateKeyResult, Data, Digest, PublicKeyRSA, VerifiedTicket};
-use crate::tcti_ldr::TctiNameConf;
-use crate::tss2_esys::*;
-use crate::utils::{
-    self, create_restricted_decryption_rsa_public,
-    create_unrestricted_encryption_decryption_rsa_public, create_unrestricted_signing_ecc_public,
-    create_unrestricted_signing_rsa_public, AsymSchemeUnion, PublicIdUnion, PublicKey,
-    PublicParmsUnion, Tpm2BPublicBuilder, TpmsContext, TpmsRsaParmsBuilder, RSA_KEY_SIZES,
-};
-use crate::Context;
-use crate::{Error, Result, WrapperErrorKind as ErrorKind};
+
 use log::error;
 use std::convert::{TryFrom, TryInto};
-
-#[derive(Copy, Clone, Debug)]
-pub struct RsaExponent {
-    value: u32,
-}
-
-impl Default for RsaExponent {
-    fn default() -> RsaExponent {
-        RsaExponent {
-            value: (1 << 16) + 1,
-        }
-    }
-}
-
-impl RsaExponent {
-    pub fn new(value: u32) -> RsaExponent {
-        RsaExponent { value }
-    }
-
-    pub fn value(&self) -> u32 {
-        self.value
-    }
-}
 
 /// Structure offering an abstracted programming experience.
 ///
@@ -127,7 +112,7 @@ impl TransientKeyContext {
         self.set_session_attrs()?;
         let key_handle = self
             .context
-            .load(self.root_key_handle, out_private, out_public)?;
+            .load(self.root_key_handle, out_private, &out_public)?;
 
         self.set_session_attrs()?;
         let key_context = self.context.context_save(key_handle.into()).or_else(|e| {
@@ -138,27 +123,18 @@ impl TransientKeyContext {
         Ok((key_context, key_auth))
     }
 
-    fn get_public_from_params(&self, params: KeyParams) -> Result<TPM2B_PUBLIC> {
+    fn get_public_from_params(&self, params: KeyParams) -> Result<Public> {
         match params {
             KeyParams::RsaSign {
                 size,
                 scheme,
                 pub_exponent,
-            } => {
-                if !RSA_KEY_SIZES.iter().any(|sz| *sz == size) {
-                    return Err(Error::local_error(ErrorKind::WrongParamSize));
-                }
-
-                Ok(create_unrestricted_signing_rsa_public(
-                    scheme,
-                    size,
-                    pub_exponent,
-                )?)
-            }
+            } => Ok(create_unrestricted_signing_rsa_public(
+                scheme,
+                size,
+                pub_exponent,
+            )?),
             KeyParams::RsaEncrypt { size, pub_exponent } => {
-                if !RSA_KEY_SIZES.iter().any(|sz| *sz == size) {
-                    return Err(Error::local_error(ErrorKind::WrongParamSize));
-                }
                 create_unrestricted_encryption_decryption_rsa_public(size, pub_exponent)
             }
             KeyParams::Ecc { curve, scheme } => {
@@ -180,26 +156,18 @@ impl TransientKeyContext {
     /// `TransientKeyContext::`set_session_attrs`, `Context::load_external_public`,
     /// `Context::context_save`, `Context::flush_context`
     pub fn load_external_rsa_public_key(&mut self, public_key: &[u8]) -> Result<TpmsContext> {
-        if !RSA_KEY_SIZES
-            .iter()
-            .any(|sz| usize::from(*sz) == public_key.len() * 8)
-        {
-            return Err(Error::local_error(ErrorKind::WrongParamSize));
-        }
-        let mut pk_buffer = [0_u8; 512];
-        pk_buffer[..public_key.len()].clone_from_slice(&public_key[..public_key.len()]);
-        let pk = TPMU_PUBLIC_ID {
-            rsa: TPM2B_PUBLIC_KEY_RSA {
-                size: public_key.len().try_into().unwrap(), // should not fail on valid targets, given the checks above
-                buffer: pk_buffer,
-            },
-        };
-        let mut public = create_unrestricted_signing_rsa_public(
-            AsymSchemeUnion::RSASSA(HashingAlgorithm::Sha256),
-            u16::try_from(public_key.len()).unwrap() * 8_u16,
-            0,
+        let rsa_key_bits = RsaKeyBits::try_from(
+            u16::try_from(public_key.len()).map_err(|e| {
+                error!("Failed to convert length of public key to u16: {:?}", e);
+                Error::local_error(ErrorKind::InvalidParam)
+            })? * 8u16,
         )?;
-        public.publicArea.unique = pk;
+        let public = create_unrestricted_signing_rsa_public_with_unique(
+            RsaScheme::create(RsaSchemeAlgorithm::RsaSsa, Some(HashingAlgorithm::Sha256))?,
+            rsa_key_bits,
+            RsaExponent::default(),
+            &PublicKeyRsa::try_from(public_key)?,
+        )?;
         self.set_session_attrs()?;
         let key_handle = self
             .context
@@ -235,28 +203,10 @@ impl TransientKeyContext {
         public_modulus: &[u8],
         public_exponent: RsaExponent,
     ) -> Result<TpmsContext> {
-        if !RSA_KEY_SIZES
-            .iter()
-            .any(|sz| usize::from(*sz) == public_modulus.len() * 8)
-        {
-            return Err(Error::local_error(ErrorKind::WrongParamSize));
-        }
-
         // Check if there is a mismatch between private key and public key sizes.
         if key_prime.len() * 2 != public_modulus.len() {
             return Err(Error::local_error(ErrorKind::WrongParamSize));
         }
-
-        // Handle public key...
-        let mut public_modulus_buffer = [0_u8; 512];
-        public_modulus_buffer[..public_modulus.len()]
-            .clone_from_slice(&public_modulus[..public_modulus.len()]);
-
-        let pub_buffer = TPM2B_PUBLIC_KEY_RSA {
-            size: public_modulus.len().try_into().unwrap(),
-            buffer: public_modulus_buffer,
-        };
-        let pub_id_union = PublicIdUnion::Rsa(Box::from(pub_buffer));
 
         let object_attributes = ObjectAttributesBuilder::new()
             .with_user_with_auth(true)
@@ -265,22 +215,27 @@ impl TransientKeyContext {
             .with_restricted(false)
             .build()?;
 
-        let rsa_parms = TpmsRsaParmsBuilder::new_unrestricted_signing_key(
-            AsymSchemeUnion::RSASSA(HashingAlgorithm::Sha256),
-            u16::try_from(public_modulus.len()).unwrap() * 8_u16,
-            public_exponent.value(),
-        )
-        .build()
-        .unwrap();
+        let key_bits = RsaKeyBits::try_from(
+            u16::try_from(public_modulus.len()).map_err(|e| {
+                error!("Failed to convert public modulus to u16: {:?}", e);
+                Error::local_error(ErrorKind::InvalidParam)
+            })? * 8u16,
+        )?;
 
-        let public = Tpm2BPublicBuilder::new()
-            .with_type(TPM2_ALG_RSA)
-            .with_name_alg(TPM2_ALG_SHA256)
+        let public = PublicBuilder::new()
+            .with_public_algorithm(PublicAlgorithm::Rsa)
+            .with_name_hashing_algorithm(HashingAlgorithm::Sha256)
             .with_object_attributes(object_attributes)
-            .with_parms(PublicParmsUnion::RsaDetail(rsa_parms))
-            .with_unique(pub_id_union)
-            .build()
-            .unwrap();
+            .with_rsa_parameters(
+                PublicRsaParametersBuilder::new_unrestricted_signing_key(
+                    RsaScheme::create(RsaSchemeAlgorithm::RsaSsa, Some(HashingAlgorithm::Sha256))?,
+                    key_bits,
+                    public_exponent,
+                )
+                .build()?,
+            )
+            .with_rsa_unique_identifier(&PublicKeyRsa::try_from(public_modulus)?)
+            .build()?;
 
         // Handle private key...
         let mut key_prime_buffer = [0_u8; 256];
@@ -327,7 +282,7 @@ impl TransientKeyContext {
         let key_handle = self.context.context_load(key_context)?;
 
         self.set_session_attrs()?;
-        let key_pub_id = self
+        let public = self
             .context
             .read_public(key_handle.into())
             .or_else(|e| {
@@ -335,20 +290,13 @@ impl TransientKeyContext {
                 Err(e)
             })?
             .0;
-        let key = match unsafe { PublicIdUnion::from_public(&key_pub_id)? } {
-            // call should be safe given our trust in the TSS library
-            PublicIdUnion::Rsa(pub_key) => {
-                let mut key = pub_key.buffer.to_vec();
-                key.truncate(pub_key.size.try_into().unwrap()); // should not fail on supported targets
-                PublicKey::Rsa(key)
-            }
-            PublicIdUnion::Ecc(pub_key) => {
-                let mut x = pub_key.x.buffer.to_vec();
-                x.truncate(pub_key.x.size.try_into().unwrap()); // should not fail on supported targets
-                let mut y = pub_key.y.buffer.to_vec();
-                y.truncate(pub_key.y.size.try_into().unwrap()); // should not fail on supported targets
-                PublicKey::Ecc { x, y }
-            }
+
+        let key = match public {
+            Public::Rsa { unique, .. } => PublicKey::Rsa(unique.value().to_vec()),
+            Public::Ecc { unique, .. } => PublicKey::Ecc {
+                x: unique.x().value().to_vec(),
+                y: unique.y().value().to_vec(),
+            },
             _ => return Err(Error::local_error(ErrorKind::UnsupportedParam)),
         };
         self.context.flush_context(key_handle)?;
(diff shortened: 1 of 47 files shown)


Superhepper:
I would like to hear any suggestions you have regarding names, types, usability ... well everything. This is just a start to get things going. I have not started with the tests yet and I have not completely removed all the things that should be removed or just moved ...

tss-esapi/src/structures/tagged/signature.rs
```diff
@@ -0,0 +1,125 @@
...
+    EcDsa(EccSignature),
+    EcDaa(EccSignature),
+    Sm2(EccSignature),
+    EcSchnorr(EccSignature),
+    Hmac(HashAgile),
+    Null,
+}
+
+impl Signature {
+    pub const fn algorithm(&self) -> SignatureScheme {
```
ionut-arm, line 26: Probably better to call it `scheme` (?)

ionut-arm (commented):
Thanks for the huge effort put in 🚀 Here's the first batch of comments.

tss-esapi/src/structures/tagged/schemes.rs
```diff
@@ -61,3 +69,424 @@ impl TryFrom<TPMT_KEYEDHASH_SCHEME> for KeyedHashScheme {
...
+        match rsa_scheme_algorithm {
+            RsaSchemeAlgorithm::RsaSsa => Ok(RsaScheme::RsaSsa(HashScheme::new(
+                hashing_algorithm.ok_or_else(|| {
+                    error!(
+                        "Hashing algorithm is required when creating RSA scheme of type RSA SSA"
+                    );
+                    Error::local_error(WrapperErrorKind::ParamsMissing)
+                })?,
+            ))),
+            RsaSchemeAlgorithm::RsaEs => Ok(RsaScheme::RsaEs),
```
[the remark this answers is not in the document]
ionut-arm: This seems to apply to a lot of enum variants, don't know if we want to go down this rabbit hole of always checking the other params are correct... I think we should, but maybe we can set that as a future issue to deal with.

tss-esapi/src/structures/tagged/schemes.rs
```diff
@@ -61,3 +69,424 @@ impl TryFrom<TPMT_KEYEDHASH_SCHEME> for KeyedHashScheme {
...
+    RsaSsa(HashScheme),
+    RsaEs,
+    RsaPss(HashScheme),
+    Oaep(HashScheme),
+    Null,
+}
+
+impl RsaScheme {
+    /// Creates a new RsaScheme
+    pub fn create(
```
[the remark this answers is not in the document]
Superhepper: I use ```new``` when it cannot fail and ```create``` when a Result is returned.

tss-esapi/src/structures/tagged/signature.rs
Superhepper: Yeah, I think you are right. Will rename the scheme and the algorithm.

tss-esapi/src/structures/tagged/schemes.rs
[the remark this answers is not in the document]
Superhepper: Yeah, I think you are right. First I thought that it might be enough to just give a warning. But then I reconsidered and liked the idea better to give an error.

tss-esapi/src/structures/tagged/signature.rs
```diff
@@ -0,0 +1,125 @@
+use crate::{
```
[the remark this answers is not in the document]
Superhepper: Fixed

tss-esapi/tests/abstraction_ak_tests.rs
```diff
@@ -162,14 +162,14 @@ fn test_create_custom_ak() {
...
-        SignatureScheme::RsaPss,
+        SignatureSchemeAlgorithm::RsaPss,
         Some(&ak_auth),
         None,
     )
     .unwrap();
 
     assert_eq!(
-        att_key_without.out_public.publicArea.objectAttributes
+        att_key_without.out_public.object_attributes().0
```
ionut-arm, line 172: Think it would be nice to implement a getter for that value stored in `ObjectAttributes`, just so it's more easy to understand what's going on. We could just throw it in a new issue, I guess

tss-esapi/src/structures/tagged/schemes.rs
```diff
@@ -61,3 +69,438 @@ impl TryFrom<TPMT_KEYEDHASH_SCHEME> for KeyedHashScheme {
...
+            RsaDecryptAlgorithm::RsaEs => Ok(RsaDecryptionScheme::RsaEs),
+            RsaDecryptAlgorithm::Oaep => Ok(RsaDecryptionScheme::Oaep(HashScheme::new(
+                hashing_algorithm.ok_or_else(|| {
+                    error!("Hashing algorithm is required when creating RSA decrypt scheme of type OEAP");
+                    Error::local_error(WrapperErrorKind::ParamsMissing)
+                })?,
+            ))),
+            RsaDecryptAlgorithm::Null => Ok(RsaDecryptionScheme::Null),
+        }
+    }
```
ionut-arm, line 459: Actually, I just realised, because these are enums there's no easy way to block the variants and their content from being public, so `create` isn't that efficient (because you have to check for conditions that are more easily enforced when an enum variant is directly created - for example, for `RsaDecryptAlgorithm::Null` you don't check that `hashing_algorithm` is `None`), but maybe a `validate` method would be good if there's anything to validate. For example, in this case it seems like any hash works with `Oaep`, but otherwise `validate` would've been a good way to implement the checks.

ionut-arm (commented):
A few more comments here and there

tss-esapi/src/structures/buffers/public/rsa.rs
```diff
@@ -0,0 +1,361 @@
...
+    }
+
+    /// Method that returns the value of the rsa exponent.
+    pub const fn value(&self) -> u32 {
+        self.value
+    }
+
+    /// Private method for checking if a value is valid.
+    fn is_valid(value: u32) -> bool {
+        if 3 > value || value == 0 {
```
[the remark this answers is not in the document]
Superhepper: *Trusted Platform Module Library 
Part 2: Structures 
Family “2.0” 
Level 00 Revision 01.59 
November 8, 2019 
Section 12.2.3.5  TPMS_RSA_PARMS *
>A TPM compatible with this specification and supporting RSA shall support two primes and an exponent 
>of zero. An exponent of zero indicates that the exponent is the default of 2^16 + 1. Support for other values 
>is optional. Use of other exponents in duplicated keys is not recommended because the resulting keys 
>would not be interoperable with other TPMs. 

*Table 196 — Definition of {RSA} TPMS_RSA_PARMS Structure *
>| exponent | U32 | The public exponent, < prime number greater than 2. 

So you are right that is wrong.

tss-esapi/src/structures/tagged/schemes.rs
Superhepper: No, it is not possible to block 'bad' combinations if one creates the variants manually. So ```create``` is a convenience function that validates the combination. Yeah, it might be a good idea to split them up in two functions so it would be possible to validate a manually created variant.

tss-esapi/tests/abstraction_ak_tests.rs
Superhepper: Hrrm there should be an Into implementation ... for ObjectAttributes.

tss-esapi/src/structures/ecc/point.rs
```diff
@@ -0,0 +1,57 @@
...
+    pub const fn y(&self) -> &EccParameter {
+        &self.y
+    }
+}
+
+impl Default for EccPoint {
+    fn default() -> Self {
+        EccPoint::new(EccParameter::default(), EccParameter::default())
+    }
+}
```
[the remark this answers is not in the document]
Superhepper: Nah not that I know. I think I made this when I needed to initialize a structure that was not going to be used.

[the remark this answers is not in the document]
Superhepper: Maybe chose a better name for it would be good.

tss-esapi/src/structures/buffers/public/rsa.rs
```diff
@@ -0,0 +1,361 @@
...
+    type Error = Error;
+
+    fn try_from(tpm_uint32_value: UINT32) -> Result<Self> {
+        if RsaExponent::is_valid(tpm_uint32_value) {
+            Ok(RsaExponent {
+                value: tpm_uint32_value,
+            })
+        } else {
+            error!(
+                "Received invalid tsa exponent value {}, from the TPM",
```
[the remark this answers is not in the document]
Superhepper: fixed

tss-esapi/src/structures/buffers/public/rsa.rs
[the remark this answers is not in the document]
Superhepper: Fixed

tss-esapi/src/structures/buffers/public/rsa.rs
```diff
@@ -0,0 +1,361 @@
...
+        Ok(RsaExponent { value })
+    }
+
+    /// Method that returns the value of the rsa exponent.
+    pub const fn value(&self) -> u32 {
+        self.value
+    }
+
+    /// Private method for checking if a value is valid.
+    fn is_valid(value: u32) -> bool {
```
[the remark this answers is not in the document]
Superhepper: Sure

tss-esapi/tests/abstraction_ak_tests.rs
Superhepper: There is ... I will change this.

Superhepper: Fixed

tss-esapi/src/structures/buffers/public/ecc.rs
```diff
@@ -0,0 +1,270 @@
...
+            && ecc_scheme.algorithm() != EccSchemeAlgorithm::Sm2
+            && ecc_scheme.algorithm() != EccSchemeAlgorithm::EcDh
+            && ecc_scheme.algorithm() != EccSchemeAlgorithm::EcMqv
+            && ecc_scheme.algorithm() != EccSchemeAlgorithm::Null
+        {
+            return Err(Error::local_error(WrapperErrorKind::InconsistentParams));
+        }
+
+        if (ecc_curve == EccCurve::BnP256 || ecc_curve == EccCurve::BnP638)
+            && ecc_scheme.algorithm() != EccSchemeAlgorithm::EcDaa
```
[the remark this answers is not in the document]
Superhepper: It looks like this in the code that was used previously:
https://github.com/parallaxsecond/rust-tss-esapi/blob/cc9a08e22f4ae88eb66c96fe601d8b1d058d00b6/tss-esapi/src/utils/mod.rs#L470
```rust
        if (self.curve == EccCurve::BnP256 || self.curve == EccCurve::BnP638)
            && scheme.scheme != TPM2_ALG_ECDAA
        {
            return Err(Error::local_error(WrapperErrorKind::InconsistentParams));
        }
```

And I did not change the logic I just changed the types used. Do you want me to change it?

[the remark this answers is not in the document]
ionut-arm: No, it's fine, the change is so large that it's difficult to tell what's new and what isn't 😅

tss-esapi/src/structures/ecc/point.rs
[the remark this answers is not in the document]
ionut-arm: Ah, I see, if it's not new then there's no worry!

tss-esapi/src/structures/buffers/public/rsa.rs
```diff
@@ -0,0 +1,357 @@
...
+    }
+
+    /// Method that returns the value of the rsa exponent.
+    pub const fn value(&self) -> u32 {
+        self.value
+    }
+
+    /// Function for checking if a value is valid rsa exponent.
+    pub const fn is_valid(value: u32) -> bool {
+        // TODO: Verify that `value` is a prime number
```
ionut-arm, line 247: I don't think the exponent has to be prime, just coprime with some other value. I think the commonly used values are primes because those work for any sensible RSA private key.

This also means we can't really check whether the value is valid (unless it is prime, but that's not strictly necessary).

tss-esapi/src/structures/tagged/schemes.rs
ionut-arm: Ok, that's fair enough - I'm wondering how strict we want to be with the parameters. If we get two parameters that are inconsistent (because one isn't needed), do we just reject the call, or do we pretend the one that shouldn't have been there doesn't exist?

Superhepper: Personally I want it to be really really hard to create something that seem to work but doesn't. So I like to be able to catch errors as soon as possible and be really strict to avoid errors later which only will make the TPM come back with one of its useless errors like 'wrong size associated with parameter 2'.

tss-esapi/src/structures/buffers/public/rsa.rs
Superhepper: Hrrm section 12.2.3.5 in the structures part of the spec says "the public exponent. A prime number greater then 2". Have I interpreted this in the wrong way?

ionut-arm: Oh, it might be that the TPM puts this restriction, I was thinking of RSA in general. In that case it could make sense. Don't know how we'd do it - maybe there are generic implementations out there, or maybe we can store a list of all primes of at most 32 bits.

tss-esapi/src/structures/buffers/public.rs
```diff
@@ -0,0 +1,520 @@
...
+    /// Returns the object attributes
+    pub const fn object_attributes(&self) -> ObjectAttributes {
+        match self {
+            Public::Rsa {
+                object_attributes,
+                name_hashing_algorithm: _,
+                auth_policy: _,
+                parameters: _,
+                unique: _,
+            } => *object_attributes,
```
ionut-arm, line 331:
```suggestion
            Public::Rsa { object_attributes, .. } => *object_attributes,
```
At least I hope that's the syntax

tss-esapi/src/structures/buffers/public.rs
```diff
@@ -0,0 +1,520 @@
...
+            error!("ObjectAttributes is required and has not been set in the PublicBuilder");
+            Error::local_error(WrapperErrorKind::ParamsMissing)
+        })?;
+
+        let name_hashing_algorithm = self.name_hashing_algorithm.ok_or_else(|| {
+            error!(
+                "The name hashing algorithm is required and has not been set in the PublicBuilder"
+            );
+            Error::local_error(WrapperErrorKind::ParamsMissing)
+        })?;
```
ionut-arm, line 216: Really nice logging and error management

ionut-arm (commented):
Ok, I think I've reviewed all the new stuff now, need to review the refactored bits

tss-esapi/src/structures/buffers/public.rs
Superhepper: Thank you I was looking for that but I couldn't find it.

tss-esapi/src/structures/buffers/public.rs
```diff
@@ -0,0 +1,520 @@
...
+                    })?,
+                })
+            },
+            PublicAlgorithm::KeyedHash => {
+                Ok(Public::KeyedHash {
+                    object_attributes,
+                    name_hashing_algorithm,
+                    auth_policy,
+                    parameters: self.keyed_hash_parameters.ok_or_else(|| {
+                        error!("Keyed hash parameters have not ben set in the Public Builder even though the keyed hash algorithm have been selected");
```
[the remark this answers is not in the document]
Superhepper: Thanks fixed it now.

tss-esapi/src/structures/buffers/public.rs
Superhepper: Thank you.

tss-esapi/src/structures/buffers/public.rs
```diff
@@ -0,0 +1,520 @@
...
+
+use ecc::PublicEccParameters;
+use keyed_hash::PublicKeyedHashParameters;
+use rsa::PublicRsaParameters;
+
+use log::error;
+use std::convert::{TryFrom, TryInto};
+use tss_esapi_sys::{TPMU_PUBLIC_ID, TPMU_PUBLIC_PARMS};
+
+/// A builder for the [Public] type.
```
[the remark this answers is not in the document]
Superhepper: Yeah that is a good idea. I have added it to the builder documentation.

tss-esapi/src/structures/buffers/public.rs
Superhepper: I have fixed so they look like this now. Very nice.

tss-esapi/src/structures/tagged/schemes.rs
Superhepper: I have updated so that it gives an error for if a hashing algorithm have been provided when none is expected in a lot of places.

tss-esapi/src/structures/buffers.rs
```diff
@@ -199,7 +201,56 @@ pub mod nonce {
...
+        pub fn new_empty_with_size(rsa_key_bits: RsaKeyBits) -> Self {
+            match rsa_key_bits {
+                RsaKeyBits::Rsa1024 => PublicKeyRsa(vec![0u8; 128].into()),
+                RsaKeyBits::Rsa2048 => PublicKeyRsa(vec![0u8; 256].into()),
+                RsaKeyBits::Rsa4096 => PublicKeyRsa(vec![0u8; 512].into()),
+            }
+        }
+    }
+
+    impl TryFrom<PublicKeyRsa> for [u8; 1024] {
```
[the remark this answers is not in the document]
Superhepper: Now I cannot remember where I got this from. But I think I got it from some where and I thought it looked rather strange. Like if they intended for each bit to be stored in a byte or something. Well I changed it now any way. We shall see if I can remember why did it like that.

tss-esapi/src/structures/buffers/public/rsa.rs
Superhepper: I added the primal crate as dependency in order to do the prime number checking for us.

tss-esapi/src/context/tpm_commands/object_commands.rs
```diff
@@ -154,7 +149,7 @@ impl Context {
                 self.optional_session_2(),
                 self.optional_session_3(),
                 private,
-                public,
+                &(*public).clone().into(),
```
ionut-arm, line 152: That looks a bit weird, is the `$(*...)` part needed? Would've thought a simple `public.clone()` was enough

ionut-arm (commented):
Only tests left to review!

tss-esapi/src/abstraction/transient.rs
```diff
@@ -327,28 +282,33 @@ impl TransientKeyContext {
...
-            }
+
+        let key = match public {
+            Public::Rsa {
+                object_attributes: _,
+                name_hashing_algorithm: _,
+                auth_policy: _,
+                parameters: _,
+                unique,
+            } => PublicKey::Rsa(unique.value().to_vec()),
```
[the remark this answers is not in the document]
Superhepper: Seems to work

tss-esapi/src/context/tpm_commands/object_commands.rs
Superhepper: I had some problems with this. It created a clone of a reference to a public which was not what I wanted.

Superhepper: I fixed it now.

ionut-arm (approved):
Looks good! Thanks for the huge effort!

One last queestion I had was whether 3072 was a valid RSA key size, I thought that was in the list as well, though I forgot where precisely it was defined.

Superhepper:
> 
> 
> Looks good! Thanks for the huge effort!
> 
> One last queestion I had was whether 3072 was a valid RSA key size, I thought that was in the list as well, though I forgot where precisely it was defined.
> 
> EDIT: The Architecture spec, Annex B, says that the reference implementation supports 1024, 2048 and 3072.

It is defined here:
https://github.com/parallaxsecond/rust-tss-esapi/blob/58458400a86a176c33cd56cbca08051e88703482/tss-esapi/src/utils/mod.rs#L383

I will look into it and add it as key size in the RsaKeyBits if I do not find anything strange.

ionut-arm (approved):
Looks good to me! Thanks for the titanic effort involved in this 🚀

ionut-arm (commented):
Had one comment, see below, but I'm keen on merging this as is and fixing it in my upcoming PR, if you agree

tss-esapi/src/structures/buffers/public.rs
```diff
@@ -0,0 +1,501 @@
...
+        self
+    }
+
+    /// Adds the symmetric cipher parameters for the [Public] structure
+    /// to the builder.
+    ///
+    /// # Details
+    /// This is required if the public algorithm is set to
+    /// [SymCipher][`crate::interface_types::algorithm::PublicAlgorithm::SymCipher].
+    pub const fn with_symmetric_cipher_parameters(
```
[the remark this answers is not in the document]
Superhepper: I will fix that. We should always try to take arguments that cannot be copied by reference.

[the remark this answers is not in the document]
Superhepper: But there are inconsistencies in there. Most of the parameters can be copied(if one should be able to copy them is another story) so they are fine to take by value. Though it is not possible to copy EccPoint so it should be taken by reference. So I will have to look at this more to fix the stuff.

tss-esapi/src/structures/buffers/public/rsa.rs
```diff
@@ -0,0 +1,356 @@
...
+
+    /// Method that returns the value of the rsa exponent.
+    pub const fn value(&self) -> u32 {
+        self.value
+    }
+
+    /// Function for checking if a value is valid rsa exponent.
+    pub fn is_valid(value: u32) -> bool {
+        (value > 2 && primal::is_prime(value.into())) || value == 0
+    }
```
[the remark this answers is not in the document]
Superhepper: I have added a few unit tests for this now.

tss-esapi/src/structures/buffers/public.rs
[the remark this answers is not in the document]
Superhepper: I have now changed so that the types that cannot be copied are taken by reference and the rest by value is this the way we want to go our should they all be taken by reference?

tss-esapi/src/structures/buffers/public/ecc.rs
```diff
@@ -0,0 +1,270 @@
...
+        self
+    }
+
+    /// Adds a flag that inidcates if the key is going to be restrictied to
+    /// the [PublicEccParametersBuilder].
+    ///
+    /// # Arguments
+    /// * `set` - `true` indicates that it is going to be a restricted key.
+    ///           `false` indicates that it is going to be a non restricted key.
+    pub const fn with_restricted(mut self, set: bool) -> Self {
```
wiktor-k, line 121: Shouldn't this be `with_is_restricted` to mirror `with_is_signing_key` and `with_is_decryption_key`? :thinking:

tss-esapi/src/structures/tagged/parameters.rs
```diff
@@ -0,0 +1,88 @@
...
+impl PublicParameters {
+    /// Returns the algorithm
+    pub const fn algorithm(&self) -> PublicAlgorithm {
+        match self {
+            PublicParameters::Rsa(_) => PublicAlgorithm::Rsa,
+            PublicParameters::KeyedHash(_) => PublicAlgorithm::KeyedHash,
+            PublicParameters::Ecc(_) => PublicAlgorithm::Ecc,
+            PublicParameters::SymCipher(_) => PublicAlgorithm::SymCipher,
+        }
+    }
```
ionut-arm, line 34: I'm having some issues with these `const fn` methods that contain `match` statements, because they were introduced in 1.46.0 and we're trying to keep Parsec compatible with 1.43.0. There are a few methods like these scattered across the structures, my question was: do you need them as `const`? If yes, I'll change them in my PR to conditionally compile them and add non-`const` for older compiler versions.

tss-esapi/src/structures/buffers.rs
```diff
@@ -199,7 +201,68 @@ pub mod nonce {
...
+    impl TryFrom<PublicKeyRsa> for [u8; 128] {
+        type Error = Error;
+
+        fn try_from(value: PublicKeyRsa) -> Result<Self> {
+            value
+                .value()
+                .try_into()
+                .map_err(|_| Error::local_error(WrapperErrorKind::WrongParamSize))
+        }
+    }
```
ionut-arm, line 232: Similarly, these fail compilation because `TryFrom<&[u8]> for [u8; ...]` (at least for this size of the array) is quite recent, but for this one I can just disable compilation for older versions.

tss-esapi/src/structures/tagged/parameters.rs
Superhepper: No, I do not need them as const. I just like to as much as possible as const (guess I am to influenced by c++ constexpr).

ionut-arm: It would be good to have them `const`, but alas we can't use it (for now, for Parsec) :( We can make them `const` again later when our MSRV changes, or I can try and work around it!

tss-esapi/src/structures/buffers.rs
Superhepper: I will fix this by rewriting the try_from functions for those array sizes.

tss-esapi/src/structures/tagged/parameters.rs
Superhepper: I have changed them for now so you should be able to use it.

tss-esapi/src/structures/buffers.rs
Superhepper: I have rewritten this now so it compiles with 1.43.0

PR #246, opened 2021-08-04
Implement Send/Sync on TctiContext

ionut-arm:
This commit adds Send and Sync implementations on TctiContext which
would allow Context and all structures depending on it to be
thread-safe.


hug-dev (approved):
👌

Superhepper (approved):

PR #248, opened 2021-08-19
Add duplication command

wiktor-k:
Hello :wave: 

I'm working on TPM duplication/import feature and would like to gradually propose changes if you don't mind.

The current code compiles but the test fails with `Error when performing duplication: authorization handle is not correct for command`. Any suggestions welcome... and I'll debug this in parallel.

Edit: this is still raw but I'm updating the PR as I go...

### DIFF
--- a/tss-esapi/src/context/tpm_commands/duplication_commands.rs
+++ b/tss-esapi/src/context/tpm_commands/duplication_commands.rs
@@ -1,9 +1,335 @@
 // Copyright 2021 Contributors to the Parsec project.
 // SPDX-License-Identifier: Apache-2.0
 use crate::Context;
+use crate::{
+    handles::ObjectHandle,
+    structures::{Data, EncryptedSecret, Private, SymmetricDefinitionObject},
+    tss2_esys::*,
+    Error, Result,
+};
+use log::error;
+
+use std::convert::TryFrom;
+use std::ptr::null_mut;
 
 impl Context {
-    // Missing function: Duplicate
+    /// Duplicate a loaded object so that it may be used in a different hierarchy.
+    ///
+    /// # Details
+    /// This command duplicates a loaded object so that it may be used in a different hierarchy.
+    /// The new parent key for the duplicate may be on the same or different TPM or the Null hierarchy.
+    /// Only the public area of `new_parent_handle` is required to be loaded.
+    ///
+    /// # Arguments
+    /// * `object_to_duplicate` - An [ObjectHandle] of the object that will be duplicated.
+    /// * `new_parent_handle` - An [ObjectHandle] of the new parent.
+    /// * `encryption_key_in` - An optional encryption key.
+    /// * `symmetric_alg` - Symmetric algorithm to be used for the inner wrapper.
+    ///
+    /// The `object_to_duplicate` need to be have Fixed TPM and Fixed Parent attributes set to `false`.
+    ///
+    /// # Returns
+    /// The command returns a tuple consisting of:
+    /// * `encryption_key_out` - TPM generated, symmetric encryption key for the inner wrapper if
+    ///   `symmetric_alg` is not `Null`.
+    /// * `duplicate` - Private area that may be encrypted.
+    /// * `out_sym_seed` - Seed protected by the asymmetric algorithms of new parent.
+    ///
+    /// ```rust
+    /// # use std::convert::{TryFrom, TryInto};
+    /// # use tss_esapi::attributes::{ObjectAttributesBuilder, SessionAttributesBuilder};
+    /// # use tss_esapi::constants::{tss::TPM2_CC_Duplicate, SessionType};
+    /// # use tss_esapi::handles::ObjectHandle;
+    /// # use tss_esapi::interface_types::{
+    /// #     algorithm::{HashingAlgorithm, PublicAlgorithm},
+    /// #     key_bits::RsaKeyBits,
+    /// #     resource_handles::Hierarchy,
+    /// #     session_handles::PolicySession,
+    /// # };
+    /// # use tss_esapi::structures::SymmetricDefinition;
+    /// # use tss_esapi::structures::{
+    /// #     PublicBuilder, PublicKeyRsa, PublicRsaParametersBuilder, RsaScheme,
+    /// #     RsaExponent,
+    /// # };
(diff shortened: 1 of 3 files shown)


tss-esapi/tests/context_tests/tpm_commands/duplication_commands_tests.rs
```diff
@@ -1,2 +1,57 @@
...
+        context.tr_set_auth(new_parent_handle, &key_auth).unwrap();
+
+        let _ = context
+            .duplicate(
+                object_to_duplicate_handle,
+                new_parent_handle,
+                None,
+                SymmetricDefinitionObject::Null,
+            )
+            .unwrap();
```
ionut-arm, line 55: I think you need to crate a policy session for this call and follow the steps mentioned in the spec for this command to set it up. Unfortunately, it's probably going to be somewhat trial-and-error.

tss-esapi/src/context/tpm_commands/duplication_commands.rs
```diff
@@ -1,9 +1,57 @@
...
+    ) -> Result<(Data, Private, EncryptedSecret)> {
+        let mut encryption_key_out = null_mut();
+        let mut duplicate = null_mut();
+        let mut out_sym_seed = null_mut();
+        let ret = unsafe {
+            Esys_Duplicate(
+                self.mut_context(),
+                object_to_duplicate.into(),
+                new_parent_handle.into(),
+                self.optional_session_1(),
```
ionut-arm, line 32: This session won't be optional, so `self.required_session_1()`

tss-esapi/tests/context_tests/tpm_commands/duplication_commands_tests.rs
wiktor-k: Okay, thanks for the hint! I already had a policy session that allowed duplication but then key creation failed. I'll check it out, thanks!

ionut-arm: I think that session should only be used for the duplicate call, not for creation as well!

tss-esapi/tests/context_tests/tpm_commands/duplication_commands_tests.rs
```diff
@@ -1,2 +1,131 @@
...
+    };
+
+    #[test]
+    fn test_duplicate() {
+        let mut context = create_ctx_with_session();
+
+        let random_digest = context.get_random(16).unwrap();
+        let key_auth = Auth::try_from(random_digest.value().to_vec()).unwrap();
+
+        let object_to_duplicate_handle: ObjectHandle = context
```
Superhepper, line 27: I do not think you can duplicate a Primary key. Because duplicate is used to move an object from one Hierarchy to another (potentially on another machine)

wiktor-k:
Okay, phew. Finally got it working by basically memorizing the integration test from here: https://github.com/utkarsh2102/tpm2-tss/blob/5eca0a4ae129f410303de1c20046ebe7527b9ad3/test/integration/esys-duplicate.int.c#L382 and rewriting it in Rust and our API.

This is ready for review now. One thing I considered is removing the separate `duplication_tests` since the doctest is basically testing the same... Opinions welcome!

tss-esapi/src/context/tpm_commands/duplication_commands.rs
```diff
@@ -1,9 +1,334 @@
...
+    ///
+    /// # Returns
+    /// The command returns a tuple consisting of:
+    /// * `encryption_key_out` - TPM generated, symmetric encryption key for the inner wrapper if
+    ///   `symmetric_alg` is not `Null`.
+    /// * `duplicate` - Private area that may be encrypted.
+    /// * `out_sym_seed` - Seed protected by the asymmetric algorithms of new parent.
+    ///
+    /// ```rust
+    /// use std::convert::{TryFrom, TryInto};
```
Superhepper, line 39: Really nice example!
It is nice that it runs and works as test. Though this will show up as an example in the documentation and there it will probably be way to long. So what we have done previously is to hide everything that does not directly involve calling the api that you are documenting. So that we only show the important parts but still the documentation will fail to compile if we change something (if you need some inspiration look at evict_control, that have like 300 lines of code but only about 20 is visible).

tss-esapi/tests/context_tests/tpm_commands/duplication_commands_tests.rs
```diff
@@ -1,2 +1,193 @@
...
+
+        let public_child = PublicBuilder::new()
+            .with_public_algorithm(PublicAlgorithm::Rsa)
+            .with_name_hashing_algorithm(HashingAlgorithm::Sha256)
+            .with_object_attributes(object_attributes)
+            .with_auth_policy(&digest)
+            .with_rsa_parameters(
+                PublicRsaParametersBuilder::new()
+                    .with_scheme(RsaScheme::Null)
+                    .with_key_bits(RsaKeyBits::Rsa2048)
```
Superhepper, line 101: If you want your test to be different from your documentation then one way of doing it is to use another type of key like a ECC key.

wiktor-k: Excellent idea! I'll rework the test case to use ECC keys.

tss-esapi/src/context/tpm_commands/duplication_commands.rs
wiktor-k: Understood. I was thinking about reducing the visible lines but there are quite some things that need to be set (like fixedTPM and fixedParent = false) for the example to work... I'll trim it down.

wiktor-k:
Okay, improved the PR in a couple of ways: as suggested by Jesper: minimized the visible example code and changed the test case to use ECC. The latter proved to be more daunting than I anticipated as it was not clear which parameter combinations were invalid so I added more error reporting (the change is minimal so I didn't open a new PR). As usual: I'm open to suggestions w.r.t. error messages.

Thanks for your time! :bow:

Superhepper (approved):

ionut-arm (approved):
👍🏻 LGTM

wiktor-k:
Greatly appreciated @ionut-arm ! :bow:

PR #253, opened 2021-08-27
Add missing key handle session checks

Superhepper:
- Added checks for a key handle session to the context
  methods that according to the specification requires
  it.

### DIFF
--- a/tss-esapi/src/context/tpm_commands/asymmetric_primitives.rs
+++ b/tss-esapi/src/context/tpm_commands/asymmetric_primitives.rs
@@ -55,7 +55,7 @@ impl Context {
             Esys_RSA_Decrypt(
                 self.mut_context(),
                 key_handle.into(),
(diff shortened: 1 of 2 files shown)


ionut-arm (approved):
Hmmm, I thought there are other methods that would need a change like this, but I'll try and hunt them down in a different PR

Superhepper:
There might be. This is for the ones I found in the esapi specification that explicitly mentioned '''keyHandkeSession'''

ionut-arm:
> This is a breaking change.

Tbh I'd rather see this as a bug fix, even if it changes the "operation" of those methods.

Superhepper:
Yeah, actually if you do not provide a session the error will come from the ESYS layer. We have just moved the error forward to the rust wrapper instead. So I am not sure I would define it as breaking if not some one uses the wrapper in a rather strange way.

PR #255, opened 2021-09-08
Add ECDH-related functions

wiktor-k:
This PR adds ECDH KeyGen and ZGen functions.

I left the naming similar to their TPM counterparts but if you have better options I'm all ears.

I don't mind if this PR is merged after https://github.com/parallaxsecond/rust-tss-esapi/pull/241 I'll then rebase and fix the raw types to their wrapper types. (OTOH adjusting shouldn't be difficult).

I'm submitting now so that this is tracked as I need the code anyway for my work :)

Have a nice day! :wave:

### DIFF
--- a/tss-esapi/src/context/tpm_commands/asymmetric_primitives.rs
+++ b/tss-esapi/src/context/tpm_commands/asymmetric_primitives.rs
@@ -321,11 +321,6 @@ impl Context {
     /// ```
     pub fn ecdh_z_gen(&mut self, key_handle: KeyHandle, in_point: EccPoint) -> Result<EccPoint> {
         let mut point = null_mut();
-        let in_point: TPMS_ECC_POINT = in_point.into();
-        let in_point = TPM2B_ECC_POINT {
-            size: in_point.x.size + in_point.y.size,
-            point: in_point,
-        };
         let ret = unsafe {
             Esys_ECDH_ZGen(
                 self.mut_context(),
@@ -333,7 +328,7 @@ impl Context {
                 self.required_session_1()?,
                 self.optional_session_2(),
                 self.optional_session_3(),
-                &in_point,
+                &in_point.into(),
                 &mut point,
             )
         };
--- a/tss-esapi/src/context/tpm_commands/asymmetric_primitives.rs
+++ b/tss-esapi/src/context/tpm_commands/asymmetric_primitives.rs
@@ -3,12 +3,13 @@
 use crate::{
     handles::KeyHandle,
     structures::Data,
-    structures::{PublicKeyRsa, RsaDecryptionScheme},
+    structures::{EccPoint, PublicKeyRsa, RsaDecryptionScheme},
     tss2_esys::*,
     Context, Error, Result,
 };
 use log::error;
 use std::convert::TryFrom;
+use std::convert::TryInto;
 use std::ptr::null_mut;
 
 impl Context {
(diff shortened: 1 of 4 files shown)


tss-esapi/src/context/tpm_commands/asymmetric_primitives.rs
```diff
@@ -75,8 +75,63 @@ impl Context {
         }
     }
 
-    // Missing function: ECDH_KeyGen
-    // Missing function: ECDH_ZGen
+    /// Generate an ephemeral key pair.
```
Superhepper, line 78: It would be nice if you could provide an example in the documentation of the added context method.

tss-esapi/src/context/tpm_commands/asymmetric_primitives.rs
```diff
@@ -75,8 +75,63 @@ impl Context {
...
+    pub fn ecdh_keygen(
```
Superhepper, line 79: I think I would prefer it to be ```ecdh_key_gen``` that would be more in line with how the other methods are named.

tss-esapi/src/context/tpm_commands/asymmetric_primitives.rs
```diff
@@ -75,8 +75,63 @@ impl Context {
...
+
+        if ret.is_success() {
+            Ok(unsafe { (*z_point, *pub_point) })
+        } else {
+            error!("Error when generating ECDH keypair: {}", ret);
+            Err(ret)
+        }
+    }
+
+    /// Recover Z value from a public point and a private key.
```
Superhepper, line 107: Same as above in my opinion context methods needs more documentation.

tss-esapi/src/context/tpm_commands/asymmetric_primitives.rs
```diff
@@ -75,8 +75,63 @@ impl Context {
...
+    pub fn ecdh_zgen(
```
Superhepper, line 108: I think I would prefer this to be name ```ecdh_z_gen```.

tss-esapi/src/context/tpm_commands/asymmetric_primitives.rs
```diff
@@ -75,8 +75,63 @@ impl Context {
...
+        &mut self,
+        key_handle: KeyHandle,
+    ) -> Result<(TPM2B_ECC_POINT, TPM2B_ECC_POINT)> {
+        let mut z_point = null_mut();
+        let mut pub_point = null_mut();
+        let ret = unsafe {
+            Esys_ECDH_KeyGen(
+                self.mut_context(),
+                key_handle.into(),
+                self.required_session_1()?,
```
Superhepper, line 89: Is this session really required?

Superhepper (changes requested):
Thank you for the changes!
I had a few comments.

tss-esapi/src/context/tpm_commands/asymmetric_primitives.rs
wiktor-k: Yes, definitely! I've added more comprehensive doc tests for key gen and z-gen.

tss-esapi/src/context/tpm_commands/asymmetric_primitives.rs
wiktor-k: Understood! :+1:

tss-esapi/src/context/tpm_commands/asymmetric_primitives.rs
wiktor-k: I'm not entirely sure how to read the TPM docs on this matter. I'll mark the session as optional for now.

wiktor-k:
@Superhepper thanks a lot for your detailed review! I've addressed the points that you found. Hopefully the PR looks better now :)

Superhepper (approved):
Nice

tss-esapi/src/context/tpm_commands/asymmetric_primitives.rs
ionut-arm: Just for completeness: no sessions are required for this op. You can tell by looking at the Command and Response diagram for `TPM2_ECDH_KeyGen` in the [Commands spec](https://trustedcomputinggroup.org/wp-content/uploads/TCG_TPM2_r1p59_Part3_Commands_pub.pdf). Quoting from the architecture part of the spec:

> The number of authorization sessions that a command will have is indicated in the command schematic in 
TPM 2.0  Part  3. If a handle in the  handle area has the "@" decoration, then an authorization session is 
required  be  present  (an  authorization  session  being  either  a  password,  a  policy  session,  or  an  HMAC 
session). 
The  authorization  sessions  occur  in  the  order  of  the  associated  entity  handles.  That  is,  the  first  handle 
with an "@" decoration will be associated with the first session in the Authorization Area. 

If you go to the same section for TPM2_ECDH_ZGen, for example, you can see that the key handle has a `@` in front, so a session is required there.

wiktor-k: Great, thanks for the explanation @ionut-arm! I believe now the session management should be adjusted in both cases.

tss-esapi/src/context/tpm_commands/asymmetric_primitives.rs
```diff
@@ -78,8 +79,274 @@ impl Context {
...
+    /// // Generate ephemeral key pair and a shared secret
+    /// let (z_point, pub_point) = context.ecdh_key_gen(key_handle).unwrap();
+    /// let z_point_gen = context.ecdh_z_gen(key_handle, pub_point).unwrap();
+    /// assert_eq!(z_point.x().value(), z_point_gen.x().value());
+    /// ```
+    pub fn ecdh_z_gen(&mut self, key_handle: KeyHandle, in_point: EccPoint) -> Result<EccPoint> {
+        let mut point = null_mut();
+        let in_point: TPMS_ECC_POINT = in_point.into();
+        let in_point = TPM2B_ECC_POINT {
+            size: in_point.x.size + in_point.y.size,
```
wiktor-k, line 326: @Superhepper @ionut-arm folks, could you take a look if this is a correct way to convert `TPMS_ECC_POINT` into `TPM2B_ECC_POINT`? Thanks!

wiktor-k:
Okay, phew, this should be ready for re-review now. I need to add clippy check format to git pre-commit hooks :)

tss-esapi/src/context/tpm_commands/asymmetric_primitives.rs
ionut-arm: I'm not sure I follow exactly what they mean in the spec by
> if  the  point  is  to  be  omitted,  the  X  and  Y  coordinates  need  to  be  individually  set  to  Empty  Buffers.  The 
minimum value for size will be four. It is checked indirectly by unmarshaling of the TPMS_ECC_POINT. If 
the type of point were BYTE, then size could have been zero. However, this would complicate the process 
of marshaling the structure. 

As in, I'm unsure what you'd put in `TPMS_ECC_POINT` if you wanted to have the `size` in `TPM2B_ECC_POINT` as 4 - do you just set everything to 0 (as in, just `Default::default`)?
The "normal" value for `size` should be `::std::mem::size_of::<TPMS_ECC_POINT>()`. I think your calculation only covers the sizes of the buffers, but not of the whole structure (since they each contain a `size` field as well)

Superhepper: hrrm... I am not sure. I will look into it. I think it would be better if the conversion into TPM2B_ECC_POINT is added to EccPoint struct.

wiktor-k: > I think it would be better if the conversion into TPM2B_ECC_POINT is added to EccPoint struct.

Yes, agreed!

> hrrm... I am not sure. I will look into it.

Greatly appreciated :bow:

Superhepper: As long as you do not want the parameter to be encrypted then it is fine to use size = 0 but that is not something we can do. We need to support encryption of parameters so the calculation of the size should be:

```rust
size = std::mem::size_of::<u16>() + point.x.len() + std::mem::size_of::<u16>() + point.y.len()
```

I think ....

ionut-arm: > As long as you do not want the parameter to be encrypted then it is fine to use size = 0 

I think that's what the spec tries to explain away in what I commented above. Have you tried using it with either @Superhepper's equation or `::std::mem::size_of::<TPMS_ECC_POINT>()`?

Superhepper: @ionut-arm I do not think ```::std::mem::size_of::<TPMS_ECC_POINT>()``` because that will suggest that the entire allocated space contains valid data. And in the case where a point is to be omitted the x and y needs to be empty buffers but still the marshalling needs to work so the size cannot be 0 it needs to be four so that the size parameters of the TPMS_ECC_POINT can be read and confirmed that they are empty buffers.

ionut-arm: Ok, but if you look at what the description of the `size` field says: `size of the remainder of this structure`. And the remainder of the structure is a `TPMS_ECC_POINT`, which at least in Rust _cannot_ fit in 4 bytes. I assume the TPM just takes the `size: 4` as an "edge" case and knows to simply assume it's empty. So maybe in the `EccPoint` => `TPM2B_ECC_POINT` conversion we can have an edge case of our own if both `x` and `y` are empty and set it to 4, otherwise just use the size of `TPMS_ECC_POINT`?

Superhepper: Yes, wording in the specification made me seriously confused because I didn't understand what they meant with ```remainder``` but I interpret it as data left in the structure that is valid. So when came up with the equation so I thought that the size field in the ```TPM2B_ECC_PARAMETER``` is the only thing that always contains valid data i.e. ```std::mem::size_of::<u16>()``` and there is one such field for x and one for y so the minimum number of valid bytes must be four. So the equation should work for the empty buffer case as well.

like this:
```rust
TPM2B_ECC_POINT {
    size: u16,
    point: TPM2_ECC_POINT {
        x: TPM2B_ECC_PARAMETER {
            size: u16 // 2 Bytes valid data always
            buffer: [u8; MAX_ECC_KEY_BYTES] // Contains 0 to MAX_ECC_KEY_BYTES valid bytes
        }
        y: TPM2B_ECC_PARAMETER {
            size: u16 // 2 Bytes valid data always
            buffer: [u8; MAX_ECC_KEY_BYTES] // Contains 0 to MAX_ECC_KEY_BYTES valid bytes
        }        
    },
}
```

wiktor-k: Thanks for the elaborate explanation of what's happening with the `size` field folks! I've adjusted the code moving conversion to `TPM2B_ECC_POINT` into a proper `From` impl and reused that in the ECDH function using @Superhepper's equation.

I think this is ready for a next round of reviews. Thanks in advance!

ionut-arm (approved):
Looks good to me! 👍🏻

tss-esapi/src/context/tpm_commands/asymmetric_primitives.rs
ionut-arm: @Superhepper - do you want to have another look at the `From` implementation?

ionut-arm:
Ok, I'm merging this and releasing a new alpha version of the crate (as I'm a bit pressured for the Parsec release) - if any other issues appear we can fix them in an upcoming release.

PR #256, opened 2021-09-13
Move TransientKeyContext away from contexts

ionut-arm:
This commit moves the TransientKeyContext away from using saved key
contexts. Instead, keys are returned and managed as pairs of public and
(encrypted) private parts which are loaded into the TPM every time an
operation is requested.
A migration mechanism is also implemented to convert from contexts to
the new key material format.

### DIFF
--- a/tss-esapi/src/abstraction/transient.rs
+++ b/tss-esapi/src/abstraction/transient.rs
@@ -14,9 +14,8 @@
 //! Object contexts thus act as an opaque handle that can, however, be used by the client to seralize
 //! and persist the underlying data.
 use crate::{
-    abstraction::cipher::Cipher,
     attributes::{ObjectAttributesBuilder, SessionAttributesBuilder},
-    constants::{tss::*, SessionType},
+    constants::{tss::*, SessionType, Tss2ResponseCodeKind},
     handles::{KeyHandle, SessionHandle},
     interface_types::{
         algorithm::{HashingAlgorithm, PublicAlgorithm, RsaSchemeAlgorithm},
@@ -25,23 +24,73 @@ use crate::{
         resource_handles::Hierarchy,
     },
     structures::{
-        Auth, CreateKeyResult, Data, Digest, EccScheme, Public, PublicBuilder, PublicKeyRsa,
-        PublicRsaParametersBuilder, RsaDecryptionScheme, RsaExponent, RsaScheme, Signature,
-        VerifiedTicket,
+        Auth, CreateKeyResult, Data, Digest, EccPoint, EccScheme, Public, PublicBuilder,
+        PublicEccParametersBuilder, PublicKeyRsa, PublicRsaParametersBuilder, RsaExponent,
+        RsaScheme, Signature, SymmetricDefinitionObject, VerifiedTicket,
     },
     tcti_ldr::TctiNameConf,
     tss2_esys::*,
     utils::{
         create_restricted_decryption_rsa_public,
-        create_unrestricted_encryption_decryption_rsa_public,
-        create_unrestricted_signing_ecc_public, create_unrestricted_signing_rsa_public,
         create_unrestricted_signing_rsa_public_with_unique, PublicKey, TpmsContext,
     },
     Context, Error, Result, WrapperErrorKind as ErrorKind,
 };
 
 use log::error;
+use serde::{Deserialize, Serialize};
 use std::convert::{TryFrom, TryInto};
+use zeroize::Zeroize;
+
+/// Parameters for the kinds of keys supported by the context
+#[derive(Debug, Clone, Copy)]
+pub enum KeyParams {
+    Rsa {
+        /// Size of key in bits
+        ///
+        /// Can only be one of: 1024, 2048, 3072 or 4096
+        size: RsaKeyBits,
+        /// Asymmetric scheme to be used with the key
+        scheme: RsaScheme,
+        /// Public exponent of the key
+        ///
+        /// If set to 0, it will default to 2^16 - 1
+        pub_exponent: RsaExponent,
+    },
+    Ecc {
+        /// Curve that the key will be based on
+        curve: EccCurve,
+        /// Asymmetric scheme to be used with the key
+        scheme: EccScheme,
+    },
+}
+
+/// Structure representing a key created or stored in the TPM
+///
+/// The `public` field represents the public part of the key in plain text,
+/// while `private` is the encrypted version of the private key.
+///
+/// # Warning
+///
+/// If the Owner hierarchy is cleared, any key material generated
+/// prior to that event will become unusable.
+#[derive(Debug, Serialize, Deserialize, Clone, Zeroize)]
+pub struct KeyMaterial {
+    public: PublicKey,
+    private: Vec<u8>,
+}
+
+impl KeyMaterial {
+    /// Get a reference to the public part of the key
+    pub fn public(&self) -> &PublicKey {
+        &self.public
+    }
+
+    /// Get a reference to the private part of the key
+    pub fn private(&self) -> &[u8] {
+        &self.private
+    }
+}
 
 /// Structure offering an abstracted programming experience.
 ///
@@ -65,7 +114,7 @@ impl TransientKeyContext {
     ///
     /// A key is created as a descendant of the context root key, with the given parameters.
     ///
-    /// If successful, the result contains the saved context of the key and a vector of
+    /// If successful, the result contains the [KeyMaterial] of the key and a vector of
     /// bytes forming the authentication value for said key.
     ///
     /// # Constraints
@@ -73,18 +122,11 @@ impl TransientKeyContext {
     ///
     /// # Errors
     /// * if the authentication size is larger than 32 a `WrongParamSize` wrapper error is returned
-    /// * for RSA keys, if the specified key size is not one of 1024, 2048, 3072 or 4096, `WrongParamSize`
-    /// is returned
-    /// * if the key_params is not for an RSA key, `InvalidParam` is returned
-    /// * if the key_params does not have an `AnySig` scheme, `InvalidParam` is returned
-    /// * errors are returned if any method calls return an error: `Context::get_random`,
-    /// `TransientKeyContext::set_session_attrs`, `Context::create`, `Context::load`,
-    /// `Context::context_save`, `Context::context_flush`
     pub fn create_key(
         &mut self,
         key_params: KeyParams,
         auth_size: usize,
-    ) -> Result<(TpmsContext, Option<Auth>)> {
+    ) -> Result<(KeyMaterial, Option<Auth>)> {
         if auth_size > 32 {
             return Err(Error::local_error(ErrorKind::WrongParamSize));
         }
@@ -103,44 +145,18 @@ impl TransientKeyContext {
             ..
         } = self.context.create(
             self.root_key_handle,
-            &self.get_public_from_params(key_params)?,
+            &TransientKeyContext::get_public_from_params(key_params, None)?,
             key_auth.as_ref(),
             None,
             None,
             None,
         )?;
-        self.set_session_attrs()?;
-        let key_handle = self
-            .context
-            .load(self.root_key_handle, out_private, &out_public)?;
 
-        self.set_session_attrs()?;
-        let key_context = self.context.context_save(key_handle.into()).or_else(|e| {
-            self.context.flush_context(key_handle.into())?;
-            Err(e)
-        })?;
-        self.context.flush_context(key_handle.into())?;
-        Ok((key_context, key_auth))
-    }
-
-    fn get_public_from_params(&self, params: KeyParams) -> Result<Public> {
-        match params {
-            KeyParams::RsaSign {
-                size,
-                scheme,
-                pub_exponent,
-            } => Ok(create_unrestricted_signing_rsa_public(
-                scheme,
-                size,
-                pub_exponent,
-            )?),
-            KeyParams::RsaEncrypt { size, pub_exponent } => {
-                create_unrestricted_encryption_decryption_rsa_public(size, pub_exponent)
-            }
-            KeyParams::Ecc { curve, scheme } => {
-                Ok(create_unrestricted_signing_ecc_public(scheme, curve)?)
-            }
-        }
+        let key_material = KeyMaterial {
+            public: out_public.try_into()?,
+            private: out_private.value().to_vec(),
+        };
+        Ok((key_material, key_auth))
     }
 
     /// Load a previously generated RSA public key.
@@ -152,10 +168,7 @@ impl TransientKeyContext {
     ///
     /// # Errors
     /// * if the public key length is different than 128, 256, 384 or 512 bytes, a `WrongParamSize` wrapper error is returned
-    /// * errors are returned if any method calls return an error:
-    /// `TransientKeyContext::`set_session_attrs`, `Context::load_external_public`,
-    /// `Context::context_save`, `Context::flush_context`
-    pub fn load_external_rsa_public_key(&mut self, public_key: &[u8]) -> Result<TpmsContext> {
+    pub fn load_external_rsa_public_key(&mut self, public_key: &[u8]) -> Result<KeyMaterial> {
         let rsa_key_bits = RsaKeyBits::try_from(
             u16::try_from(public_key.len()).map_err(|e| {
                 error!("Failed to convert length of public key to u16: {:?}", e);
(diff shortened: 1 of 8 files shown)


tss-esapi/src/context/tpm_commands/duplication_commands.rs
```diff
@@ -1,9 +1,57 @@
...
+};
+use log::error;
+use mbox::MBox;
+use std::convert::TryFrom;
+use std::ptr::null_mut;
 
 impl Context {
-    // Missing function: Duplicate
+    /// Change authorization for a TPM-resident object.
+    pub fn duplicate(
```
ionut-arm, line 16: Note: this function isn't used anywhere, I added it because I thought I might use it but ended up relying on `ObjectChangeAuth` instead.

tss-esapi/src/structures/buffers.rs
```diff
@@ -106,7 +106,7 @@ pub mod sensitive_data {
 }
 
 pub mod private {
-    buffer_type!(Private, 256, TPM2B_PRIVATE);
+    buffer_type!(Private, 1550, TPM2B_PRIVATE);
```
paulhowardarm, line 109: Where does 1550 derive from? Should this be defined as a named constant with some description? To an outside observer like me, this looks like an arbitrary number and I can't see why the number has changed when the structure hasn't.

tss-esapi/src/context/tpm_commands/duplication_commands.rs
```diff
@@ -1,9 +1,57 @@
...
+    Context, Error, Result,
+};
+use log::error;
+use mbox::MBox;
+use std::convert::TryFrom;
+use std::ptr::null_mut;
 
 impl Context {
-    // Missing function: Duplicate
+    /// Change authorization for a TPM-resident object.
```
paulhowardarm, line 15: The comment is wrong - seems to have been lifted from `object_change_auth()`.

tss-esapi/src/utils/mod.rs
```diff
@@ -124,7 +131,7 @@ pub fn create_restricted_decryption_rsa_public(
             )
             .build()?,
         )
-        .with_rsa_unique_identifier(&PublicKeyRsa::new_empty_with_size(rsa_key_bits))
+        .with_rsa_unique_identifier(&PublicKeyRsa::default())
```
paulhowardarm, line 134: Is this change a necessary part of the fix, or just some tidying up? I assume it's intentional that we no longer take account of the key size in this line?

tss-esapi/tests/transient_key_context.rs
```diff
@@ -92,32 +91,6 @@ fn load_bad_sized_key() {
     );
 }
 
-#[test]
```
paulhowardarm, line 95: What's the reason for removing some tests in this file?

tss-esapi/src/abstraction/transient.rs
```diff
@@ -172,173 +192,41 @@ impl TransientKeyContext {
...
-        }
-        self.set_session_attrs()?;
+        let key_handle = self.load_key(key_params, key_material, key_auth)?;
+        let decrypt_scheme = if let KeyParams::Rsa { scheme, .. } = key_params {
+            scheme.try_into()?
+        } else {
+            return Err(Error::local_error(ErrorKind::InvalidParam));
+        };
 
+        self.set_session_attrs()?;
```
paulhowardarm, line 221: Just for my education, when and why does `self.set_session_attrs()` need to be called? The reason I ask is that this change seems to have deleted a lot of the calls, but retained a few. For example, we no longer call it immediately before `self.load_key()`, but we _do_ still call it before `self.context.rsa_encrypt()` for example. Do we only need to call it before using `self.context`, and otherwise not?

tss-esapi/src/structures/buffers.rs
ionut-arm: That's true, I should probably use `::std::mem::size_of::<_PRIVATE>()`. The reason is that the spec isn't exactly clear about that size, it just says it must be at most essentially `::std::mem::size_of::<_PRIVATE>()`, with some definition given for `_PRIVATE`. Will change.

tss-esapi/tests/transient_key_context.rs
ionut-arm: The tests highlighted here were using an operation that I've removed (loading external keys). I guess it's a bit inconsistent that I've only commented out the one below, did that to not lose the actual key values which are a pain to get (though I guess I could just come back to an old commit and copy them from there). I'll remove that one too.

tss-esapi/src/utils/mod.rs
ionut-arm: I'm doubting myself now, let me check again if this had an effect on the primary key.

The short answer is that this has no effect - the `unique` field is replaced by the TPM during the operation anyway, so it shouldn't matter what we put there. As long as the content is in line with what the type of the object is, it should be ok. The old approach (obviously) put a bunch of zero bytes in there and set the length to that of the RSA key, the new one just sets it to an empty buffer. I changed it because I thought it might be influencing the seeding of the key. Let me check again.

tss-esapi/src/abstraction/transient.rs
ionut-arm: `self.set_session_attrs()` is called to enable some flags on the sessions that we use for almost all TPM operations (so yes, before calling methods on `self.context`). We use them to enable parameter encryption over the wire between Parsec and the TPM.

tss-esapi/src/utils/mod.rs
ionut-arm: Yeah, so it seems `ObjectChangeAuth` fails because it doesn't recognise the primary key as the "true" parent of the keys loaded from contexts. That means the "name" of the primary key is computed over the `TPM2B_PUBLIC` received from the user (or, in any case, includes this `unique` field), and since before we were simply leaving the field empty, adding these empty bytes would prevent us from migrating in this way.

paulhowardarm (approved):
It looks good to me now that a couple of my comments are resolved - I don't have much TPM/ESAPI expertise, though.

tss-esapi/src/context/tpm_commands/duplication_commands.rs
wiktor-k: If you don't mind I'd suggest removing it then.

I will add it as part of the https://github.com/parallaxsecond/rust-tss-esapi/pull/248 PR including tests. Sorry that it takes some time but I've got a work schedule but duplicate is now my next work item and I also found integration test in C that may help me get the test in shape: https://github.com/utkarsh2102/tpm2-tss/blob/5eca0a4ae129f410303de1c20046ebe7527b9ad3/test/integration/esys-duplicate.int.c#L382

:wave:

tss-esapi/src/abstraction/transient.rs
```diff
@@ -494,6 +421,117 @@ impl TransientKeyContext {
...
+    fn load_key(
+        &mut self,
+        params: KeyParams,
+        material: KeyMaterial,
+        auth: Option<Auth>,
+    ) -> Result<KeyHandle> {
+        let public = TransientKeyContext::get_public_from_params(params, Some(material.public))?;
+
+        self.set_session_attrs()?;
+        let key_handle = match material.private.is_empty() {
```
wiktor-k, line 515: Why not make it a good ol' `if`?

tss-esapi/src/abstraction/transient.rs
```diff
@@ -645,9 +686,9 @@ impl TransientKeyContextBuilder {
             .create_primary(
                 self.hierarchy,
                 &create_restricted_decryption_rsa_public(
-                    self.default_context_cipher.try_into()?,
+                    self.default_context_cipher,
                     root_key_rsa_key_bits,
-                    RsaExponent::default(),
+                    RsaExponent::create(0).unwrap(),
```
wiktor-k, line 691: That's a bit unfortunate change. I guess this is related to the fact that RsaExponent's Default implementation now uses `value: (1 << 16) + 1` and not `0`?

[The spec](https://trustedcomputinggroup.org/wp-content/uploads/TCG_TPM2_r1p59_Part2_Structures_pub.pdf) in paragraph 12.2.3.5 (TPMS_RSA_PARMS) says that:

> An exponent of zero indicates that the exponent is the default of 2^16 + 1.

It's just... changing the `default()` to `RsaExponent::create(0).unwrap()` looks like a small code smell... maybe that's just me :)

tss-esapi/tests/transient_key_context.rs
```diff
@@ -499,27 +432,110 @@ fn create_ecc_key_decryption_scheme() {
...
+                Digest::try_from(HASH.to_vec()).unwrap(),
+                signature,
+            )
             .unwrap();
     }
 }
+
+#[test]
+fn ctx_migration_test() {
+    // Create two key contexts with a
```
wiktor-k, line 467: It seems the sentence was cut here.

tss-esapi/tests/transient_key_context.rs
```diff
@@ -499,27 +432,110 @@ fn create_ecc_key_decryption_scheme() {
...
+        )
+        .unwrap();
+    let _ = ctx
+        .verify_signature(
+            pub_key,
+            key_params,
+            Digest::try_from(HASH.to_vec()).unwrap(),
+            signature,
+        )
+        .expect("the signature should be valid");
```
wiktor-k, line 540: Not sure if that's a good idea but maybe comparing the public keys before migration and after migration and making sure they're equal would be a good idea?

Currently from my reading it seems that the migration tests if the key from migration is usable but it might have been any working key that `migrate_key_from_ctx` can return. Or sign using migrated key but verify signature using non-migrated one... Maybe I'm overthinking this :thinking:

wiktor-k:
LGTM with a couple of minor questions :+1:

tss-esapi/src/context/tpm_commands/duplication_commands.rs
ionut-arm: I'll remove it, then.

tss-esapi/src/abstraction/transient.rs
ionut-arm: Yes, similar to Paul's comment above about `PublicKeyRsa::new_empty_with_size(rsa_key_bits)` => `PublicKeyRsa::default()`, just in the opposite direction. If we use the new one the attributes of the key aren't kept the same and the primary key names becomes something else.

tss-esapi/src/context/tpm_commands/duplication_commands.rs
wiktor-k: Thank you :bow:

Now I remembered that Jesper also brought interesting point that new context functions could use some more improved documentation (including working examples): https://github.com/parallaxsecond/rust-tss-esapi/pull/255#discussion_r704571544 so this would also be problematic here. I'll keep that in mind in my PR.

tss-esapi/src/abstraction/transient.rs
wiktor-k: :thinking: I think in this case 0 is exactly the same as 2^16 + 1 (== `(1 << 16) + 1`) by the TPM spec. It's just `create(0).unwrap()` triggers a red-flag in my mind as 0 is defined by the spec to be a good default and unwraps... are just suspicious :) But this is just a minor thing. I don't want to split hair on something this small :)

tss-esapi/tests/transient_key_context.rs
ionut-arm: I'll add a check. There's a couple of reasons why we don't really do this: almost every operation is integrity protected in some way, e.g. for `TPM2_Load`:
> After  inPrivate.buffer  is  decrypted  using  the  symmetric  key  of  the  parent,  the  integrity  value  shall  be 
checked before the sensitive area is used, or unmarshaled
> ...
> Before  use,  a  loaded  object  shall  be  checked  to  validate  that  the  public  and  sensitive  portions  are 
properly linked, cryptographically.

Second reason is that this tests the TPM underneath more than it tests our crate. But I think we're note very consistent on this one

tss-esapi/src/abstraction/transient.rs
ionut-arm: It's not exactly the same only because 0 is not the same number as 2^16+1, even if they're treated the same by the TPM. The reason it's a problem is that when the cryptographic hash is computed over the `TPM2B_Public` structure the fact that it's some number instead of 0 completely changes the result. 

Yeah, we tend to avoid `unwrap`, but it's accepted in cases where we know it isn't going to fail. Tbh we could have a test somewhere making sure that `RsaExponent::create(0).unwrap` doesn't panic, since 0 is a fairly standard value. I'm debating changing the `RsaExponent::default` to 0, though, thoughts? @Superhepper

tss-esapi/src/abstraction/transient.rs
ionut-arm: hmm, true, I think I started with a different approach and switched to this without thinking of `if`.

tss-esapi/src/context/tpm_commands/duplication_commands.rs
ionut-arm: > new context functions could use some more improved documentation

Yeah, I was thinking of that... 😬

tss-esapi/src/abstraction/transient.rs
Superhepper: If the default value used by RsaExponent is wrong why not change it? Or add ``` const ZERO_EXPONENT: RsaExponent = RsaExponent { value: 0}``` to the RsaExponent struct.

Superhepper (approved):
It looks good to me (with 600x400 resolution).

One thing to consider is if it would be a good idea to create a ```tss-esapi/src/abstraction/transient/migration.rs``` where we can put functions we need to change in order to support migration but might not want to change otherwise?

tss-esapi/src/abstraction/transient.rs
ionut-arm: It's not really wrong per se, `(1<<16) + 1` is equally correct. I'll add the `const`, was mulling that idea over.

ionut-arm: The architecture spec does state that the default should be 2^16+1 in a few places, so I'll leave it as is, I think 0 is just the lazy/shorthand way of doing the same thing.

ionut-arm:
> One thing to consider is if it would be a good idea to create a tss-esapi/src/abstraction/transient/migration.rs where we can put functions we need to change in order to support migration but might not want to change otherwise?

I'm not sure I follow - we'd put the migration-related functions (which is just `migrate_key_from_ctx` for now) so that we keep them separate from the main functionality, in case we'll need to implement more in the future?

tss-esapi/src/abstraction/transient.rs
```diff
@@ -494,6 +421,117 @@ impl TransientKeyContext {
...
+            .with_name_hashing_algorithm(HashingAlgorithm::Sha256)
+            .with_object_attributes(object_attributes);
+        match params {
+            KeyParams::Rsa {
+                size,
+                scheme,
+                pub_exponent,
+            } => {
+                let unique = match pub_key {
+                    None => PublicKeyRsa::default(),
```
Superhepper, line 466: Feel free to ignore this but that I had some difficulties reading that match-expression. So I came up with an alternative way of writing that might work. But if something is hard or5 easy to read is a completely subjective matter.

```rust
let unique = possible_pub_key
    .map(|pub_key| {
        if let PublicKey::Rsa(val) = pub_key {
            PublicKeyRsa::try_from(val)
        } else {
            Err(Error::local_error(ErrorKind::InconsistentParams))
        }
    })
    .transpose()?
    .unwrap_or_default();
```

Superhepper:
> 
> 
> > One thing to consider is if it would be a good idea to create a tss-esapi/src/abstraction/transient/migration.rs where we can put functions we need to change in order to support migration but might not want to change otherwise?
> 
> I'm not sure I follow - we'd put the migration-related functions (which is just `migrate_key_from_ctx` for now) so that we keep them separate from the main functionality, in case we'll need to implement more in the future?

I was thinking about ```create_restricted_decryption_rsa_public```. If the previous behaviour was 'more correct' but modifications were needed in order to support migration. The one option would be to keep the original one but then create a modified one that is only available internally.

ionut-arm:
> I was thinking about create_restricted_decryption_rsa_public. If the previous behaviour was 'more correct' but modifications were needed in order to support migration. The one option would be to keep the original one but then create a modified one that is only available internally.

Hmm, I think the "more correct" part describes the `RsaExponent::default()` behaviour, but not necessarily using it in `create_restricted_decryption_rsa_public` now, because it will lead to a rather subtle breaking change. As in, if someone is using that function and moves to the new version of the crate (changing the types they feed as parameters, but the values stay the same) and expects the result to be the same, they'd have the same issue as I did. That'd actually be an argument for modifying the other similar functions that in `utils/mod.rs`, the ones I'm not using anymore, `create_unrestricted_signing_rsa_public`...

ionut-arm:
> I was thinking about create_restricted_decryption_rsa_public. If the previous behaviour was 'more correct'

Wait, actually, the change I made in `create_restricted_decryption_rsa_public` was to use `default` instead of another method, so theoretically that should be the "correct" way? I'll switch the other functions to use `PublicKeyRsa::default()` for RSA unique values, does that sound sensible?

tss-esapi/src/abstraction/transient.rs
ionut-arm: Done

tss-esapi/src/abstraction/transient.rs
```diff
@@ -103,44 +145,25 @@ impl TransientKeyContext {
...
-                pub_exponent,
-            )?),
-            KeyParams::RsaEncrypt { size, pub_exponent } => {
-                create_unrestricted_encryption_decryption_rsa_public(size, pub_exponent)
-            }
-            KeyParams::Ecc { curve, scheme } => {
-                Ok(create_unrestricted_signing_ecc_public(scheme, curve)?)
-            }
-        }
+        let key_material = KeyMaterial {
```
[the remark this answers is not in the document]
ionut-arm: Done

tss-esapi/tests/transient_key_context.rs
ionut-arm: Done

tss-esapi/tests/transient_key_context.rs
ionut-arm: Fixed

tss-esapi/src/abstraction/transient.rs
ionut-arm: Done

tss-esapi/src/abstraction/transient.rs
```diff
@@ -494,6 +404,121 @@ impl TransientKeyContext {
...
+                            .with_exponent(pub_exponent)
+                            .with_is_signing_key(true)
+                            .with_is_decryption_key(decrypt_flag)
+                            .with_restricted(false)
+                            .build()?,
+                    )
+                    .with_rsa_unique_identifier(&unique);
+            }
+            KeyParams::Ecc { scheme, curve } => {
+                let unique = match pub_key {
```
Superhepper, line 475: Maybe you should change this one to look like its RSA counter part?

ionut-arm: Done - sorry, I mindlessly copied your example without checking.

ionut-arm:
Thank you everyone for the reviews! I think I've covered all the points made by now, so I'll leave this open for another couple of hours in case you want to have another look, then I'll merge.

wiktor-k:
LGTM :+1: great work!

ionut-arm:
I've squashed the commits, will merge when the CI passes.

wiktor-k:
> I've squashed the commits, will merge when the CI passes.

There is only one thing I miss that's available in Gitlab - "merge when CI succeeds" button...

anta5010:
there's "auto-commit" button which does exactly that

PR #263, opened 2021-09-29
Fix abstraction nv handle leaks

rshearman:


tss-esapi/src/abstraction/nv.rs
```diff
@@ -56,9 +57,14 @@ pub fn list(context: &mut Context) -> Result<Vec<(NvPublic, Name)>> {
...
             .collect::<Result<Vec<ObjectHandle>>>()?
-            .iter()
-            .map(|h| NvIndexHandle::from(*h))
-            .map(|h| ctx.nv_read_public(h))
+            .iter_mut()
+            .map(|h| {
+                let r = ctx.nv_read_public(NvIndexHandle::from(*h));
+                match (r, ctx.tr_close(h)) {
+                    (Ok(nv), Ok(_)) => Ok(nv),
+                    (Err(e), _) | (Ok(_), Err(e)) => Err(e),
```
Superhepper, line 65: This looks confusing to me. Actually I we should change the entire function to look something like this:
```rust
pub fn list(context: &mut Context) -> Result<Vec<(NvPublic, Name)>> {
    context.execute_without_session(|ctx| {
        ctx.get_capability(
            CapabilityType::Handles,
            TPM2_NV_INDEX_FIRST,
            TPM2_PT_NV_INDEX_MAX,
        )
        .and_then(|(capability_data, _)| match capability_data {
            CapabilityData::Handles(tpm_handles) => Ok(tpm_handles),
            _ => Err(Error::local_error(WrapperErrorKind::WrongValueFromTpm)),
        })
        .and_then(|tpm_handles| {
            tpm_handles
                .iter()
                .map(|&tpm_handle| {
                    ctx.tr_from_tpm_public(tpm_handle)
                        .and_then(|mut object_handle| {
                            ctx.nv_read_public(NvIndexHandle::from(object_handle))
                                .map_err(|e| {
                                    let _ = ctx.tr_close(&mut object_handle);
                                    e
                                })
                                .and_then(|(nv_public, name)| {
                                    ctx.tr_close(&mut object_handle)?;
                                    Ok((nv_public, name))
                                })
                        })
                })
                .collect()
        })
    })
}
```

Or maybe split it up in two functions.

Superhepper (changes requested):
Nice catch! I only had one comment regarding how that list function looks.

tss-esapi/src/abstraction/nv.rs
rshearman: I agree that it's a bit confusing and what you suggest doing with the iterator is clearer, but I'm not sure that the chain of and_thens in the main body of the execute_without_session closure aids readability.

rshearman: Ok.

Superhepper: Do you think it would be a good idea to split it up even further?

```rust
/// Lists all the currently defined NV Indexes' names and public components
pub fn list(context: &mut Context) -> Result<Vec<(NvPublic, Name)>> {
    context.execute_without_session(|ctx| {
        get_tpm_handles(ctx)
            .and_then(|tpm_handles| {
                tpm_handles
                    .iter()
                    .map(|&tpm_handle| get_info(ctx, NvIndexTpmHandle::try_from(tpm_handle)?))
                    .collect()
            })
    })
}

/// Returns tpm handles to currently defined NV Indexes
fn get_tpm_handles(&mut context) -> Result<Vec<TpmHandle>> {
        context.get_capability(
            CapabilityType::Handles,
            TPM2_NV_INDEX_FIRST,
            TPM2_PT_NV_INDEX_MAX,
        )
        .and_then(|(capability_data, _)| match capability_data {
            CapabilityData::Handles(tpm_handles) => Ok(tpm_handles),
            _ => Err(Error::local_error(WrapperErrorKind::WrongValueFromTpm)),
        })
}

/// Returns the NvPublic and Name associated with a nv index TPM handle.
fn get_info(
    context: &mut Context,
    nv_index_tpm_handle: NvIndexTpmHandle,
) -> Result<(NvPublic, Name)> {
    context
        .tr_from_tpm_public(nv_index_tpm_handle.into())
        .and_then(|mut object_handle| {
            context
                .nv_read_public(NvIndexHandle::from(object_handle))
                .map_err(|e| {
                    let _ = context.tr_close(&mut object_handle);
                    e
                })
                .and_then(|(nv_public, name)| {
                    context.tr_close(&mut object_handle)?;
                    Ok((nv_public, name))
                })
        })
}

rshearman: I don't see any particular need myself.

Superhepper: I meant from a readability point of view. If not then keep it as it is.

Superhepper (approved):

tss-esapi/src/abstraction/nv.rs
rshearman: Right, I don't see any particular need from a readability point of view. IMHO, the complexity of the functions is more than manageable.

ionut-arm (approved):

PR #264, opened 2021-09-29
Add property tag variant for MaxCapBuffer

rshearman:
As per "TCG TSS 2.0 Overview and Common Structures Specification" v0.9
rev 03, and tpm2-tss >= 2.4.0, add a constant for
TPM2_PT_MAX_CAP_BUFFER and a corresponding PropertyTag variant,
MaxCapBuffer.

This avoids an error for "value = 302 did not match any PropertyTag"
that may be seen with some TPM implementations.

Signed-off-by: Rob Shearman <rob@graphiant.com>

### DIFF
--- a/tss-esapi/src/constants/property_tag.rs
+++ b/tss-esapi/src/constants/property_tag.rs
@@ -56,6 +56,7 @@ pub enum PropertyTag {
     VendorCommands = TPM2_PT_VENDOR_COMMANDS,
     NvBufferMax = TPM2_PT_NV_BUFFER_MAX,
     Modes = TPM2_PT_MODES,
+    MaxCapBuffer = TPM2_PT_MAX_CAP_BUFFER,
     // Variable
     Permanent = TPM2_PT_PERMANENT,
     StartupClear = TPM2_PT_STARTUP_CLEAR,
--- a/tss-esapi/src/constants/tss.rs
+++ b/tss-esapi/src/constants/tss.rs
@@ -423,6 +423,7 @@ pub const TPM2_PT_LIBRARY_COMMANDS: TPM2_PT = TPM2_PT_FIXED + 42; /* number of c
 pub const TPM2_PT_VENDOR_COMMANDS: TPM2_PT = TPM2_PT_FIXED + 43; /* number of vendor commands that are implemented */
 pub const TPM2_PT_NV_BUFFER_MAX: TPM2_PT = TPM2_PT_FIXED + 44; /* the maximum data size in one NV write command */
 pub const TPM2_PT_MODES: TPM2_PT = TPM2_PT_FIXED + 45; /* a TPMA_MODES value indicating that the TPM is designed for these modes. */
+pub const TPM2_PT_MAX_CAP_BUFFER: TPM2_PT = TPM2_PT_FIXED + 46; /* the maximum size of a TPMS_CAPABILITY_DATA structure returned in TPM2_GetCapability(). */
 pub const TPM2_PT_VAR: TPM2_PT = TPM2_PT_GROUP * 2; /* the group of variable properties returned as TPMS_TAGGED_PROPERTY. The properties in this group change because of a Protected Capability other than a firmware update. The values are not necessarily persistent across all power transitions. */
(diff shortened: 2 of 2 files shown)


tss-esapi/src/constants/property_tag.rs
```diff
@@ -56,6 +56,7 @@ pub enum PropertyTag {
     VendorCommands = TPM2_PT_VENDOR_COMMANDS,
     NvBufferMax = TPM2_PT_NV_BUFFER_MAX,
     Modes = TPM2_PT_MODES,
+    MaxCapBuffer = TPM2_PT_MAX_CAP_BUFFER,
```
ionut-arm, line 59: Do you know if there are any other fields like this that we might've missed? Asking because it's a fairly annoying (for users) change to make, so I'm not sure if we should just mark this enum as `non_exhaustive`

ionut-arm (approved):
Looks alright. I think we need a wider discussion about potentially moving our dependency further from 2.3.3, we'll have to see how that plays on with any stability guarantees we have.

tss-esapi/src/constants/property_tag.rs
rshearman: Comparing with recent tpm2-tss, there are several new TPM2_CC constants, but they aren't used to construct an enum so there won't be any impact to users in adding those, either now or later.

I don't have a good feel for how likely this is to change in the future (presumably with future TCG spec revisions), so I don't have an opinion on whether or not to mark this as non_exhaustive.

Superhepper: I am all for moving to new version of tpm2-tss but as @ionut-arm said there will be a rather big impact for the crate users. I do not think it is reasonable to try to bump the version in the major release 7 of the crate but I think it would be quite possible to do so in version 8.

rshearman: Maybe I'm missing something, but why does this imply needing to move to a new version of tpm2-tss?

Superhepper: No there is no need to rise the version of tpm2-tss it was just me not reading the change nor the commit message properly.

Superhepper (approved):

tss-esapi/src/constants/property_tag.rs
ionut-arm: > but why does this imply needing to move to a new version of tpm2-tss?

Sorry, I think I didn't explain myself. Since this constant was added in version 2.4.0 and we mandate >= 2.3.3, it means there are some users for whom our crate builds properly (as they have 2.3.3), but they can't use this particular constant. This is combined with the fact that 2.3.3 is itself quite old, and we should move on a bit.

@Superhepper - why not bump the version in release 7?

rshearman: But the constant is defined in this crate and so is independent of the tpm2-tss version?

Superhepper: Yes, I tried to track the history of the constants.rs file but I could not find if it at some point it had been generated from something or if it was something that was handwritten by us. Anyway that files is now a days updated by us manually. Adding the constant should work fine because it is a value that is returned from the TPM and bound to the TPM specification and not a specific version of tpm2-tss library. It was just a define for the value that was missing in the library and it got added in 2.4.0.

Superhepper: @ionut-arm 
Well it is no problem to do it in major version 7 because it is still in alpha. I just thought it might be nice to give people a heads up before we do something like that.

ionut-arm: (About the constant) Ah, that's fair, I thought it was something the tpm2-tss library would also have to parse and understand, rather than just return from the TPM.

(About bumping the dependency version) Yeah, works for me to just wait for the next major release. It's a bit frustrating that there's still this gap between the spec and our support and I'm not sure how to bridge it, given that for every breaking change we have to bump the major. I'd lean towards just sticking with alpha for a while, releasing a 7.0.0, then whenever we get another breaking change we move to alpha for 8.0.0 and so on. It's nothing critical right now, it just constantly irks me

PR #265, opened 2021-09-29
Bump bindings version; add Darwin bindings

ionut-arm:
This commit bumps the version of bindgen we use in tss-esapi-sys
and updates the existing bindings with ones generated with the
new version. New bindings have also been added for x86-64 Darwin
platforms.


Superhepper (approved):
LGTM

tss-esapi-sys/build.rs
```diff
@@ -29,6 +29,7 @@ fn main() {
         match (target.architecture, target.operating_system) {
             (Architecture::Arm(_), OperatingSystem::Linux) => {}
             (Architecture::Aarch64(_), OperatingSystem::Linux) => {}
+            (Architecture::X86_64, OperatingSystem::Darwin) => {}
```
hug-dev, line 32: 🥳

tss-esapi-sys/regenerate-bindings.sh
```diff
@@ -0,0 +1,92 @@
...
+# Download OpenSSL source code
+if [ ! -d "/tmp/openssl" ]; then
+    pushd /tmp
+    git clone https://github.com/openssl/openssl.git --branch $OPENSSL_VERSION
+    popd
+fi
+
+# Regenerate bindings for x86_64-unknown-linux-gnu
+cargo clean
+cargo build --features generate-bindings
```
hug-dev, line 68: It would be nice to make the `generate-bindings` feature copy the bindings to their right location and use that directly, to not have to copy it manually. That would be good for all of our `-sys` crate actually.

hug-dev (approved):
Thanks 👌 New fresh bindings

tss-esapi-sys/regenerate-bindings.sh
hug-dev: ah yeah true... And in some places those directories are read-only

PR #266, opened 2021-10-01
Add TPM Import command

wiktor-k:
This PR adds TPM Import command - dual to the Duplicate command that we already have. The PR also modifies the duplicate test case to include testing for import.


Superhepper (approved):
Thank you! Looks good to me.

tss-esapi/src/context/tpm_commands/duplication_commands.rs
```diff
@@ -331,5 +333,356 @@ impl Context {
...
+    ) -> Result<Private> {
+        let mut out_private = null_mut();
+        let ret = unsafe {
+            Esys_Import(
+                self.mut_context(),
+                parent_handle.into(),
+                self.required_session_1()?,
+                self.optional_session_2(),
+                self.optional_session_3(),
+                &encryption_key.unwrap_or_default().into(),
```
ionut-arm, line 671: Hmm, following [the discussion on Slack](https://cloud-native.slack.com/archives/C01EARH2ZB3/p1633333906080300), maybe this should be documented (by this I mean what the result is of using `None` for `encryption_key`)

ionut-arm (commented):
Thanks! Code looks good, just a line more of doc needed

tss-esapi/src/context/tpm_commands/duplication_commands.rs
wiktor-k: Yep :+1: I'll do that for import and duplicate in a separate commit in a moment.

Browsing through the code it seems `unwrap_or_default` appears heavily in [`create_primary`](https://docs.rs/tss-esapi/7.0.0-alpha.1/src/tss_esapi/context/tpm_commands/hierarchy_commands.rs.html#30-96). I'm thinking through @Superhepper's ["explicit default"](https://cloud-native.slack.com/archives/C01EARH2ZB3/p1633340711082700?thread_ts=1633333906.080300&cid=C01EARH2ZB3) and yeah, if writing `Default::default()` was as short as `None` I wouldn't see a reason to prefer `None`. (it seems `Default::default` [cannot be imported directly](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=0012f4cb66c28b48a6aa43a57f2c0991), but having `None` and `Some(Default::default())` mean the same seems redundant at best).

Superhepper (approved):
Looks good to me

wiktor-k:
Okay, I've given it a second thought, changed a bit in my client crate and looked at what [`load` does](https://github.com/parallaxsecond/rust-tss-esapi/blob/main/tss-esapi/src/context/tpm_commands/object_commands.rs#L121). And now I think it's OK if we take Public by value as the low-level function needs it this way anyway. `load` clones internally and I think this is sub-optimal (because it hides complexity) but probably this is a legacy artifact.

So, sorry for the noise. I got to the same conclusion as @ionut-arm and I think this PR is yet again, eh... ready for re-review :)

PR #267, opened 2021-10-01
Fix tss2_esys version detection + new load_external test

rshearman:
* Fix checks for tpm2-tss version
* Export the tss2_esys library version even when not generating bindings
* Add a test for TPM2_LoadExternal

Without the fixes the new TPM2_LoadExternal test fails for as I'm using tpm2-tss v3.0.x as the translation of the null handle doesn't match what tpm2-tss requires.


ionut-arm:
Does the test actually "prove" anything when using the 2.3.3 library? I get the feeling we should instead create another test workflow where we use v3 of tpm2-tss, if we're claiming to "support" it, from an interface perspective. It doesn't have to be in this PR, though

rshearman:
> Does the test actually "prove" anything when using the 2.3.3 library? I get the feeling we should instead create another test workflow where we use v3 of tpm2-tss, if we're claiming to "support" it, from an interface perspective. It doesn't have to be in this PR, though

The test still has value when using the 2.3.3 library, i.e. that the hierarchy is translated into the correct type of handle for that library version, and that the public portion of the object is correctly translated into something the TPM understands.

However, you are right that in order to prove that the fix/test works with v3 of tpm2-tss then another test workflow would be required (assuming the current ones are 2.3.3 only).

Superhepper (approved):

PR #270, opened 2021-10-04
Add a hint for unique_identifier functions

wiktor-k:
Previously a unique identifier was not explicitly required and if not
provided a default value was used. After migration it is not clear what
value need to be passed so this patch improves the error message
suggesting the `Default::default()` as the value that will preserve old
behavior in a code that tries to upgrade to tss-esapi 7.x.


tss-esapi/src/structures/buffers/public.rs
```diff
@@ -233,7 +233,7 @@ impl PublicBuilder {
                         Error::local_error(WrapperErrorKind::ParamsMissing)
                     })?,
                     unique: self.rsa_unique_identifier.ok_or_else(|| {
-                        error!("RSA unique identifier has not been set in the PublicBuilder even though the RSA algorithm had been selected.");
+                        error!("RSA unique identifier has not been set in the PublicBuilder even though the RSA algorithm had been selected. Consider using: .with_rsa_unique_identifier(&Default::default())");
```
ionut-arm, line 236: Could you please replace the `Default` with the corresponding type, e.g. `PublicKeyRsa` in this case? Seemsa tiny bit easier for people to figure out what they're doing, and what else they could be doing.

wiktor-k: Fixed!

ionut-arm (approved):
Thanks!

Superhepper (approved):

PR #271, opened 2021-10-04
Rewrite builders to store symmetric value unwrapped (and a couple of small fixes)

wiktor-k:
Okay, this PR was about reworking internal implementation of Builders with regards to symmetric but then I found out a couple of small inconsistencies (probably due to code rewriting) but as they are small I packed everything in a bigger PR.

I hope this is still easy to review. Comments welcome!


tss-esapi/src/structures/buffers/public/ecc.rs
```diff
@@ -35,12 +35,12 @@ impl PublicEccParametersBuilder {
     }
 
     /// Create parameters for a restricted decryption key (i.e. a storage key)
-    pub fn new_restricted_decryption_key(
+    pub const fn new_restricted_decryption_key(
         symmetric: SymmetricDefinitionObject,
         curve: EccCurve,
```
Superhepper, line 40: Why would I want to specify EccCurve if I create a RSA key?

ionut-arm (approved):
Seems fair to me

Superhepper:
Sorry for taking such long time to review this. But I am thinking if it from a users perspective would be useful to differentiate between the case where ```Null``` has been set but it is not allowed or the value has not been set at all. 

I mean would it be confusing if I had used ```new``` then not set symmetric and I receive an error that said that ```Null``` was not allowed for a parameter that I have not even specified?

wiktor-k:
Thanks for the review @Superhepper!

> I mean would it be confusing if I had used new then not set symmetric and I receive an error that said that Null was not allowed for a parameter that I have not even specified?

That's a restricted+decryption key use case. The [error in this case would be](https://github.com/parallaxsecond/rust-tss-esapi/pull/271/files#diff-e58d6a9d97c4bd2b0824f77af142a72ef856b101215562d299f279a6289f42a7R153) `Symmetric parameter has not been provided even though 'restricted' and 'is_decrypt_key' are set to true`.

Maybe it's my Java background but reading that "null was not allowed for a parameter" I'd immediately think it hasn't been set explicitly where it needs to.

I could still rework the PR but internally it still needs to treat None and Null as the same values (maybe with the exception of error reporting?) as currently setting symmetric to Null in keys that are not restricted+decryption triggers the `Found symmetric parameter, expected it to be None because 'restricted' and 'is_decrypt_key' are set to false` error which I think it's not OK.

Superhepper:
> Maybe it's my Java background but reading that "null was not allowed for a parameter" I'd immediately think it hasn't been set explicitly where it needs to.

That is Java or C# for you.

Yes, ```Some(Null)``` and ```None``` would be treated the same way (i.e. they would both result in Null). I did not think this had to be changed I was merely wondering if it would be useful to differentiate between the two scenarios. If you do not think it would make any difference from a users perspective, that is making it easier to understand what might have gone wrong. The keep it as it is.

wiktor-k:
Okay, rewritten this quite a bit and also added test cases simplified from the code I'm using. @Superhepper if you could see if commit https://github.com/parallaxsecond/rust-tss-esapi/pull/271/commits/4f31afa7141273dd7615560bf5dfe2149fcd1316 does what you think is OK I'd greatly appreciate it! (if it's OK I'll add the similar check to ECC builder)

wiktor-k:
Okay. Moved tests to a separate directory, added tests and checks for ECC and RSA keys (in separate commits for easier review) and added more tests with error cases too. @Superhepper if you could re-review it I'd greatly appreciate it! :bow:

Issue #279, opened 2021-10-15
Document PublicKey


ionut-arm:
The `PublicKey` type (used in `TransientKeyContext`) isn't documented as to what the format of the data in each variant is. This must be rectified.

ionut-arm:
More documentation (more specifically code examples for how to actually use the `TransientKeyContext`) should be added, to make the context more user-friendly.

bootstrap-prime:
I'm attempting to build an ssh-agent on top of this library and the lack of documentation is confusing. Judging from parsec, is it just the raw components? Even if you don't document them, could you describe them? Same for signature. also judging from parsec, I think they're also raw?

ionut-arm:
Hey! Thanks for the feedback, I'll try to add some documentation today, though it will focus on `TransientKeyContext` - for the lower level stuff (around `Context`) it's better to just use the spec as the documentation, and we can help with navigating that as well. I'll also summarise it in a comment here, since I'm not sure when we'll be pushing out the next release.

ionut-arm:
I've opened #347 - let me know if you'd also like a summary here as well, or if the PR is enough to get you going :)

bootstrap-prime:
That helps, thank you very much! I might have more questions in the future but this is probably enough to get me going.

ionut-arm:
Done in #347

PR #284, opened 2021-10-28
Add ActivateCredential support for TKC

ionut-arm:
This commit adds support for two operations, enabling key attestation
via the ActivateCredential call. A test has also been added to verify
attestation using the Endorsement Key.

Intended for https://github.com/parallaxsecond/parsec/issues/539

cc @wiktor-k

### DIFF
--- a/tss-esapi/src/abstraction/transient/key_attestation.rs
+++ b/tss-esapi/src/abstraction/transient/key_attestation.rs
@@ -10,36 +10,23 @@ use crate::{
         session_handles::{AuthSession, PolicySession},
     },
     structures::{EncryptedSecret, IDObject, SymmetricDefinition},
-    tss2_esys::{TPM2B_PUBLIC, TPMT_PUBLIC},
+    tss2_esys::{Tss2_MU_TPM2B_PUBLIC_Marshal, TPM2B_PUBLIC},
     utils::PublicKey,
-    Result,
+    Error, Result,
 };
-use std::convert::{TryFrom, TryInto};
+use log::error;
+use std::convert::TryFrom;
 
 #[derive(Debug)]
 /// Wrapper for the parameters needed by MakeCredential
 pub struct MakeCredParams {
     /// TPM name of the object
-    name: Vec<u8>,
+    pub name: Vec<u8>,
     /// Encoding of the public parameters of the object whose name
     /// will be included in the credential computations
-    public: Vec<u8>,
+    pub public: Vec<u8>,
     /// Public part of the key used to protect the credential
-    attesting_key_pub: PublicKey,
-}
-
-impl MakeCredParams {
-    pub fn name(&self) -> &[u8] {
-        &self.name
-    }
-
-    pub fn public(&self) -> &[u8] {
-        &self.public
-    }
-
-    pub fn attesting_key_pub(&self) -> &PublicKey {
-        &self.attesting_key_pub
-    }
+    pub attesting_key_pub: PublicKey,
 }
 
 impl TransientKeyContext {
@@ -68,29 +55,29 @@ impl TransientKeyContext {
         self.context.flush_context(object_handle.into())?;
 
         let public = TPM2B_PUBLIC::from(object_public);
-        let public = unsafe {
-            std::mem::transmute::<TPMT_PUBLIC, [u8; std::mem::size_of::<TPMT_PUBLIC>()]>(
-                public.publicArea,
+        let mut pub_buf = [0u8; std::mem::size_of::<TPM2B_PUBLIC>()];
+        let mut offset = 0;
+        let result = unsafe {
+            Tss2_MU_TPM2B_PUBLIC_Marshal(
+                &public,
+                &mut pub_buf as *mut u8,
+                pub_buf.len() as u64,
+                &mut offset,
             )
         };
-        let attesting_key_pub = match key {
-            None => {
-                let key_handle =
-                    ek::create_ek_object(&mut self.context, AsymmetricAlgorithm::Rsa, None)?;
-                let (attesting_key_pub, _, _) =
-                    self.context.read_public(key_handle).or_else(|e| {
-                        self.context.flush_context(key_handle.into())?;
-                        Err(e)
-                    })?;
-                self.context.flush_context(key_handle.into())?;
+        let result = Error::from_tss_rc(result);
+        if !result.is_success() {
+            error!("Error in marshalling TPM2B");
+            return Err(result);
+        }
 
-                attesting_key_pub.try_into()?
-            }
+        let attesting_key_pub = match key {
+            None => get_ek_object_public(&mut self.context)?,
             Some(key) => key.material.public,
         };
         Ok(MakeCredParams {
             name: object_name.value().to_vec(),
-            public: public.to_vec(),
+            public: pub_buf.to_vec(),
             attesting_key_pub,
         })
     }
@@ -202,4 +189,15 @@ impl TransientKeyContext {
             session,
         ))
     }
+}
+
+fn get_ek_object_public(context: &mut crate::Context) -> Result<PublicKey> {
+    let key_handle = ek::create_ek_object(context, AsymmetricAlgorithm::Rsa, None)?;
+    let (attesting_key_pub, _, _) = context.read_public(key_handle).or_else(|e| {
+        context.flush_context(key_handle.into())?;
+        Err(e)
+    })?;
+    context.flush_context(key_handle.into())?;
+
+    PublicKey::try_from(attesting_key_pub)
 }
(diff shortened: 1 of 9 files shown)


tss-esapi/src/abstraction/transient/key_attestation.rs
```diff
@@ -0,0 +1,203 @@
...
+        secret: Vec<u8>,
+    ) -> Result<Vec<u8>> {
+        let credential_blob = IDObject::try_from(credential_blob)?;
+        let secret = EncryptedSecret::try_from(secret)?;
+        let object_handle = self.load_key(object.params, object.material, object.auth)?;
+        let session_2;
+        let key_handle = match key {
+            None => {
+                // No key was given, use the EK. This requires using a Policy session
+                session_2 = self
```
Superhepper, line 125: I think it should probably put the two cases in separate private methods so its get easier to read.

ionut-arm: Ok, done, though I kept the `match key` - I find it more intuitive to follow

anta5010 (approved):

tss-esapi/src/abstraction/transient/key_attestation.rs
```diff
@@ -0,0 +1,205 @@
...
+
+        let public = TPM2B_PUBLIC::from(object_public);
+        let public = unsafe {
+            std::mem::transmute::<TPMT_PUBLIC, [u8; std::mem::size_of::<TPMT_PUBLIC>()]>(
+                public.publicArea,
+            )
+        };
+        let attesting_key_pub = match key {
+            None => {
+                let key_handle =
```
Superhepper, line 78: Would it possible to put this in a ```get_ek_object_public``` method?

tss-esapi/src/abstraction/transient/key_attestation.rs
```diff
@@ -0,0 +1,203 @@
...
+                &mut pub_buf as *mut u8,
+                pub_buf.len() as u64,
+                &mut offset,
+            )
+        };
+        let result = Error::from_tss_rc(result);
+        if !result.is_success() {
+            error!("Error in marshalling TPM2B");
+            return Err(result);
+        }
```
ionut-arm, line 72: I wonder if we should implement methods on the types that can be marshalled, to convert them to `Vec<u8>` using these functions from the library

tss-esapi/src/abstraction/transient/key_attestation.rs
ionut-arm: Done!

Superhepper (approved):

ionut-arm:
@Superhepper @anta5010 - want to review again the last two commits? The one about changing what type we serialize is more consequential, the other just adds tests.

anta5010 (approved):

tss-esapi/src/abstraction/transient/key_attestation.rs
```diff
@@ -0,0 +1,208 @@
...
+    /// * `object` - the object whose TPM name is included in the credential
+    /// * `key` - the key used to encrypt the secret that wraps the credential
+    /// * `credential_blob` - encrypted credential that will be returned by the
+    /// TPM
+    /// * `secret` - encrypted secret that was used to encrypt the credential
+    ///
+    /// **Note**: if no `key` is given, the default Endorsement Key
+    /// will be used. You can find more information about the default Endorsement
+    /// Key in the [ek] module.
+    pub fn activate_credential(
```
wiktor-k, line 103: It would be nice if this also had a code example like the `Context` functions. Copying something working from end to end tests is OK for first version IMHO. (doctests are a good way for newcomers to learn how to use the API)

tss-esapi/src/abstraction/transient/key_attestation.rs
```diff
@@ -0,0 +1,208 @@
...
+                Err(e)
+            })?;
+        self.context.flush_context(object_handle.into())?;
+
+        // Name of objects is derived from their publicArea, i.e. the marshaled TPMT_PUBLIC
+        let public = TPM2B_PUBLIC::from(object_public);
+        let mut pub_buf = [0u8; std::mem::size_of::<TPMT_PUBLIC>()];
+        let mut offset = 0;
+        let result = unsafe {
+            Tss2_MU_TPMT_PUBLIC_Marshal(
```
wiktor-k, line 62: Not sure if this was mentioned elsewhere but this looks like a good idea to have this function exposed for other code as well. Would it make sense to move it to the public wrapper or do you think it's not worth it?

wiktor-k (commented):
Okay. In general this looks OK and I posted some minor issues.
I'll check out https://github.com/parallaxsecond/parsec/pull/558 on how to handle this API :)

tss-esapi/src/abstraction/transient/key_attestation.rs
ionut-arm: I'll add this as a TODO in #279 and will add examples for most/all of the methods in `TransientKeyContext`.

tss-esapi/src/abstraction/transient/key_attestation.rs
ionut-arm: I'll raise an issue for it and we can figure it out independently of this PR, but yeah, I think it would be a good addition to the API, and it won't break the API either.

tss-esapi/src/abstraction/transient/key_attestation.rs
```diff
@@ -0,0 +1,208 @@
...
+use std::convert::{TryFrom, TryInto};
+
+#[derive(Debug)]
+/// Wrapper for the parameters needed by MakeCredential
+pub struct MakeCredParams {
+    /// TPM name of the object
+    pub name: Vec<u8>,
+    /// Encoding of the public parameters of the object whose name
+    /// will be included in the credential computations
+    pub public: Vec<u8>,
```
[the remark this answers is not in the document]
ionut-arm: For name it's true, I could just return `Name`, and let the user deserialize, but they're both intended for sending across to a 3rd party to check. And `public` isn't exactly the same as `Public` - it's the marshaled version of the `TPMT_PUBLIC` inside `TPM2B_PUBLIC`, so it'd be a bit of low-level work for users, which doesn't really align well with this abstraction level. `public` isn't _strictly_ required to compute `MakeCredential`, it's more like helper data, meant for the 3rd party to be able to easily verify that the object they're creating the credential for is the correct one.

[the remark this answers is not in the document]
wiktor-k: Okay, the reasoning is sound. Please leave name as is but maybe it would be a good idea to enrich the doc comment mentioning what you said, e.g. `Wrapper for the parameters needed by MakeCredential intended for sending across to a 3rd party to check` (or something similar). (A similar API exists for U2F attestations so I guess people will know why it's bytes if they see it's for sending...)

Thanks!

wiktor-k (approved):

tss-esapi/tests/integration_tests/abstraction_tests/transient_key_context_tests.rs
```diff
@@ -597,3 +600,232 @@ fn ctx_migration_test() {
...
+    let (material, auth) = ctx.create_key(params, 16).unwrap();
+    let obj = ObjectWrapper {
+        material,
+        auth,
+        params,
+    };
+
+    let _ = ctx
+        .activate_credential(obj.clone(), None, vec![], vec![])
+        .unwrap_err();
```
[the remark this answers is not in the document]
ionut-arm: Done, though now I have to understand why it doesn't work the same way with `swtpm`.

tss-esapi/src/abstraction/transient/key_attestation.rs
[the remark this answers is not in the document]
ionut-arm: Done

tss-esapi/src/abstraction/transient/key_attestation.rs
ionut-arm: Done

tss-esapi/src/abstraction/transient/key_attestation.rs
ionut-arm: Done, #291

tss-esapi/tests/integration_tests/abstraction_tests/transient_key_context_tests.rs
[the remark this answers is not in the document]
ionut-arm: Understood, the `swtpm` returns `FAILURE` for those types of errors, for some reason. I updated the test to check for either `TPM_RC_VALUE` or `TPM_RC_FAILURE`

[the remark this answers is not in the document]
wiktor-k: Great! Actually it's nice to see this issue pop-up here. When I worked on my client's code I frequently consulted your test code so you just saved someone debugging in the future :)

PR #221, opened 2021-04-15
Add TPM2_Certify

ionut-arm:
This commit adds the TPM2_Certify command and Rust native types for all
parameters and return types involved in the signature of the certify
function. A new test is also included.

TODO: 

- [x] documentation

- [x] more tests

PR #286, opened 2021-11-05
Added SignatureScheme type.

Superhepper:
- Took implementation of SignatureScheme from #221 and created
a separate PR for it in order to remove bindgen type from context
methods and adding conversion test for the type.

### DIFF
--- a/tss-esapi/src/structures/tagged/schemes.rs
+++ b/tss-esapi/src/structures/tagged/schemes.rs
@@ -3,12 +3,12 @@
 use crate::{
     interface_types::algorithm::{
         EccSchemeAlgorithm, HashingAlgorithm, KeyDerivationFunction, KeyedHashSchemeAlgorithm,
-        RsaDecryptAlgorithm, RsaSchemeAlgorithm,
+        RsaDecryptAlgorithm, RsaSchemeAlgorithm, SignatureSchemeAlgorithm,
     },
     structures::schemes::{EcDaaScheme, HashScheme, HmacScheme, XorScheme},
     tss2_esys::{
         TPMT_ECC_SCHEME, TPMT_KDF_SCHEME, TPMT_KEYEDHASH_SCHEME, TPMT_RSA_DECRYPT, TPMT_RSA_SCHEME,
-        TPMU_ASYM_SCHEME, TPMU_KDF_SCHEME, TPMU_SCHEME_KEYEDHASH,
+        TPMT_SIG_SCHEME, TPMU_ASYM_SCHEME, TPMU_KDF_SCHEME, TPMU_SCHEME_KEYEDHASH, TPMU_SIG_SCHEME,
     },
     Error, Result, WrapperErrorKind,
 };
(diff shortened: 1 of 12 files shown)


tss-esapi/src/structures/tagged/schemes.rs
```diff
@@ -543,3 +543,114 @@ impl TryFrom<RsaScheme> for RsaDecryptionScheme {
...
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum SignatureScheme {
+    RsaSsa { hash_scheme: HashScheme },
+    RsaPss { hash_scheme: HashScheme },
+    EcDsa { hash_scheme: HashScheme },
+    Sm2 { hash_scheme: HashScheme },
+    EcSchnorr { hash_scheme: HashScheme },
+    EcDaa { ecdaa_scheme: EcDaaScheme },
+    Hmac { hmac_scheme: HmacScheme },
+    Null { hash_scheme: HashScheme },
```
ionut-arm, line 560: is the inner `hash_scheme` needed here? based on the spec it seems when the selector is null the details are irrelevant

tss-esapi/tests/integration_tests/structures_tests/tagged_tests/tagged_signature_scheme_tests.rs
```diff
@@ -0,0 +1,138 @@
...
+
+        assert_eq!(
+            native_actual, native_expected,
+            "The actual SignatureScheme did not match expected",
+        );
+    };
+}
+
+macro_rules! test_valid_conversions_hash_scheme {
+    (SignatureScheme::$item:ident, $union_field_name:ident) => {
```
ionut-arm, line 93: Hmm, I think you might be able to keep all of the cases in the same macro with something like
```
(SignatureScheme::EcDaa, $union_field_name:ident) => {
```
before the generic macro, but I'm not entirely sure

ionut-arm (commented):
I'm ok with the changes, but waiting for consensus on the `Null` scheme thing

tss-esapi/src/structures/tagged/schemes.rs
Superhepper: hrrm I think it is used because of 'any'. But I think it would be better to add Any instead and let Null be Null. Or what do you think?

tss-esapi/tests/integration_tests/structures_tests/tagged_tests/tagged_signature_scheme_tests.rs
Superhepper: Interesting, I will look into that. Thanks.

tss-esapi/src/structures/tagged/schemes.rs
ionut-arm: I'm not entirely sure, the spec isn't very clear - what exact selector corresponds to `Any`? I don't know if I saw one... I did see that for `Null` you don't have any inner type

Superhepper: Any is used to set hash_scheme for any of the selectors that has a hash scheme. So it would look exactly as Null does today.

ionut-arm: Ah, fair, yeah, then rebranding this one as `Any` and having a separate, empty `Null` sounds good

Superhepper: Hrrmm it was not that easy. There is no way to parse 'Any' it has no selector. So I think I just will skip it.

tss-esapi/tests/integration_tests/structures_tests/tagged_tests/tagged_signature_scheme_tests.rs
Superhepper: Fixed

tss-esapi/src/structures/tagged/schemes.rs
Superhepper: I have been looking through every specification I could find in order to figure out how any works. And I have still not really understood it. So I need to do some more digging before fixing this.

Superhepper: Ok I think I have come up with suitable substitute for ```anySig```. Don't know if it something we want though.

tss-esapi/src/structures/tagged/schemes.rs
```diff
@@ -543,3 +543,168 @@ impl TryFrom<RsaScheme> for RsaDecryptionScheme {
...
+
+    /// Sets digest( i.e. hashing algorithm) of a signing scheme.
+    ///
+    /// # Details
+    /// This is intended to provide the functionality of writing
+    /// to the 'anySig' field in the TPMU_SIG_SCHEME union.
+    ///
+    /// # Error
+    /// Returns an InvalidParam error if the trying to read from
+    /// SignatureScheme that is not a signing scheme.///
```
wiktor-k, line 601: Unnecessary `///` at the end.

tss-esapi/src/structures/tagged/schemes.rs
```diff
@@ -543,3 +543,168 @@ impl TryFrom<RsaScheme> for RsaDecryptionScheme {
...
+    /// Null scheme, representing "no scheme" or "default"
+    pub const NULL: SignatureScheme = SignatureScheme::Null {
+        hash_scheme: HashScheme::new(HashingAlgorithm::Null),
+    };
+
+    /// Returns the digest( i.e. hashing algorithm) of a signing scheme.
+    ///
+    /// # Details
+    /// This is intended to provide the functionality of reading
+    /// from the 'anySig' field in the TPMU_SIG_SCHEME union.
```
wiktor-k, line 573: Using backticks (```) in `anySig` would render them in fixed font.

tss-esapi/tests/integration_tests/context_tests/tpm_commands/attestation_commands_tests.rs
```diff
@@ -33,7 +27,7 @@ mod test_quote {
             .quote(
                 key_handle,
                 &Data::try_from(qualifying_data).unwrap(),
-                scheme,
+                SignatureScheme::NULL,
```
wiktor-k, line 30: Nice to see this finally being resolved :clap:

wiktor-k (approved):
:thinking: This looks fine for me (with minor issues). Now it will be the `TPMT_TK_HASHCHECK` that strains my eyes ;)

tss-esapi/src/structures/tagged/schemes.rs
Superhepper: fixed

tss-esapi/src/structures/tagged/schemes.rs
Superhepper: fixed

tss-esapi/src/structures/tagged/schemes.rs
ionut-arm: Is that the new `set_signing_scheme` method below? I'm ok with that as an option. Still think `Null` should be void of parameters, though

Superhepper: That is just me not paying attention. It should be void.

ionut-arm (approved):
Excellent, thanks!

PR #288, opened 2021-11-09
Change Name to TPM2B_NAME conversion

ionut-arm:
This commit fixes a current issue flagged on the CI where conversion
from Name to its FFI equivalent is not done properly.

cc @wiktor-k

### DIFF
--- a/tss-esapi/src/structures/names/name.rs
+++ b/tss-esapi/src/structures/names/name.rs
@@ -35,7 +35,7 @@ impl TryFrom<Vec<u8>> for Name {
         }
         let size = bytes.len() as u16;
         let mut name = [0; Name::MAX_SIZE];
-        name.copy_from_slice(&bytes);
+        name[..bytes.len()].copy_from_slice(&bytes);
         Ok(Name {
             value: TPM2B_NAME { size, name },
         })
--- a/tss-esapi/src/context/tpm_commands/enhanced_authorization_ea_commands.rs
+++ b/tss-esapi/src/context/tpm_commands/enhanced_authorization_ea_commands.rs
@@ -429,8 +429,8 @@ impl Context {
                 self.optional_session_1(),
                 self.optional_session_2(),
                 self.optional_session_3(),
-                &object_name.try_into()?,
-                &new_parent_name.try_into()?,
+                &object_name.into(),
+                &new_parent_name.into(),
                 if include_object { 1 } else { 0 },
             )
         };


wiktor-k:
Thanks for the fix Ionut! It seems it was a drift between the Name refactoring and others.

It seems it still prints some errors:

```
[2021-11-09T16:03:15Z ERROR tss_esapi::structures::buffers::nonce] Error: Invalid Vec<u8> size(> 64)
[2021-11-09T16:03:15Z ERROR tss_esapi::structures::buffers::public::ecc] Found symmetric parameter, expected it to be Null nor not set at all because 'restricted' and 'is_decrypt_key' are set to false
```

I wonder if that's due to the dummy Name construction in the duplicate test case 🤔

ionut-arm:
I'm very confused as to how the CI passed initially 🤔

wiktor-k:
My guess is that the CI checks only the top commit and not the merge commit that is created when you press the Merge button. This is because it would have to rebuild all open PRs each time one PR is merged and maybe GitHub wants to optimize... 

Thus there can be two green PRs but when you combine them they fail in a subtle way. 🤷‍♂️

This is just my guess. I don't have any insider knowledge ☺️

Superhepper:
The tests were made a long time ago. And it was not approved by me until recently so the tests were out of date when the actual merge were conducted. Tests for the current PRs are not re-run just because something else gets merged.

Superhepper:
I wish there was a configuration we could set so that when something gets merged all the remianing PRs gets status 'merge conflict' so that you would be forced to rebase upon the latest changes from main and then trigger a run from the CI.

Superhepper (approved):

ionut-arm:
> 
> 
> I wish there was a configuration we could set so that when something gets merged all the remianing PRs gets status 'merge conflict' so that you would be forced to rebase upon the latest changes from main and then trigger a run from the CI.

There is an option - not sure how to "debate" enabling this, though. Do we have some sort of poll here in Github? I'm ok with the idea - it puts a damper on fast-moving projects, but I wouldn't say this repo moves too quickly, so it wouldn't slow us down

tss-esapi/src/structures/names/name.rs
```diff
@@ -35,7 +35,7 @@ impl TryFrom<Vec<u8>> for Name {
         }
         let size = bytes.len() as u16;
         let mut name = [0; Name::MAX_SIZE];
-        name.copy_from_slice(&bytes);
+        name[..bytes.len()].copy_from_slice(&bytes);
```
anta5010, line 38: why not `..size`?

anta5010 (approved):

tss-esapi/src/structures/names/name.rs
ionut-arm: Because `size` is `u16`, while this number has to be `usize`. Could do a cast, but it ends up being just as bad

wiktor-k:
> There is an option - not sure how to "debate" enabling this, though. Do we have some sort of poll here in Github? 

I've seen reactions being (ab)used for votes. E.g. you write: ":+1: for rebase workflow :-1: for keeping it as is" etc etc. and count the votes after some time (if you do go with this approach send the info to slack so that all people can participate).

> I'm ok with the idea - it puts a damper on fast-moving projects, but I wouldn't say this repo moves too quickly, so it wouldn't slow us down

Yep, agreed. FTR in Sequoia we require rebasing on top of main every time and it has been working fine. Just like you said fast-moving projects may not like it but for now it should make the merges safer (and if it becomes problematic the decision can be reverted too).

PR #289, opened 2021-11-10
Suppress deref_nullptr warnings

ionut-arm:


ionut-arm:
> 
> 
> Is it possible to only disable it for one module?

This is only for the `-sys` crate which doesn't really have any modules. I think it won't be applied up (to the `tss-esapi` crate, for example)

Superhepper (approved):

anta5010 (approved):

PR #293, opened 2021-11-17
Adds the attest structures

Superhepper:
Adds the Attest type and all the different attest info types.


tss-esapi/src/structures/buffers/attest.rs
```diff
@@ -0,0 +1,86 @@
...
+use log::error;
+use std::{convert::TryFrom, ops::Deref};
+use zeroize::Zeroizing;
+
+/// Attestation data buffer.
+///
+/// # Details
+/// Corresponds to `TPM2B_ATTEST`. The contents of
+/// the buffer can be unmarshalled into an [Attest]
+/// structure.
```
ionut-arm, line 14: Maybe you can copy the unmarshaling implementation from [here](https://github.com/parallaxsecond/rust-tss-esapi/pull/221/files#diff-4dc0721739064edc9acea6781a4ad04a612679852920fca6c3b8232d741d3a18R259-R281)?

ionut-arm (commented):
🤯 That's a lot of new types, thanks!! Looks pretty good overall, had just a few comments/questions.

I'm also curious, I thought there was at least one place where one of the FFI types is used, `TPM2B_ATTEST`, I think. Did you not want to update that yet?

tss-esapi/src/interface_types/structure_tags.rs
```diff
@@ -0,0 +1,68 @@
+// Copyright 2021 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{constants::StructureTag, tss2_esys::TPMI_ST_ATTEST, Error, Result, WrapperErrorKind};
+use std::convert::TryFrom;
+
+/// Type of attestation.
+///
+/// # Details
+/// Corresponds to `TPMI_ST_ATTEST`.
```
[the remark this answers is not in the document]
Superhepper: Yes, TPMI is for interface types

tss-esapi/src/structures/attest.rs
```diff
@@ -0,0 +1,118 @@
...
+/// # Details
+/// Corresponds to `TPMS_ATTEST`.
+#[derive(Debug, Clone)]
+pub struct Attest {
+    attestation_type: AttestationType,
+    qualified_signer: Name,
+    extra_data: Data,
+    clock_info: ClockInfo,
+    firmware_version: u64,
+    attested: AttestInfo,
```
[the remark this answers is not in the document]
Superhepper: Well not really I guess. I just put it there because it is how the native type is structured I think.

tss-esapi/src/structures/buffers/attest.rs
Superhepper: Yes, I will do that.

tss-esapi/src/interface_types/structure_tags.rs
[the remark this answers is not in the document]
Superhepper: So I have placed all TPMI under interface types.

tss-esapi/src/structures/buffers/attest.rs
Superhepper: fixed

tss-esapi/src/structures/attest.rs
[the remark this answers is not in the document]
Superhepper: it is not needed but convenient to have it for converting purposes.

tss-esapi/src/structures/attest.rs
```diff
@@ -0,0 +1,173 @@
...
+                })?,
+                &mut offset,
+            )
+        });
+
+        if !ret.is_success() {
+            return Err(ret);
+        }
+
+        unsafe { buffer.set_len(Self::BUFFER_SIZE) };
```
ionut-arm, line 144: I don't think this is correct - the number of marshalled bytes might be smaller than this and the exact size is given in `offset`. See, for example, the implementation [here](https://github.com/parallaxsecond/rust-tss-esapi/blob/main/tss-esapi/src/abstraction/transient/key_attestation.rs#L91). That `BUFFER_SIZE` is just the max possible value. The [spec](https://trustedcomputinggroup.org/wp-content/uploads/TCG_TSS_Marshaling_Unmarshaling_API_v1p0_r07_pub.pdf) says:

> The `offset` parameter will be updated by the function to the index of the next byte after the marshaled data 
in the `buffer` byte array.

Pointing this out because in the case I link to above I need the buffer to be truncated correctly, otherwise hashing it doesn't result in the correct object name. Other structures might have similar requirements.

tss-esapi/src/traits.rs
```diff
@@ -0,0 +1,18 @@
...
+    /// Returns the type in the form of marshalled data
+    fn marshall(&self) -> Result<Vec<u8>>;
+}
+
+/// Trait for types that can be created from
+/// TPM marshalled data.
+pub trait UnMarshall: Sized {
+    /// Creates the type from marshalled data.
+    fn un_marshall(marshalled_data: &[u8]) -> Result<Self>;
+}
```
ionut-arm, line 18: Nice! These will come in handy for future implementations

tss-esapi/src/traits.rs
```diff
@@ -0,0 +1,18 @@
...
+    const BUFFER_SIZE: usize;
+    /// Returns the type in the form of marshalled data
+    fn marshall(&self) -> Result<Vec<u8>>;
+}
+
+/// Trait for types that can be created from
+/// TPM marshalled data.
+pub trait UnMarshall: Sized {
+    /// Creates the type from marshalled data.
+    fn un_marshall(marshalled_data: &[u8]) -> Result<Self>;
```
ionut-arm, line 17: I think this can just be `fn unmarshall`, similar to how the library functions are named

tss-esapi/src/structures/attest.rs
Superhepper: Yes, you are right I was going to use it like that but for some reason forgot to use the offset parameter.

tss-esapi/src/traits.rs
wiktor-k: What's the benefit over just an `impl TryFrom<&[u8]>`?

wiktor-k (commented):
Quite a nice collection of types.
They are currently not used by any context methods? Just curious how would they fit in the larger scheme of things here...

tss-esapi/src/traits.rs
Superhepper: fixed

ionut-arm: > What's the benefit over just an impl TryFrom<&[u8]>?

I'd say this is a clearer representation of the fact that you're (un)marshalling structures with some functions intended for that purpose. For some types of structures (`TPM2B_...`) it wouldn't be clear at all that the input/output is the marshalled version and not the contents. For others it could, maybe, be more easily understood, but I think it's nice to make it explicit and consistent.

Superhepper:
From a pure functionality point of view there is little difference. But the new traits are more explicit, the data they produce and consumes are not just any bytes it is specifically the bytes that are produced or consumed by the Tss2_MU_* functions for each individual type.

wiktor-k:
Good point. Thanks for the explanation @Superhepper and @ionut-arm 👍!

Superhepper:
The types will be used later by context method implementation of ```TPM2_Certify``` for example. I actually took some code from the @ionut-arm Certify PR and made the types more 'complete'.

Superhepper:
> 🤯 That's a lot of new types, thanks!! Looks pretty good overall, had just a few comments/questions.
> 
> I'm also curious, I thought there was at least one place where one of the FFI types is used, `TPM2B_ATTEST`, I think. Did you not want to update that yet?

I was so focused on the Attest type and all the info types that I did not even think of to check if the TPM2B_ATTEST or TPMS_ATTEST was used any where. But I fixed that now.

ionut-arm (approved):
👍🏻

Superhepper:
@wiktor-k: Is there something you would like to be changed before this can be merged?

wiktor-k:
I've re-read it all and the only place I'm less-than-super-positive is `unsafe { buffer.set_len(offset as usize) };`. I'm wondering if truncate is not sufficient in this case? Maybe I'm paranoid but I wouldn't want the vector to contain some random memory in case offset is bigger than the initial capacity.

If you think it's good as it is I won't block it though :)

Superhepper:
> I've re-read it all and the only place I'm less-than-super-positive is `unsafe { buffer.set_len(offset as usize) };`. I'm wondering if truncate is not sufficient in this case? Maybe I'm paranoid but I wouldn't want the vector to contain some random memory in case offset is bigger than the initial capacity.
> 
> If you think it's good as it is I won't block it though :)

Well truncate wont work here because the vector can in this case not know the number of bytes that have been written to it. 

What can be done is to allocate an array with a maximum size and write the data to it and then convert a slice of that array into an vector. It would require some extra copying. But it might be worth it in order to avoid the unsafe part.

Superhepper:
I actually come up with a way to use truncate instead. Will change it in one sec.

wiktor-k:
Ha, very cool! One less unsafe thing to worry about. Thanks for taking the extra steps for safety @Superhepper. 👍

PR #294, opened 2021-11-21
Removes TPMA_LOCALITY from context methods.

Superhepper:
- Adds the LocalityAttributes type, this corresponds to TPMA_LOCALITY.
- Changes the context methods to use LocalityAttributes instead of TPMA_LOCALITY.
- Changes the CreationData type to use LocalityAttributes internally instead of TPMA_LOCALITY.

### DIFF
--- a/tss-esapi/src/attributes/locality.rs
+++ b/tss-esapi/src/attributes/locality.rs
@@ -97,7 +97,7 @@ impl LocalityAttributesBuilder {
         let mut locality_attributes = LocalityAttributes(0);
         for locality in self.localities {
             if locality_attributes.is_extended() {
-                print!("Failed to add locality {} to locality attributes because a value that requires extended has been added", locality);
+                error!("Locality attribute {new} and locality attribute {prev} cannot be combined because locality attribute {prev} is extended", new=locality, prev=locality_attributes.0);
                 return Err(Error::local_error(WrapperErrorKind::InvalidParam));
             }
             match locality {
@@ -108,14 +108,14 @@ impl LocalityAttributesBuilder {
                 4 => locality_attributes.set_locality_four(true),
                 5..=31 => {
                     error!(
-                        "{} is an invalid locality and cannot be added to locality attributes",
-                        locality
+                        "Locality attribute {new} is invalid and cannot be combined with other locality attributes",
+                        new=locality
                     );
                     return Err(Error::local_error(WrapperErrorKind::InvalidParam));
                 }
                 32.. => {
                     if locality_attributes.0 != 0 {
-                        error!("locality attribute {} requires extended locality attributes and cannot not be combined with other locality attributes", locality);
+                        error!("Locality attribute {new} is extended and cannot be combined with locality attribute(s) {old}", new=locality, old=locality_attributes.0);
                         return Err(Error::local_error(WrapperErrorKind::InvalidParam));
                     }
                     locality_attributes.0 = locality;
--- a/tss-esapi/src/attributes/locality.rs
+++ b/tss-esapi/src/attributes/locality.rs
@@ -1,0 +1,127 @@
+// Copyright 2021 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{tss2_esys::TPMA_LOCALITY, Error, Result, WrapperErrorKind};
+use bitfield::bitfield;
+use log::error;
+
+bitfield! {
+    /// Bitfield representing the locality attributes.
+    #[derive(Copy, Clone, Eq, PartialEq)]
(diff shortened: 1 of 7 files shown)


tss-esapi/tests/integration_tests/attributes_tests/locality_attributes_tests.rs
```diff
@@ -0,0 +1,192 @@
...
+        "Locality FOUR was not properly set"
+    );
+    assert!(
+        !locality_attributes.is_extended(),
+        "Locality attributes unexpectedly indicated extended"
+    );
+}
+
+#[test]
+fn test_builder_valid_extended() {
```
ionut-arm, line 133: Could you add a test for invalid "extended"?

ionut-arm (commented):
👍🏻 Nice, thanks! Couple of comments

tss-esapi/src/attributes/locality.rs
```diff
@@ -0,0 +1,127 @@
...
+                5..=31 => {
+                    error!(
+                        "{} is an invalid locality and cannot be added to locality attributes",
+                        locality
+                    );
+                    return Err(Error::local_error(WrapperErrorKind::InvalidParam));
+                }
+                32.. => {
+                    if locality_attributes.0 != 0 {
+                        error!("locality attribute {} requires extended locality attributes and cannot not be combined with other locality attributes", locality);
```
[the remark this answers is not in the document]
Superhepper: Sort of. But I need to write a better message.

tss-esapi/tests/integration_tests/attributes_tests/locality_attributes_tests.rs
Superhepper: No, perhaps a poorly chosen name for the test function. Because there are no invalid extended values. This is how it work all values from ```33u8``` to ```u8::MAX``` will make one or more of the bits between 5:7 to be set and that will make ```is_extended``` return true.

ionut-arm: Ah, yes, I just meant one where `as_extended` returns an error.

Superhepper: Aha! of course I will do that right away.

tss-esapi/src/attributes/locality.rs
```diff
@@ -0,0 +1,127 @@
...
+                5..=31 => {
+                    error!(
+                        "{} is an invalid locality and cannot be added to locality attributes",
+                        locality
+                    );
+                    return Err(Error::local_error(WrapperErrorKind::InvalidParam));
+                }
+                32.. => {
+                    if locality_attributes.0 != 0 {
+                        error!("Failed to add extended locality {} because locality attributes is not empty.", locality);
```
wiktor-k, line 118: Typo: "because locality attributes _are_ not empty"
:thinking: too bad that it's not possible to check if the correct log message has been produced in unit tests as there are several things that can return `WrapperErrorKind::InvalidParam` here.

tss-esapi/tests/integration_tests/attributes_tests/locality_attributes_tests.rs
```diff
@@ -0,0 +1,192 @@
...
+        let expected_locality_attributes = LocalityAttributesBuilder::new()
+            .with_locality(locality)
+            .build()
+            .expect("Failed to build locality attributes");
+        let tpma_locality: TPMA_LOCALITY = expected_locality_attributes.into();
+        assert_eq!(
+            1u8.checked_shl(locality.into())
+                .expect("Unable to create locality value"),
+            tpma_locality,
+            "Locality did not convert inte expected TPMA_LOCALITY value"
```
wiktor-k, line 22: inte -> into

wiktor-k (approved):
Looks good with a couple of minor issues...

tss-esapi/tests/integration_tests/attributes_tests/locality_attributes_tests.rs
Superhepper: fixed

tss-esapi/tests/integration_tests/attributes_tests/locality_attributes_tests.rs
Superhepper: fixed

tss-esapi/src/attributes/locality.rs
Superhepper: That would have been nice to check.

I am mixing things up. Sometimes the value can contain several attributes and sometimes it cannot. I have rewritten the messages hope they make better sense now.

ionut-arm (approved):
👍🏻

PR #297, opened 2021-11-26
Support Certify context command

rshearman:
Add support for the Certify context command, which is useful for
proving that a TPM posseses an object as part of remote attestation.

Signed-off-by: Rob Shearman <rob@graphiant.com>

I'm happy to wait for #293 and/or #286 to be merged and rebase on top of them.

### DIFF
--- a/tss-esapi/src/context/tpm_commands/attestation_commands.rs
+++ b/tss-esapi/src/context/tpm_commands/attestation_commands.rs
@@ -1,7 +1,7 @@
 // Copyright 2021 Contributors to the Parsec project.
 // SPDX-License-Identifier: Apache-2.0
 use crate::{
-    handles::KeyHandle,
+    handles::{KeyHandle, ObjectHandle},
     structures::{Attest, AttestBuffer, Data, PcrSelectionList, Signature, SignatureScheme},
     tss2_esys::*,
     Context, Error, Result,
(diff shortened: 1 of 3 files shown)


tss-esapi/src/context/tpm_commands/attestation_commands.rs
```diff
@@ -12,7 +12,45 @@ use std::convert::TryFrom;
...
+    ) -> Result<(TPM2B_ATTEST, Signature)> {
+        let mut certify_info = null_mut();
+        let mut signature = null_mut();
+        let ret = unsafe {
+            Esys_Certify(
+                self.mut_context(),
+                object_handle.into(),
+                signing_key_handle.into(),
+                self.required_session_1()?,
+                self.optional_session_2(),
```
ionut-arm, line 34: I think this one is required as well, as both `object_handle` and `signing_key_handle` need separate authorizations.

ionut-arm (commented):
Thanks for this patch as well! I agree, I think it would make sense to wait for the previous PRs to be merged and then you can return the abstracted types from here. We're aiming for a fully-abstracted `Context` interface for the stable 7.0.0 release

tss-esapi/src/context/tpm_commands/attestation_commands.rs
rshearman: It's optional because signing_key_handle may be Null. From the "Trusted Platform Module Library, Part 3: Commands" doc:

`
Type                | Name           | Description
TPMI_ST_COMMAND_TAG | tag            | TPM_ST_SESSIONS
UINT32              | commandSize    |
TPM_CC              | commandCode    | TPM_CC_Certify
TPMI_DH_OBJECT      | @objectHandle  | handle of the object to be certified / Auth Index: 1 / Auth Role: ADMIN
TPMI_DH_OBJECT+     | @signHandle    | handle of the key used to sign the attestation structure / Auth Index: 2 / Auth Role: USER
TPM2B_DATA          | qualifyingData | user provided qualifying data
TPMT_SIG_SCHEME+    | inScheme       | signing scheme to use if the scheme for signHandle is TPM_ALG_NULL
`

Where + is defined to mean:
> A Type decoration – When appended to a value in the Type column of a command, this symbol
indicates that the parameter is allowed to use the “null” value of the data type (see "Conditional
Types" in TPM 2.0 Part 2). The null value is usually TPM_RH_NULL for a handle or
TPM_ALG_NULL for an algorithm selector

ionut-arm: Ah, yes! Also didn't notice that

> If signHandle  is  TPM_RH_NULL,  the  TPMS_ATTEST  structure  is  returned  and  signature  is  a  NULL Signature.

That actually raises a few other questions.

First would be, is it possible at the moment to create a null `KeyHandle`? I can't see a constructor for it, so it's only through a conversion from `ESYS_TR`. Would it make sense to add a static value for null, kinda like [this](https://github.com/parallaxsecond/rust-tss-esapi/blob/main/tss-esapi/src/structures/buffers/public/rsa.rs#L231)?

Second, should we check here whether `signing_key_handle` is null or not before calling either `optional_session_2` or `required_session_2`?

tss-esapi/src/context/tpm_commands/attestation_commands.rs
```diff
@@ -12,7 +12,45 @@ use std::convert::TryFrom;
 use std::ptr::null_mut;
 
 impl Context {
-    // Missing function: Certify
+    /// Prove that an object is loaded in the TPM
+    ///
+    /// # Errors
```
wiktor-k, line 17: Sorry to be "that guy" but could you also add more thorough command description? Something like this: https://github.com/parallaxsecond/rust-tss-esapi/pull/248/files#diff-c58cba8761b8466b464d02417779fc58f2cff4200e47d5937b3a1b9de6890bfaR18-R294

The code example can be similar to the test case that you already have.

tss-esapi/tests/integration_tests/context_tests/tpm_commands/attestation_commands_tests.rs
```diff
@@ -39,4 +41,63 @@ mod test_quote {
...
+                    Some(AuthSession::Password),
+                    Some(AuthSession::Password),
+                    None,
+                ),
+                |ctx| {
+                    ctx.certify(
+                        obj_key_handle.into(),
+                        sign_key_handle,
+                        &Data::try_from(qualifying_data).unwrap(),
+                        scheme,
```
wiktor-k, line 82: Nice test :+1: too bad we don't have a wrapper for `TPMT_SIG_SCHEME`...

tss-esapi/src/context/tpm_commands/attestation_commands.rs
rshearman: Good suggestion. I'll try to add something as good as the example you link to.

tss-esapi/src/context/tpm_commands/attestation_commands.rs
rshearman: I ended up making the both required as originally suggested as although the second handle is optional, upon testing with a software TPM and re-reading the spec, it seems the session is still required even if the handle is NULL.

Superhepper (approved):
Thanks, looks good to me.

PR #298, opened 2021-11-26
SIgnature and Public marshalling/unmarshalling for remote attestation

rshearman:
When performing remote attestation then it is often necessary to transmit certain TPM objects over a network to the remote entity and the Public and Signature are such objects, so being able to marshall/unmarshall them is useful.

This set of changes adds implementations to marshall/unmarshall to support this use case.

### DIFF
--- a/tss-esapi/src/structures/buffers/public.rs
+++ b/tss-esapi/src/structures/buffers/public.rs
@@ -291,7 +291,7 @@ impl PublicBuilder {
 ///
 /// # Details
 /// This corresponds to TPM2B_PUBLIC
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
 pub enum Public {
     Rsa {
         object_attributes: ObjectAttributes,
--- a/tss-esapi/src/structures/buffers/public.rs
+++ b/tss-esapi/src/structures/buffers/public.rs
@@ -8,6 +8,7 @@ use crate::{
     attributes::ObjectAttributes,
     interface_types::algorithm::{HashingAlgorithm, PublicAlgorithm},
     structures::{Digest, EccPoint, PublicKeyRsa, SymmetricCipherParameters},
+    traits::{Marshall, UnMarshall},
     tss2_esys::{TPM2B_PUBLIC, TPMT_PUBLIC},
     Error, Result, WrapperErrorKind,
 };
@@ -18,7 +19,9 @@ use rsa::PublicRsaParameters;
 
 use log::error;
 use std::convert::{TryFrom, TryInto};
-use tss_esapi_sys::{TPMU_PUBLIC_ID, TPMU_PUBLIC_PARMS};
+use tss_esapi_sys::{
+    Tss2_MU_TPMT_PUBLIC_Marshal, Tss2_MU_TPMT_PUBLIC_Unmarshal, TPMU_PUBLIC_ID, TPMU_PUBLIC_PARMS,
+};
 
 /// A builder for the [Public] type.
 #[derive(Debug, Clone)]
@@ -290,7 +293,7 @@ impl PublicBuilder {
 /// Enum representing the Public structure.
 ///
 /// # Details
-/// This corresponds to TPM2B_PUBLIC
+/// This corresponds to TPMT_PUBLIC
 #[derive(Debug, Clone, PartialEq)]
 pub enum Public {
     Rsa {
@@ -375,7 +378,7 @@ impl Public {
     }
 }
 
-impl From<Public> for TPM2B_PUBLIC {
+impl From<Public> for TPMT_PUBLIC {
     fn from(public: Public) -> Self {
         match public {
             Public::Rsa {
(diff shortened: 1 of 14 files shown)


ionut-arm:
Hi, thanks for the patch! 

I was going to recommend waiting for [this PR](https://github.com/parallaxsecond/rust-tss-esapi/pull/293) and implementing this conversion using [the new traits](https://github.com/parallaxsecond/rust-tss-esapi/pull/293/files#diff-1eea7e9678016ce59d904a65994a75b4eb065c2819b73d7922badbd9a715b646R7), however I don't think this is particularly helpful or desirable in this situation because `Public` is theoretically an abstracted representation of `TPM2B_PUBLIC`, not of `TPMT_PUBLIC`. So I'm wondering how to handle this - maybe adding two pairs of methods on `Public`, `marshall_tpm2b`/`unmarshall_tpm2b` and `marshall_tpmt`/`unmarshall_tpmt`?

Thoughts?

cc @Superhepper @wiktor-k

Superhepper:
I wonder if it would be a good idea to use serde? And make implementation for how to serialize our types using that?

rshearman:
> I wonder if it would be a good idea to use [Serde](https://github.com/serde-rs/serde) And make implementation for how to serialize/deserialize our types using that?

I don't believe using serde is a good fit for this as it assumes a field-by-field serialisation mechanism which doesn't work with unions with external discriminants, such as are present in many structures, and it wouldn't handle any custom serialisation of types above primitive types. Plus we already have the marshal/unmarshal implementations available in the tpm2-tss library, so using serde, even if it was a good fit for the serialisation format, wouldn't be of much benefit.

> I was going to recommend waiting for this PR and implementing this conversion using the new traits, however I don't think this is particularly helpful or desirable in this situation because Public is theoretically an abstracted representation of TPM2B_PUBLIC, not of TPMT_PUBLIC. So I'm wondering how to handle this - maybe adding two pairs of methods on Public, marshall_tpm2b/unmarshall_tpm2b and marshall_tpmt/unmarshall_tpmt?

Good suggestion on the traits - I wasn't aware of those new traits being proposed, but I'm happy to reimplement at least the Signature conversion using them. Regarding the TPM2B_PUBLIC versus TPMT_PUBLIC - that's an interesting point - my thoughts were that I didn't consider TPM2B_PUBLIC to be a useful marshalled format due to redundancy in data, but I can see how it could be useful in some cases. We could rename the `Public` type to `PublicBuffer`, analogous to the `AttestBuffer` type being introduced one of @Superhepper's PRs, and then introduce a new `Public` type with conversions to/from `PublicBuffer`, but I'm not sure if the disruption is worth it.

wiktor-k:
> So I'm wondering how to handle this - maybe adding two pairs of methods on Public, marshall_tpm2b/unmarshall_tpm2b and marshall_tpmt/unmarshall_tpmt?

If this will be common (e.g. extended beyond Public) maybe even create a newtype wrapper over `Vec<u8>` so that it's clear what form are we dealing with (of course with a way to get the inner bytes via `into_inner` or Deref or AsRef...)

> We could rename the Public type to PublicBuffer, analogous to the AttestBuffer type being introduced one of @Superhepper's PRs, and then introduce a new Public type with conversions to/from PublicBuffer, but I'm not sure if the disruption is worth it.

I wouldn't be against the Public/PublicBuffer proposal if this makes it clear what's the backing TPM structure (Sorry I'm using this crate as safe abstractions over raw TPM data but I don't want it to be "too" smart).

ionut-arm:
> We could rename the Public type to PublicBuffer, analogous to the AttestBuffer type being introduced one of @Superhepper's PRs, and then introduce a new Public type with conversions to/from PublicBuffer, but I'm not sure if the disruption is worth it.

Hmm, I'd actually make the case the other way around, i.e. we should instead say that the current `Public` type represents `TPMT_PUBLIC`. After all, `TPM2B_PUBLIC` is nothing more than `TPMT_PUBLIC` + it's size, so all the important bits are actually stored by the `TPMT` type. For the buffer type we could just implement another type either wrapping this one or just wrapping a `TPM2B_PUBLIC` directly.

Another thing that I've been thinking of is that we probably don't use that `size` field in `TPM2B_PUBLIC` correctly, we [always use the size of the `TPMT_PUBLIC` struct](https://github.com/parallaxsecond/rust-tss-esapi/blob/main/tss-esapi/src/structures/buffers/public.rs#L388), but I'm wondering if instead that should be the size of the marshalled version of `TPMT_PUBLIC`.

Superhepper:
I think I agree with @ionut-arm. The best thing to do would be to create PublicBuffer and make that convert into TPM2B_PUBLIC, and then make the Public structure convert into TPMT_PUBLIC. And behave like Attest and AttestBuffer does.
Then one can add some clever read and write implementations for the Buffer types so that they can easily be written to file or to socket.

ionut-arm:
@rshearman - would you like to make those changes around `Public`? I'd be happy to put up a PR for it.

Oh, and something else that we didn't discuss is whether we should take in (and/or spit out) `Public` or `PublicBuffer` in methods like `create` and `create_primary`. I'm ok with just `Public`, even though the "technically" correct one is `PublicBuffer`. But I think `Public` is generally more useful, and the conversions shouldn't be too hard if someone wants a `TPM2B_PUBLIC`.

rshearman:
@ionut-arm, if you're happy to put up a PR for the changes around `Public` then I'm happy to let you!

ionut-arm:
I should've checked what your PR implements before doing #302, I essentially duplicated a whole bunch of stuff 🙃 Will review this again to see what to merge from where.

ionut-arm:
@rshearman - would you be ok with me copying the `Signature`-related bits from your PR into a commit in #302 and crediting you as the author? Might add some more tests as well.

rshearman:
> @rshearman - would you be ok with me copying the `Signature`-related bits from your PR into a commit in #302 and crediting you as the author? Might add some more tests as well.

Absolutely, feel free to slice and dice as you see fit.

ionut-arm:
Excellent. I've just pushed a commit for that. Should I close this PR? Still need to fix some stuff in my branch... :-]

rshearman:
Thank you. Closing since this is superseded by #302.

PR #300, opened 2021-11-30
Command code list

Superhepper:
This adds the CommandCodeList type. It was supposed to be part of #299 but got left out.

### DIFF
--- a/tss-esapi/src/structures/lists/command_code.rs
+++ b/tss-esapi/src/structures/lists/command_code.rs
@@ -7,7 +7,7 @@ use crate::{
     Error, Result, WrapperErrorKind,
 };
 use log::error;
-use std::convert::TryFrom;
+use std::{convert::TryFrom, ops::Deref};
 
 /// A list of command codes.
 #[derive(Debug, Clone, Default)]
@@ -92,5 +92,13 @@ impl From<CommandCodeList> for Vec<CommandCode> {
 impl AsRef<[CommandCode]> for CommandCodeList {
     fn as_ref(&self) -> &[CommandCode] {
         self.command_codes.as_slice()
+    }
+}
+
+impl Deref for CommandCodeList {
+    type Target = Vec<CommandCode>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.command_codes
     }
 }
--- a/tss-esapi/src/structures/lists/command_code.rs
+++ b/tss-esapi/src/structures/lists/command_code.rs
@@ -1,0 +1,96 @@
+// Copyright 2021 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{
+    constants::CommandCode,
+    tss2_esys::{TPM2_MAX_CAP_CC, TPML_CC},
+    Error, Result, WrapperErrorKind,
+};
+use log::error;
+use std::convert::TryFrom;
+
+/// A list of command codes.
+#[derive(Debug, Clone, Default)]
+pub struct CommandCodeList {
+    command_codes: Vec<CommandCode>,
+}
+
+impl CommandCodeList {
+    pub const MAX_SIZE: usize = TPM2_MAX_CAP_CC as usize;
+    /// Creates a new CommandCodeList
+    pub const fn new() -> Self {
+        CommandCodeList {
+            command_codes: Vec::new(),
+        }
+    }
+
+    /// Adds a command code to the command code list.
+    pub fn add(&mut self, command_code: CommandCode) -> Result<()> {
+        if self.command_codes.len() + 1 > CommandCodeList::MAX_SIZE {
+            error!(
+                "Adding command code to list will make the list exceeded its maximum count(> {})",
+                CommandCodeList::MAX_SIZE
+            );
+            return Err(Error::local_error(WrapperErrorKind::WrongParamSize));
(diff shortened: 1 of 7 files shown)


tss-esapi/src/structures/lists/command_code.rs
```diff
@@ -0,0 +1,82 @@
...
+    pub const MAX_SIZE: usize = TPM2_MAX_CAP_CC as usize;
+    /// Creates a new CommandCodeList
+    pub const fn new() -> Self {
+        CommandCodeList {
+            command_codes: Vec::new(),
+        }
+    }
+
+    /// Returns a command codes in the list as a slice.
+    pub fn value(&self) -> &[CommandCode] {
```
wiktor-k, line 28: Just wondering if we shouldn't also implement `AsRef<[CommandCode]>` ([`AsRef` docs](https://doc.rust-lang.org/std/convert/trait.AsRef.html)). This would make it consistent with all other Rust APIs that expose inner buffers (e.g. String exposes ref of bytes etc).

Thinking on this a little bit longer maybe there should be also a function to convert the CommandCodeList into raw `Vec<CommandCode>` just like [other Rust APIs using `into_inner`](https://users.rust-lang.org/t/explanation-of-into-inner/13872).

On more thing that could be useful is `impl TryFrom<Vec<CommandCode>> for CommandCodeList` that'd just check for the correct length and if that's fine wrap the vector with the CommandCodeList.

Thoughts?

Superhepper: I see no problem with it what so ever. I will start working on it.

Superhepper: fixed

wiktor-k (approved):
Looks good to me. 👍 Thanks for the additions @Superhepper.

tss-esapi/src/structures/lists/command_code.rs
```diff
@@ -0,0 +1,106 @@
...
+
+    /// Returns the number of elements in the list.
+    pub fn len(&self) -> usize {
+        self.command_codes.len()
+    }
+
+    /// Returns true if there are no command codes in the list
+    pub fn is_empty(&self) -> bool {
+        self.command_codes.is_empty()
+    }
```
ionut-arm, line 48: Would this not be covered by the `AsRef` implementation below?

tss-esapi/tests/integration_tests/structures_tests/lists_tests/command_code_list_tests.rs
```diff
@@ -0,0 +1,99 @@
...
+        .zip(converted_command_code_list.as_ref().iter())
+        .for_each(|(expected, actual)| {
+            assert_eq!(
+                expected, actual, "Command code in command code list converted from TPML_CC did not match the expected value"
+            )
+        });
+}
+
+#[test]
+fn test_invalid_conversions() {
```
ionut-arm, line 86: Since they follow different avenues, could you also add checks for conversions from `TPML_CC` and `Vec<CommandCode>`?

ionut-arm (commented):
Looks great, thanks! Couple of minor comments.

tss-esapi/src/structures/lists/command_code.rs
Superhepper: How do you mean?

ionut-arm: Ah, sorry, I confused `AsRef` with `Deref`. That might not actually be a bad idea, actually, or did we agree not to use it for some reason? 🤔

wiktor-k: I guess Ionut means that is_empty can be executed via AsRef (that returns the inner buffer reference). TBH I don't mind this here, it seems as a useful convenience function 🤷‍♂️

tss-esapi/tests/integration_tests/structures_tests/lists_tests/command_code_list_tests.rs
Superhepper: Fixed

tss-esapi/src/structures/lists/command_code.rs
Superhepper: It has received Deref now as well.

wiktor-k: Are the `is_empty` and `len` functions now obsolete since they are Deref'd from the vector?

Superhepper: Yes, they are. I removed them.

wiktor-k:
One more thing, a typo in `Made use of CommnadCodeList type.`

tss-esapi/src/structures/lists/command_code.rs
```diff
@@ -0,0 +1,104 @@
...
+            return Err(Error::local_error(WrapperErrorKind::WrongParamSize));
+        }
+        self.command_codes.push(command_code);
+        Ok(())
+    }
+
+    /// Returns the inner type.
+    pub fn into_inner(self) -> Vec<CommandCode> {
+        self.command_codes
+    }
```
ionut-arm, line 43: Every now and then I wish we had something like this implemented for every type.

ionut-arm (approved):
Looks great, thanks!

Issue #290, opened 2021-11-15
Steps to a stable 7.0.0


ionut-arm:
Since we're currently in an alpha release (yet again) it would be good to know what we'd like to achieve before we can tag a stable release. Our focus right now is to remove all not-abstracted FFI types from the `tss-esapi` interface, replacing them with types that we have more control over.

FFI types still present in the `Context` interface (not going to link to where they appear, a simple search through the repo should be enough for that):
- [x] TPMT_SIG_SCHEME ( https://github.com/parallaxsecond/rust-tss-esapi/pull/286 )
- [x] TPM2B_ATTEST ( ~~will be lifted from https://github.com/parallaxsecond/rust-tss-esapi/pull/221~~ ; #293 covers this  )
- [x] TPMA_LOCALITY ( https://github.com/parallaxsecond/rust-tss-esapi/pull/294 )
- [x] TPM2_CC ( #296 )
- [x] TPM2B_SENSITIVE ( #306 )

Another type that currently uses other FFI types in its interface is `CapabilityData`. The types:
- [x] TPM2_PT (just replacing the use in `CapabilityData` with the existing abstraction, https://github.com/parallaxsecond/rust-tss-esapi/pull/301)
- [x] TPM2_PT_PCR (#305, #308)
- [x] TPMA_ALGORITHM (https://github.com/parallaxsecond/rust-tss-esapi/pull/304)
- [x] TPMA_CC(#315)
- [x] TPML_CCA(#315)
- [x] TPM2_ECC_CURVE ( #312  )

Other bits of work:
- [x] Identify the cause of #285 and fix it
- [x] Finish off the request in #277 ( https://github.com/parallaxsecond/rust-tss-esapi/pull/281 )
- [x] Fix PCR structures issues (#310 #311 #317 )

I was also wondering about changing a few other things, please list any other ideas/concerns below:
- [x] ~~Some key-bits types seem redundant, for example `Sm4KeyBits` can only have one value. I wonder if it wouldn't be better to just remove the type completely and have a "default" implementation that produces the correct FFI values wherever they're needed to replace the current conversions~~
- [x] For structures that can be built using `...Builder` structures, we should either document this or create a method on the type to get a new builder, something like `pub fn builder() -> ...` ( #316 )

ionut-arm:
We should also try to identify the cause of #285  and fix it before the release. Also, #277 should be included (via https://github.com/parallaxsecond/rust-tss-esapi/pull/281 ).

[...]

PR #302, opened 2021-12-07
Implement PublicBuffer

ionut-arm:
This commit adds a new type to serve as a native placeholder for
TPM2B_PUBLIC, with Public being switched to represent TPMT_PUBLIC.

Conversions and Marshalling/UnMarshalling implementations are also added
for Public and PublicBuffer, along with tests for these conversions.

cc @rshearman


Superhepper:
I need to try to read this without constantly thinking "structures needs to be refactored". But from a quick overview I think it looked pretty good.

ionut-arm:
> I need to try to read this without constantly thinking "structures needs to be refactored". But from a quick overview I think it looked pretty good.

Yeah - as long as we don't change what users would see, we can do it in a future release.

Also, just a note: I've implemented the conversions to/from `TPM2B_PUBLIC` on `Public` as well because it's rather annoying otherwise - you have to go through marshalling/unmarshalling if you want to use `PublicBuffer` instead. So annoying to have a "buffer" type that isn't actually a buffer... 😐

tss-esapi/src/context/tpm_commands/object_commands.rs
```diff
@@ -71,7 +71,7 @@ impl Context {
                 self.optional_session_2(),
                 self.optional_session_3(),
                 &sensitive_create,
-                &public.clone().into(),
+                &public.clone().try_into()?,
```
wiktor-k, line 74: This is not directly related to your PR but the `clone()` call... ugh... maybe we should track or decide on whether to take the Public by value (so that the caller has to explicitly clone if they want to reuse public). Because currently it always clones public even if passing by value a single thing would suffice...

tss-esapi/src/structures/buffers/public.rs
```diff
@@ -1,507 +1,168 @@
...
-    /// [KeyedHash][`crate::interface_types::algorithm::PublicAlgorithm::KeyedHash].
-    pub fn with_keyed_hash_unique_identifier(
-        mut self,
-        keyed_hash_unique_identifier: &Digest,
-    ) -> Self {
-        self.keyed_hash_unique_identifier = Some(keyed_hash_unique_identifier.clone());
-        self
+    fn try_from(bytes: &[u8]) -> Result<Self> {
+        if bytes.len() > Self::MAX_SIZE {
+            error!("Error: Invalid &[u8] size(> {})", Self::MAX_SIZE);
```
wiktor-k, line 60: :thinking: why not attach `bytes.len()` to the error message such as:
```rust
error!("Error: Invalid &[u8] size: {} > {}", bytes.len(), Self::MAX_SIZE);
```

tss-esapi/src/structures/buffers/public.rs
```diff
@@ -1,507 +1,168 @@
...
-    /// [Ecc][`crate::interface_types::algorithm::PublicAlgorithm::Ecc].
-    ///
-    /// The unique identifier is a ecc point.
-    pub fn with_ecc_unique_identifier(mut self, ecc_unique_identifier: &EccPoint) -> Self {
-        self.ecc_unique_identifier = Some(ecc_unique_identifier.clone());
-        self
+    fn try_from(tss: TPM2B_PUBLIC) -> Result<Self> {
+        let size = tss.size as usize;
+        if size > Self::MAX_SIZE {
+            error!("Error: Invalid buffer size(> {})", Self::MAX_SIZE);
```
wiktor-k, line 73: Likewise.

tss-esapi/src/context/tpm_commands/object_commands.rs
ionut-arm: So, I didn't notice this before in the Rust API guidelines, but it seems to actually be specified there: [we should ask for ownership instead of cloning from a reference](https://rust-lang.github.io/api-guidelines/flexibility.html#caller-decides-where-to-copy-and-place-data-c-caller-control). Given that this essentially breaks the API, the question is whether we want to ship it in 7.0.0. Thoughts? Should we discuss it under #290 ? cc @Superhepper

Also, I'm not sure if I actually answered the question or if you meant that we could not take ownership at all, since we only need a reference in the end (though the reference is to `TPM2B_PUBLIC`)

ionut-arm: Damn, deleted my comment by accident...

It seems, from the Rust API guidelines, that [we should ask for ownership of the value instead of cloning from a reference](https://rust-lang.github.io/api-guidelines/flexibility.html#caller-decides-where-to-copy-and-place-data-c-caller-control). Since this breaks the API we need to decide whether to do this for 7.0.0 or leave it for a subsequent release. It's annoying, but it's not the end of the world, and we have a lot of changes due in 7.0.0, so I'd prefer sticking with the current API, but can be convinced otherwise. Thoughts? cc @Superhepper 

Also, I don't know if the paragraph above answers your question - were you asking if we really need ownership here at all, since we only need to pass down a reference?

tss-esapi/src/structures/buffers/public.rs
ionut-arm: Fixed

tss-esapi/src/structures/buffers/public.rs
ionut-arm: Fixed

tss-esapi/src/context/tpm_commands/object_commands.rs
wiktor-k: There are two things in play here:
  - first, conversion from `&Public` to `Public` that allows to omit the `clone()` call, that was what I was talking about, I don't think this is a big deal with regards to API breakage (I've been updating my crate continuously to 7-alpha and there were bigger issues but I understand if you'd like to defer that),
  - then I thought: it's actually just doing `try_into` to convert to `PublicBuffer` (checked docs for `TPM2B_PUBLIC`) so a next question raises: why not take `PublicBuffer` parameter instead?

Superhepper: We have already declared that there will be some huge interface changes in 7.0.0. And I think I would prefer that we do all the interface changes that we have been wanted to do now when we have the chance. I am already prepared that moving my code to 7.0.0 will be a huge task. And it would annoy me if I now fixed all my code and then some one makes interfaces changes for lets say 8.0.0 and new functionality that I need are based on those changes.

I am not to found of the method API dealing with PublicBuffer. Because, and correct me if I am wrong, that is not a type that is returned from the context methods. Public is the type we uses because it contains the values in a human readable form, PublicBuffer is something you use if you want the type as bytes.

ionut-arm: > it's actually just doing try_into to convert to `PublicBuffer`

It's not, that's why I implemented `TryFrom<Public> for TPM2B_PUBLIC`, so we don't have to go through `PublicBuffer`, because getting a `PublicBuffer` (or getting something from `PublicBuffer`) _always_ requires either marshalling or unmarshalling, so you'd have two useless operations in-between.

wiktor-k: Oh, great! I must've overlooked that fact. Thanks for the explanation!

wiktor-k (approved):

tss-esapi/src/structures/tagged/signature.rs
```diff
@@ -125,3 +128,59 @@ impl TryFrom<TPMT_SIGNATURE> for Signature {
...
+                    error!("Failed to convert size of buffer to TSS size_t type: {}", e);
+                    Error::local_error(WrapperErrorKind::InvalidParam)
+                })?,
+                &mut offset,
+            )
+        };
+        let ret = Error::from_tss_rc(ret);
+        if ret.is_success() {
+            unsafe {
+                buffer.set_len(offset as usize);
```
Superhepper, line 154: This should be changed to the vector version that does not ned to use 'unsafe'.

Superhepper (commented):
Looks good to me. Only had one remark. Don't know if that was in code written by you or something that have been moved but it would be nice to get it fixed.

tss-esapi/src/structures/tagged/signature.rs
ionut-arm: Nice, thanks for pointing that out!

Superhepper (approved):

PR #303, opened 2021-12-08
Fix typos and introduce Check spelling CI step

wiktor-k:
Note that I've never worked with GitHub Actions so the workflow needs additional eye check (I'm not sure if codespell is installed by default on "ubuntu-latest")

Most of these typos were fixed with the `--write-changes` and I needed to fix only the 5 that were ambiguous.


wiktor-k:
Oh, wait. I found https://github.com/codespell-project/actions-codespell :) re-adjusting :)

tss-esapi/src/attributes/nv_index.rs
```diff
@@ -92,7 +92,7 @@ impl NvIndexAttributes {
         // TPMA_NV_AUTHWRITE, or TPMA_NV_POLICYWRITE shall be SET."
         if !(self.pp_write() | self.owner_write() | self.auth_write() | self.policy_write()) {
             error!(
-                "Non of the attributes PPWRITE, OWERWRITE, AUTHWRITE, POLICYWRITE have been set"
+                "Non of the attributes PPWRITE, OVERWRITE, AUTHWRITE, POLICYWRITE have been set"
             );
```
Superhepper, line 96: I do not think this is correct. I think it should be OWNERWRITE

tss-esapi/src/handles/tpm.rs
```diff
@@ -114,7 +114,7 @@ macro_rules! create_tpm_handle_type {
             pub fn new(value: u32) -> Result<$handle_type_name> {
                 if value.to_be_bytes()[0] != $tpm_handle_type_id {
                     error!(
-                        "Errro: TPM Handle ID of the input value did not match the {} (!={})",
+                        "Error: TPM Handle ID of the input value did not match the {} (!={})",
                         stringify!($handle_type_name),
```
Superhepper, line 118: I know it is not really part of the ticket. But there is no need to write Error in a lock::error it will just result in something looking like ```rust-esapi-tss ERROR - Error ...." ``` so remove the error completely in the message.

Superhepper (changes requested):
Thanks a lot!
I think like 99% of all those spelling errors are from me. I only had two comments for you.

tss-esapi/src/handles/tpm.rs
wiktor-k: Yep. Good point :+1:

tss-esapi/src/attributes/nv_index.rs
wiktor-k: :facepalm: yep, that's right. Adjustin'...

tss-esapi/src/handles/tpm.rs
wiktor-k: There are quite a few of these in the codebase. I'll fix them all at once.

wiktor-k:
> I think like 99% of all those spelling errors are from me. I only had two comments for you. But they were really fun to read when pointed out to me.

Haha, no worries. I actually recalled that the Sequoia CI uses codespell so I just figured out we can transplant it here for more automatic reviews :)

tss-esapi/src/context/tpm_commands/context_management.rs
```diff
@@ -163,23 +163,23 @@ impl Context {
     ///            specified in [Provision].
     /// * `object_handle` - The handle of a loaded object.
     /// * `persistant` - If the `object_handle` is transient object then this
```
Superhepper, line 165: Is the parameter in the code spelled incorrectly as well?

wiktor-k: :thinking: yeah it seems so... it seems codespell ignores things between ` and ` (considers them as raw names perhaps)... I'll fix that one manually.

Superhepper: So it only cares about comments without backticks not things that are in the actual code?

wiktor-k: It checks for common misspellings but I guess there is a heuristic that disables the use for the actual code or stuff that looks like code (I've removed backticks and it corrected that). The advantage is that the false-positives are rare, but the downside - as we saw - is that it doesn't catch everything :-/

(FTR more details are at: https://github.com/codespell-project/codespell#codespell)

Superhepper (approved):

tss-esapi/src/context/tpm_commands/context_management.rs
```diff
@@ -280,22 +280,22 @@ impl Context {
...
     /// #     .expect("Failed to flush context");
-    /// # // Close the persistant_handle returned by evict_control
+    /// # // Close the persistent_handle returned by evict_control
     /// # context
     /// #     .tr_close(&mut persistent_object_handle)
-    /// #     .expect("Failed to close persistant handle");
+    /// #     .expect("Failed to close persistent handle");
     /// # // Retrieve the handle from the tpm again.
-    /// # let retireved_persistant_handle = context.execute_without_session(|ctx| {
+    /// # let retireved_persistent_handle = context.execute_without_session(|ctx| {
```
ionut-arm, line 288: Hmm, these should be `retrieved_...`

tss-esapi/tests/integration_tests/context_tests/tpm_commands/context_management_tests.rs
```diff
@@ -264,24 +264,24 @@ mod test_evict_control {
...
-        // Close the persistant_handle returned by evict_control
+        // Close the persistent_handle returned by evict_control
         context
             .tr_close(&mut persistent_primary_key_handle)
-            .expect("Failed to close persistant handle");
+            .expect("Failed to close persistent handle");
 
         // Retrieve the handle from the tpm again.
-        let retireved_persistant_handle = context.execute_without_session(|ctx| {
+        let retireved_persistent_handle = context.execute_without_session(|ctx| {
```
ionut-arm, line 273: Same here.

ionut-arm (approved):
Thanks, looks good, just a couple of nitpicks 😜

wiktor-k:
Thanks for the additional check folks! As we see it's not a bullet-proof solution but I think it will increase the signal/noise ratio of PRs.

I'm merging it now as it touches quite a lot of files and I don't want to deal with more conflicts later :speak_no_evil:

PR #306, opened 2022-01-05
Sensitive buffers

ionut-arm:
This PR is built on top of #302. Once that one is merged I'll mark this one as ready to review.

Adding `Sensitive` and `SensitiveBuffer` (much like `Public` and `PublicBuffer` before). I've also moved some things around in `buffers.rs` just to put them in alphabetical order. 

I'm not entirely sure what to do about the `Vendor` variant in `Sensitive` - theoretically it's there, in the union, but there's no selector for it so we can't really convert, and thus use it.... Thoughts?


Superhepper:
Questions:
Is this based on public buffer PR or on master?

The 'any' fields in unions seems like a hack to me sometimes. I am having a hard time understanding what to do with them sometimes. I am perfectly fine with leaving them right now and add something that could support vendor specific later. In other places in the code we even throw an error when trying to convert something that contains vendor specific data I think.

ionut-arm:
> Is this based on public buffer PR or on master?

Public buffer PR, because I wanted to use some common stuff for testing.

tss-esapi/tests/integration_tests/common/mod.rs
```diff
@@ -47,7 +47,7 @@ pub const HASH: [u8; 64] = [
 ];
 
 #[allow(dead_code)]
-pub const KEY: [u8; 512] = [
+pub const KEY: [u8; 256] = [
```
ionut-arm, line 50: Just a note - I've changed these because using them in the new tests below resulted in the TPM getting angry, because the key itself was not 512 bytes long, and now we're using `KEY` as the full key (via the new buffer implementation, `PrivateKeyRsa`)

tss-esapi/src/structures/buffers.rs
```diff
@@ -185,6 +169,30 @@ pub mod digest {
...
+        TPM2B_ECC_PARAMETER
+    );
+}
+
+pub mod encrypted_secret {
+    named_field_buffer_type!(EncryptedSecret, 256, TPM2B_ENCRYPTED_SECRET, secret);
+}
+
+pub mod id_object {
+    named_field_buffer_type!(IDObject, 256, TPM2B_ID_OBJECT, credential);
```
Superhepper, line 185: Should we change the name of this to ```IdObject```? The camel case naming of types is usually what is used every else.

Superhepper (approved):
Looks good to me. I had only one question. But I am approving this.

tss-esapi/src/structures/buffers.rs
ionut-arm: Done!

wiktor-k (approved):
Looks good :+1:

wiktor-k:
Hopefully this won't collide too much with #307 :)

ionut-arm:
> Hopefully this won't collide too much with #307 :)

I'm ok with handling the collisions :]

PR #307, opened 2022-01-11
Update API to take ownership where needed

ionut-arm:
Updating functions and methods in the API (particularly for `Context`
methods) where we take a reference and then clone the object, instead of
taking ownership directly, letting the user clone whenever needed.

This includes updates to the tests and documentation.

### DIFF
--- a/tss-esapi/src/context/tpm_commands/non_volatile_storage.rs
+++ b/tss-esapi/src/context/tpm_commands/non_volatile_storage.rs
@@ -28,8 +28,8 @@ impl Context {
     pub fn nv_define_space(
         &mut self,
         nv_auth: Provision,
-        auth: Option<&Auth>,
-        public_info: &NvPublic,
+        auth: Option<Auth>,
+        public_info: NvPublic,
     ) -> Result<NvIndexHandle> {
         let mut object_identifier: ESYS_TR = ESYS_TR_NONE;
         let ret = unsafe {
@@ -39,8 +39,8 @@ impl Context {
                 self.optional_session_1(),
                 self.optional_session_2(),
                 self.optional_session_3(),
-                &auth.cloned().unwrap_or_default().into(),
-                &public_info.clone().try_into()?,
+                &auth.unwrap_or_default().into(),
+                &public_info.try_into()?,
                 &mut object_identifier,
             )
         };
@@ -135,7 +135,7 @@ impl Context {
         &mut self,
         auth_handle: NvAuth,
         nv_index_handle: NvIndexHandle,
-        data: &MaxNvBuffer,
+        data: MaxNvBuffer,
         offset: u16,
     ) -> Result<()> {
         let ret = unsafe {
@@ -146,7 +146,7 @@ impl Context {
                 self.optional_session_1(),
                 self.optional_session_2(),
                 self.optional_session_3(),
-                &data.clone().into(),
+                &data.into(),
                 offset,
             )
         };
--- a/tss-esapi/src/structures/pcr/select.rs
+++ b/tss-esapi/src/structures/pcr/select.rs
@@ -30,7 +30,7 @@ impl PcrSelect {
     pub fn new(size_of_select: PcrSelectSize, pcr_slots: &[PcrSlot]) -> Self {
         PcrSelect {
             size_of_select,
-            selected_pcrs: pcr_slots.iter().cloned().collect(),
+            selected_pcrs: pcr_slots.iter().copied().collect(),
         }
     }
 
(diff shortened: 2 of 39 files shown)


tss-esapi/src/context/tpm_commands/non_volatile_storage.rs
```diff
@@ -39,8 +39,8 @@ impl Context {
                 self.optional_session_1(),
                 self.optional_session_2(),
                 self.optional_session_3(),
-                &auth.cloned().unwrap_or_default().into(),
```
wiktor-k, line 42: Interesting that this function was called `cloned` with **d** :thinking:

wiktor-k (approved):
Looks great :+1: Thanks for going through the entire code and adjusting :)

tss-esapi/src/context/tpm_commands/non_volatile_storage.rs
Superhepper: I think it is to distinguish from cloning the reference or cloning the referenced data.

wiktor-k: Understood, thanks for context. Since now this has been removed maybe that function is no longer needed? Just asking as I never used it :)

wiktor-k: `cloned`... but after a careful study of other PRs I see it in a few places here and there... I wonder why haven't I seen it earlier :thinking: It's probably all OK now :+1:

ionut-arm:
I've rebased and also found a few more places that were doing extra clones, so changed that as well. No need to constantly re-review unless you're particularly keen. It'll be necessary once the other PRs are out of the way and I've rebased on top of them (I'll mark this as ready for review)

tss-esapi/src/structures/pcr/select.rs
```diff
@@ -27,10 +27,10 @@ pub struct PcrSelect {
 
 impl PcrSelect {
     /// Creates a new PcrSelect
-    pub fn new(size_of_select: PcrSelectSize, pcr_slots: &[PcrSlot]) -> Self {
+    pub fn new(size_of_select: PcrSelectSize, mut pcr_slots: Vec<PcrSlot>) -> Self {
         PcrSelect {
             size_of_select,
-            selected_pcrs: pcr_slots.iter().cloned().collect(),
+            selected_pcrs: pcr_slots.drain(..).collect(),
```
ionut-arm, line 33: I wasn't sure about this one, whether `Vec` is a fair replacemenet or whether I should use something else, maybe something generic?

Superhepper: hrrmm, that is weird it is only a value.. should be no need to clone it.

ionut-arm: Before we used to get `&[PcrSlot]` - so we had to clone the stuff behind the reference.

Superhepper: This might work:
```rust
    /// Creates a new PcrSelect
    pub fn new(size_of_select: PcrSelectSize, pcr_slots: &[PcrSlot]) -> Self {
        PcrSelect {
            size_of_select,
            selected_pcrs: pcr_slots
                .iter()
                .fold(BitFlags::<PcrSlot>::empty(), |mut acc, &v| {
                    acc.insert(v);
                    acc
                }),
        }
    }
```

ionut-arm: True, but that's essentially doing a (potentially more expensive) version of `cloned`, by hand

ionut-arm: Oh, that's literally what's happening under the hood with the `.collect()` implementations: https://docs.rs/enumflags2/0.7.3/src/enumflags2/lib.rs.html#853

Superhepper: Yeah, it is a set under the hood. Would be nice to be able to it without cloning. Either using from_iterator or collect.

Superhepper:
```
    /// Creates a new PcrSelect
    pub fn new(size_of_select: PcrSelectSize, pcr_slots: &[PcrSlot]) -> Self {
        PcrSelect {
            size_of_select,
            selected_pcrs: BitFlags::<PcrSlot>::from_iter(pcr_slots.iter().copied()),
        }
    }
```

```
    /// Creates a new PcrSelect
    pub fn new(size_of_select: PcrSelectSize, pcr_slots: &[PcrSlot]) -> Self {
        PcrSelect {
            size_of_select,
            selected_pcrs: pcr_slots.iter().copied().collect(),
        }
    }
```

Both works and are better then the previous.

Superhepper (approved):
Nice thanks for this.

ionut-arm:
@Superhepper - should I wait for your other PRs to merge this?

PR #312, opened 2022-01-17
Replace type in CapabilityData::EccCurves

ionut-arm:
Replacing the FFI type with its abstraction for the contents of ECC
Curve capabilities.
The CapabilityData tests are also enhanced. A few of the variants are
renamed to follow the proper capitalization rules.

### DIFF
--- a/tss-esapi/src/constants/ecc.rs
+++ b/tss-esapi/src/constants/ecc.rs
@@ -14,6 +14,9 @@ use num_traits::{FromPrimitive, ToPrimitive};
 use std::convert::TryFrom;
 /// Enum that contains the constants for the
 /// implemented elliptic curves.
+///
+/// # Details
+/// This corresponds to `TPM2_ECC_CURVE`
 #[derive(FromPrimitive, ToPrimitive, Debug, Copy, Clone, PartialEq, Eq)]
 #[repr(u16)]
 pub enum EccCurveIdentifier {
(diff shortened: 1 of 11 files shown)


Superhepper:
Do you think it would make any sense to wrap this up in a list type and then maybe add a 'contains' function in order to make it easier to determine if a certain curve is supported?

ionut-arm:
> Do you think it would make any sense to wrap this up in a list type and then maybe add a 'contains' function in order to make it easier to determine if a certain curve is supported?

Hmm, `Vec` already has a [`contains`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.contains) method that does exactly that, and I think `EccCurveIdentifier` has the right traits (just `PartialEq`, actually)

ionut-arm:
Oh, actually, given that those are just enum variants, I think the best option would actually be `HashSet`, to ensure no duplicates exist. Will change that.

Superhepper:
The reason I like wrapping it up is so that we can if needed change the underlaying container without breaking the api.

ionut-arm:
> The reason I like wrapping it up is so that we can if needed change the underlaying container without breaking the api.

Ok, done, I've added an `EccCurveList` type. I noticed the `Handles` variant in `CapabilityData` had the same issue, so I changed that one too. I've not actually added a `contains` method, since that one is inherited from the `Vec`, but we can implement one in the future if we need to change it.

Superhepper (approved):
Nice! This will make things nice and consistent.

wiktor-k (approved):
This looks very nice, thanks! :+1:

PR #314, opened 2022-01-19
Make the crate compatible with 1.53 toolchain

ionut-arm:
While building Parsec with the latest version of the crate (from
`main`), the legacy build failed due to range syntax that was not
supported in 1.53. This commit modifies the range to align with the
requirements in Parsec.

(failed Parsec build [here](https://github.com/parallaxsecond/parsec/runs/4869272668?check_suite_focus=true))

### DIFF
--- a/tss-esapi/src/attributes/locality.rs
+++ b/tss-esapi/src/attributes/locality.rs
@@ -113,7 +113,7 @@ impl LocalityAttributesBuilder {
                     );
                     return Err(Error::local_error(WrapperErrorKind::InvalidParam));
                 }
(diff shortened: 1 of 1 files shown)


wiktor-k (approved):
Looks okay to me 👍 I guess this could also use `_` (catch all)?

ionut-arm:
> I guess this could also use `_` (catch all)?

I was thinking of that, but explicitly specifying the values seems more useful, as you don't need to know the type of `locality` when reading the code.

Superhepper (approved):

PR #315, opened 2022-01-19
Creates native type for TPML_CCA

Superhepper:
This creates the native type for TPML_CCA, cleans up CapabilityData and refactor the names and location of some files.


tss-esapi/tests/integration_tests/attributes_tests/command_code_attributes_tests.rs
```diff
@@ -0,0 +1,220 @@
...
+        expected_tpma_cc,
+        command_code_attributes.into(),
+        "CommandCodeAttributes did not convert into the expected TPMA_CC value"
+    );
+}
+
+#[test]
+fn test_invalid_conversions_non_vendor_specific_invalid_command_index() {
+    let invalid_tpma_cc = create_tpma_cc(
+        0xFFFFu16, // Invalid command code index,
```
ionut-arm, line 190: I don't get it, is there something invalid about the index? Seems to be the same one as the previous test.

ionut-arm (approved):
Excellent, thank you! A couple of minor comments

tss-esapi/src/structures/capability_data.rs
```diff
@@ -0,0 +1,146 @@
...
+    PpCommands(CommandCodeList),
+    AuditCommands(CommandCodeList),
+    AssignedPcr(PcrSelectionList),
+    TpmProperties(TaggedTpmPropertyList),
+    PcrProperties(TaggedPcrPropertyList),
+    EccCurves(EccCurveList),
+    // These are in the TPM TPMU_CAPABILITIES, but are not defined by esapi-2.4.1
+    // AuthPolicies(),
+    // ActData(),
+}
```
wiktor-k, line 35: :thinking: Hmm... I wonder if we need [`#[non_exhaustive]` on the enum](https://turreta.com/2019/12/20/using-non_exhaustive-for-non-exhaustive-enums/) if this could be further extended with AuthPolicies and ActData... Adding more variants later would be a breaking API change without this attribute I think.

tss-esapi/tests/integration_tests/attributes_tests/command_code_attributes_tests.rs
wiktor-k: I has `false` as a last argument here indicating that this should *not* be a vendor specific (compare last arguments). I agree the long list of true+falses is kind of confusing :)

Maybe moving parameters to explicit variables like in the last test would help?

wiktor-k (approved):
Looks good in general :+1:

tss-esapi/src/structures/capability_data.rs
ionut-arm: Good spot!

tss-esapi/tests/integration_tests/attributes_tests/command_code_attributes_tests.rs
ionut-arm: Nah, I think what confused me is that the comment seems to refer to the index, even though it's not just the index that's invalid, but the index+another parameter.

wiktor-k: Ack. A potential for some improvement here :)

Superhepper: Well the command index ```0xFFFFu16``` is invalid if it is not vendor specific. because if it is vendor specific it can be any u16 number.

tss-esapi/src/structures/capability_data.rs
Superhepper: I wrote a patch for this to the tpm2-tss. So it will be available in later versions I think.

tss-esapi/tests/integration_tests/attributes_tests/command_code_attributes_tests.rs
ionut-arm: True, but that's what I meant, that the index is not inherently invalid, that depends on whether the CC is vendor-specific (so it should be reflected in that comment)

Superhepper: Yeah, that is a good point. It was reflected in the function name but that got so long that it got hard to read. I will fix it and make it clear.

Superhepper:
I was actually thinking of using the bitfield crate in the test(do not know if that is possible) in order to make it more clear what you do instead of using the ```create_tpma_cc``` function.

tss-esapi/src/structures/capability_data.rs
Superhepper: But yeah I think we need it.

Superhepper:
Actually I think I will add a builder. In case you want to create specific attributes for comparison.

tss-esapi/src/structures/capability_data.rs
Superhepper: Fixed

tss-esapi/tests/integration_tests/attributes_tests/command_code_attributes_tests.rs
Superhepper: Fixed

tss-esapi/src/structures/lists/command_code_attributes.rs
```diff
@@ -0,0 +1,108 @@
+// Copyright 2021 Contributors to the Parsec project.
```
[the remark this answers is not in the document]
Superhepper: Fixed

ionut-arm (approved):
🚀

PR #316, opened 2022-01-21
Add builder methods and move NvPublic

ionut-arm:
This commit brings two changes:
* implementing `builder` methods on all types that have their own
`...Builder` type, so that users are more easily directed to them in the
documentation.
* moving the `nv` top-level module within the `structures` module to
make things a bit more consistent.


tss-esapi/src/attributes/nv_index.rs
```diff
@@ -107,6 +107,11 @@ impl NvIndexAttributes {
 
         Ok(())
     }
+
+    /// Get a builder for the structure
+    pub fn builder() -> NvIndexAttributesBuilder {
```
Superhepper, line 112: Should be made const

Superhepper (commented):
You should make the builder functions const were it is possible.

ionut-arm:
> You should make the builder functions const were it is possible.

I've now made all the `builder` methods that I could constant.

Superhepper (approved):
Nice!

wiktor-k (approved):
Looks straightforward :+1:

PR #323, opened 2022-02-07
Allow customizing AK public

### LINKED ISSUE #322: 7.0.0-beta.1: AK template can no longer be customised
Post commit a4dbdc797c1de994b5ef901e9b25c299a3a6e605, the create_ak API no longer supports customising the key beyond its object attributes. This is due to the following change:

```
-    };
-
-    let key_builder = if let Some(ref k) = key_customization {
-        k.template(key_builder)
-    } else {
-        key_builder
-    };
-
-    key_builder.build()
+    }
```

I don't know if this was intentional or not, but if it is then it's probably worth documenting that the key customisation template method won't be called.

Cc: @Superhepper

ionut-arm:
Following feedback from #322, this reintroduces customization for AK
public objects in the abstraction layer.
`Hash` is also added to the derive list of `EccCurveIdentifier`.

Fixes #322 
cc @rshearman

### DIFF
--- a/tss-esapi/Cargo.toml
+++ b/tss-esapi/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "tss-esapi"
-version = "7.0.0-beta.1"
+version = "7.0.0-beta.2"
 authors = ["Parsec Project Contributors"]
 edition = "2018"
 description = "Rust-native wrapper around TSS 2.0 Enhanced System API"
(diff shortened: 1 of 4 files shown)


Superhepper (approved):

anta5010 (approved):

paulhowardarm (approved):
LGTM!

PR #327, opened 2022-02-15
Check for documentation link errors

wiktor-k:
This change adds an attribute and a CI job that will fail if there are
issues with doc links in the crate.

See: https://github.com/parallaxsecond/rust-tss-esapi/pull/326
Signed-off-by: Wiktor Kwapisiewicz <wiktor@metacode.biz>


wiktor-k:
:thinking: hmm... building docs requires native libraries... I'll rework it :)

wiktor-k:
Okay, I've attached it to one of the docker-using jobs and verified that the command fails before https://github.com/parallaxsecond/rust-tss-esapi/pull/326.

This is ready for review & merge now. Thanks!

ionut-arm (approved):
💯 thanks!

Superhepper (approved):

PR #328, opened 2022-03-05
Adds more pcr structure tests.

Superhepper:
Adds more tests for the different PCR related structures
in order to increase the test coverage.

### DIFF
--- a/tss-esapi/tests/integration_tests/structures_tests/pcr_tests/pcr_select_size_tests.rs
+++ b/tss-esapi/tests/integration_tests/structures_tests/pcr_tests/pcr_select_size_tests.rs
@@ -4,6 +4,13 @@
 use std::convert::TryFrom;
 use tss_esapi::{structures::PcrSelectSize, Error, WrapperErrorKind};
 
+fn bad_u8_values() -> Vec<u8> {
+    let mut bad_values = Vec::<u8>::with_capacity(u8::MAX as usize);
+    bad_values.push(0);
+    bad_values.extend(5u8..=u8::MAX);
+    bad_values
+}
+
 macro_rules! test_valid_conversions {
     (PcrSelectSize::$expected:ident, $value:expr) => {
         let expected_u8 = $value;
@@ -101,11 +108,11 @@ fn test_valid_conversions() {
 
 #[test]
 fn test_invalid_conversions() {
-    for value in 5..=u8::MAX {
+    for value in bad_u8_values() {
         assert_eq!(
             Err(Error::WrapperError(WrapperErrorKind::InvalidParam)),
             PcrSelectSize::try_from(value),
-            "Converting an invalid size_of_select value{} did not result in the expected error",
+            "Converting an invalid size_of_select value {} did not result in the expected error",
             value,
         );
     }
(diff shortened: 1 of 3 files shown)


tss-esapi/tests/integration_tests/structures_tests/pcr_tests/pcr_select_size_tests.rs
```diff
@@ -121,3 +128,39 @@ fn test_default() {
...
     );
 }
+
+#[test]
+fn test_try_parse_u8_with_invalid_values() {
+    for value in bad_u8_values() {
+        assert_eq!(
+            Err(Error::WrapperError(WrapperErrorKind::InvalidParam)),
+            PcrSelectSize::try_parse_u8(value),
+            "try_parse_u8 using an invalid size_of_select value{} did not result in the expected error",
```
wiktor-k, line 138: Hmm... a space between `value` and `{}` is missing?

wiktor-k (approved):

tss-esapi/tests/integration_tests/structures_tests/pcr_tests/pcr_selection_tests.rs
```diff
@@ -69,39 +70,243 @@ fn test_size_of_select_handling() {
...
+        PcrSelectSize::TwoOctets,
+        &[PcrSlot::Slot6],
+    )
+    .expect("Failed to create PcrSelection pcr_selection_2");
+
+    assert_eq!(
+            Err(Error::WrapperError(WrapperErrorKind::InvalidParam)),
+            pcr_selection_1.subtract_exact(&pcr_selection_2),
+            "Subtract exact PcrSelections with different size of select did not produce the expected error",
+        );
```
ionut-arm, line 288: For both this test and the next one - if you change the stuff that's currently causing the error, would it still error? I'm thinking here of two things: would it cause an error because the 2nd selection contains a slot that the first doesn't? And, would it cause an error if you end up with an empty selection at the end?

ionut-arm (commented):
Thanks, looks good overall, just one question about the last few tests

tss-esapi/tests/integration_tests/structures_tests/pcr_tests/pcr_selection_tests.rs
Superhepper: The order in which things are checked is HashingAlgorithm->PcrSelectSize->slots. So I tried to only change one parameter at the time in order to trigger the errors.

Superhepper: Oh I see what you mean. That is not correct.

tss-esapi/tests/integration_tests/structures_tests/pcr_tests/pcr_select_size_tests.rs
Superhepper: Fixed

tss-esapi/tests/integration_tests/structures_tests/pcr_tests/pcr_selection_tests.rs
Superhepper: Fixed

ionut-arm (approved):
👍🏻

PR #331, opened 2022-03-11
Adds more session attributes tests.

Superhepper:
Adds tests for SessionAttributes, SessionAttributesMask and
SessionAttributesBuilder.

### DIFF
--- a/tss-esapi/tests/integration_tests/attributes_tests/session_attributes_tests.rs
+++ b/tss-esapi/tests/integration_tests/attributes_tests/session_attributes_tests.rs
@@ -1,0 +1,179 @@
+// Copyright 2022 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+use tss_esapi::{
+    attributes::{SessionAttributes, SessionAttributesBuilder, SessionAttributesMask},
+    tss2_esys::TPMA_SESSION,
+};
+
+macro_rules! test_valid_conversion {
+    ($tpm_value:expr, $method:ident) => {
+        let tpma_session_attribute: TPMA_SESSION = $tpm_value;
+        let session_attributes = SessionAttributes::from(tpma_session_attribute);
+        assert_eq!(
+            true,
+            session_attributes.$method(),
+            "SessionAttributes converted from TPMA_SESSION = {} did not produce the expected result with regard to {}.", tpma_session_attribute, stringify!($method),
+        );
+        assert_eq!(
+            tpma_session_attribute,
+            session_attributes.into(),
+            "Converting session attributes with {} set did not convert into the expected TPMA_SESSION value", std::stringify!($method),
+        );
+    };
+}
+
+macro_rules! test_valid_mask_conversion {
+    ($tpm_value:expr, $attribute:tt) => {
+        let tpma_session_attribute: TPMA_SESSION = $tpm_value;
+        let session_attributes_mask = SessionAttributesMask::from(tpma_session_attribute);
+        assert_eq!(
+            tpma_session_attribute,
+            session_attributes_mask.into(),
+            "Converting session attributes mask with {} set did not convert into the expected TPMA_SESSION value", $attribute,
+        );
+    };
+}
+
+#[test]
+fn test_valid_session_attributes_conversions() {
+    test_valid_conversion!(
+        1u8.checked_shl(0)
+            .expect("Failed to create 'continue session' value"),
+        continue_session
+    );
+    test_valid_conversion!(
+        1u8.checked_shl(1)
+            .expect("Failed to create 'audit exclusive' value"),
+        audit_exclusive
+    );
+    test_valid_conversion!(
+        1u8.checked_shl(2)
+            .expect("Failed to create 'audit reset' value"),
+        audit_reset
+    );
+    test_valid_conversion!(
+        1u8.checked_shl(5)
+            .expect("Failed to create 'decrypt' value"),
+        decrypt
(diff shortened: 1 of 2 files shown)


tss-esapi/tests/integration_tests/attributes_tests/session_attributes_tests.rs
```diff
@@ -0,0 +1,149 @@
...
+
+macro_rules! test_valid_conversion {
+    ($tpm_value:expr, $method:ident) => {
+        let tpma_session_attribute: TPMA_SESSION = $tpm_value;
+        let session_attributes = SessionAttributes::from(tpma_session_attribute);
+        assert_eq!(
+            true,
+            session_attributes.$method(),
+            "SessionAttributes converted from TPMA_SESSION = {} did not produce the expected result with regard to {}.", tpma_session_attribute, stringify!($method),
+        );
```
ionut-arm, line 16: This can probably be changed to 
```
assert!(
            session_attributes.$method(),
            "SessionAttributes converted from TPMA_SESSION = {} did not produce the expected result with regard to {}.", tpma_session_attribute, stringify!($method),
        );
```

tss-esapi/tests/integration_tests/attributes_tests/session_attributes_tests.rs
```diff
@@ -0,0 +1,149 @@
...
+
+    assert_eq!(
+        expected_session_attributes, actual_session_attributes,
+        "SessionAttributes builder did not produce the expected session attributes value"
+    );
+    assert_eq!(
+        expected_session_attributes_mask, actual_session_attributes_mask,
+        "SessionAttributes builder did not produce the expected session attributes mask value"
+    )
+}
```
ionut-arm, line 149: Could you add one like this, but with `0u8` as the value and default builder output?

ionut-arm (commented):
Thanks for the tests 🙏🏻

tss-esapi/tests/integration_tests/attributes_tests/session_attributes_tests.rs
Superhepper: Done

tss-esapi/tests/integration_tests/attributes_tests/session_attributes_tests.rs
```diff
@@ -0,0 +1,149 @@
...
+        encrypt
+    );
+    test_valid_conversion!(
+        1u8.checked_shl(7).expect("Failed to create 'audit' value"),
+        audit
+    );
+}
+
+#[test]
+fn test_invalid_session_attributes_conversions() {
```
wiktor-k, line 71: I wonder if this test can be implemented even if it's failing and then marked with [the `#[ignore]` attribute](https://doc.rust-lang.org/reference/attributes/testing.html#the-ignore-attribute).

tss-esapi/tests/integration_tests/attributes_tests/session_attributes_tests.rs
wiktor-k: Isn't this exactly what the `assert!` macro is doing? I'm reading https://doc.rust-lang.org/std/macro.assert.html and it says `Asserts that a boolean expression is true at runtime.`.

wiktor-k (approved):

tss-esapi/tests/integration_tests/attributes_tests/session_attributes_tests.rs
Superhepper: Yeah, but I am not testing an expression. I am testing a return value that in this case happen to be a bool.

so with assert, you will get a message saying like assertion error looking something like this:
```thread 'main' panicked at 'assertion failed: some_computation()', src/main.rs:9:1```

But with assert_eq you get:
```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `true`,
 right: `false`', src/main.rs:9:1
 ```
 
 To me the latter shows more clearly what I expected and what I got both in the code and in the message.

tss-esapi/tests/integration_tests/attributes_tests/session_attributes_tests.rs
Superhepper: Yes, it should.

tss-esapi/tests/integration_tests/attributes_tests/session_attributes_tests.rs
ionut-arm: I think there are good arguments for both. I like seeing what was expected and what we got, but it would be nice if it actually specified which one is which so we don't have to look at the code, but such is life. Leave it as is!

Superhepper: Yes, it would have been nice if it would have been 'expected' and 'actual' as it is in google test for c++ for example. I wonder if there is a crate that cam creates asserts for testing that are more like that?

tss-esapi/tests/integration_tests/attributes_tests/session_attributes_tests.rs
ionut-arm: Don't think you've pushed this

tss-esapi/tests/integration_tests/attributes_tests/session_attributes_tests.rs
Superhepper: Fixed

ionut-arm (approved):

PR #341, opened 2022-05-06
Add support for converting Public to SubjectPublicKeyInfo

THS-on:
This allows users to easily use TPM generated keys in other libraries.

Some of the ECC curves seem to not have an OID to map to, is returning an error the correct way here?
tpm2-pytss does also not handle them.

I've tested it manually against tpm2-tools using a RSA and ECC key and it works.

### DIFF
--- a/tss-esapi/src/abstraction/public.rs
+++ b/tss-esapi/src/abstraction/public.rs
@@ -1,0 +1,126 @@
+// Copyright 2022 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::interface_types::ecc::EccCurve;
+use crate::structures::{Public, RsaExponent};
+use crate::{Error, WrapperErrorKind};
+
+use core::convert::TryFrom;
+use oid::ObjectIdentifier;
+use picky_asn1::bit_string::BitString;
+use picky_asn1::wrapper::{IntegerAsn1, OctetStringAsn1};
+use picky_asn1_x509::{
+    AlgorithmIdentifier, EcParameters, EcPoint, PublicKey, RsaPublicKey, SubjectPublicKeyInfo,
+};
+use serde::{Deserialize, Serialize};
+
+/// Can be converted from [`crate::structures::Public`] when not a fully constructed
+/// [`picky_asn1_x509::SubjectPublicKeyInfo`] is required.
+///
+/// # Details
+///
+/// Holds either [`picky_asn1_x509::RsaPublicKey`] for [`crate::structures::Public::Rsa`] or
+/// [`picky_asn1_x509::EcPoint`] for [`crate::structures::Public::Ecc`].
+#[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
+pub enum DecodedKey {
+    RsaPublicKey(RsaPublicKey),
+    EcPoint(EcPoint),
+}
+
+impl TryFrom<Public> for DecodedKey {
+    type Error = Error;
+
+    fn try_from(value: Public) -> Result<Self, Self::Error> {
+        public_to_decoded_key(&value)
+    }
+}
+
+impl TryFrom<Public> for SubjectPublicKeyInfo {
+    type Error = Error;
+
+    /// Converts [`crate::structures::Public::Rsa`] and [`crate::structures::Public::Ecc`] to [`picky_asn1_x509::SubjectPublicKeyInfo`].
+    ///
+    /// # Details
+    /// The result can be used to convert TPM public keys to DER using `picky_asn1_der`.
+    ///
+    /// # Errors
+    /// * if other instances of [`crate::structures::Public`] are used `UnsupportedParam` will be returned.
+    fn try_from(value: Public) -> Result<Self, Self::Error> {
+        let decoded_key = public_to_decoded_key(&value)?;
+
+        match (value, decoded_key) {
+            (Public::Rsa { .. }, DecodedKey::RsaPublicKey(key)) => Ok(SubjectPublicKeyInfo {
+                algorithm: AlgorithmIdentifier::new_rsa_encryption(),
+                subject_public_key: PublicKey::Rsa(key.into()),
+            }),
+            (Public::Ecc { parameters, .. }, DecodedKey::EcPoint(point)) => {
+                Ok(SubjectPublicKeyInfo {
+                    algorithm: AlgorithmIdentifier::new_elliptic_curve(EcParameters::NamedCurve(
+                        curve_oid(parameters.ecc_curve())?.into(),
+                    )),
+                    subject_public_key: PublicKey::Ec(BitString::with_bytes(point).into()),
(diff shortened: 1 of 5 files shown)


Superhepper (commented):
Thanks for changes!

I had a few questions. Would it be possible to wrap the &Public in its own type so that the conversions functions can be replaced with the more standard ```TryFrom``` trait implementations?


example:
```
pub struct DecodableKey {
  public: &Public,
};

impl DecodableKey {
  const fn create(public: &Public) -> Self {
    DecodableKey { public }
  }
}

impl TryFrom<..> for DecodableKey {
...
}
```

Or would that be bad idea in your opinion?

THS-on:
@Superhepper that should work. I've have not written that much rust code, so I have no opinion on what the correct way here is.

Superhepper:
> @Superhepper that should work. I've have not written that much rust code, so I have no opinion on what the correct way here is.

It might be a little tricky with declaring life times but not that tricky I think.

ionut-arm:
> so that the conversions functions can be replaced with the more standard `TryFrom` trait implementations?

Hmm, why not directly convert to `DecodedKey` from `Public`?

THS-on:
> Hmm, why not directly convert to DecodedKey from Public?

That is easier:

```rust

impl TryFrom<&Public> for DecodedKey {
    type Error = Error;

    fn try_from(value: &Public) -> Result<Self, Self::Error> {
        public_to_decoded_key(value)
    }
}

impl TryFrom<&Public> for SubjectPublicKeyInfo {
    type Error = Error;

    fn try_from(value: &Public) -> Result<Self, Self::Error> {
        public_to_subject_public_key_info(value)
    }
}
``

**Comment by Superhepper** [comment]
It is not a great idea to implement From or TryFrom on a reference. Because the they are supposed to consume the object they are converting from.

If you implement TryFrom<Public> for DecodedKey then it can no longer take a reference and I thought that was something you wanted to do.

**Comment by ionut-arm** [comment]
> It is not a great idea to implement From or TryFrom on a reference.

Then we can just implement it on `Public` - `Public` has `Clone` derived, should be good enough for anyone using it, if they want to also use the same data for something else!

File: tss-esapi/src/abstraction/public.rs [hunk 1]
```diff
@@ -0,0 +1,116 @@
...
+use oid::ObjectIdentifier;
+use picky_asn1::bit_string::BitString;
+use picky_asn1::wrapper::{IntegerAsn1, OctetStringAsn1};
+use picky_asn1_x509::{
+    AlgorithmIdentifier, EcParameters, EcPoint, PublicKey, RsaPublicKey, SubjectPublicKeyInfo,
+};
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
+pub enum DecodedKey {
```
ionut-arm, line 15: Could you add some doc here to explain what the purpose of the type is?

THS-on: The question is if we actually need the conversion to `RsaPublicKey` and `EcPoint` exposed.
`EcPoint` is not really usable when you don't know the curve.

ionut-arm: At least for Parsec it helps - that's the format we export public keys in (the other metadata, such as the exact curve, is obtained separately)

THS-on:
I'll add tests next week. What is the best way to load an key into the TPM, so that I can check that the returned key is the same?
My idea currently is to create the keys with openssl,  use the PublicBuilder and then pass them through those functions and convert them back to DER to check if they are equal.

ionut-arm:
> What is the best way to load an key into the TPM

Have a look at `test_load_external_public` in the same file (`tss-esapi/tests/integration_tests/context_tests/tpm_commands/object_commands_tests.rs`). You just need to change the unique value set in `get_ext_rsa_pub`.

tss-esapi/src/abstraction/public.rs
```diff
@@ -0,0 +1,116 @@
...
+    /// The result can be used to convert TPM public keys to DER using `picky_asn1_der`.
+    ///
+    /// # Errors
+    /// * if other instances of [`crate::structures::Public`] are used `UnsupportedParam` will be returned.
+    fn try_from(value: Public) -> Result<Self, Self::Error> {
+        let decoded_key = public_to_decoded_key(&value)?;
+
+        match (value, decoded_key) {
+            (Public::Rsa { .. }, DecodedKey::RsaPublicKey(key)) => Ok(SubjectPublicKeyInfo {
+                algorithm: AlgorithmIdentifier::new_rsa_encryption(),
```
THS-on, line 43: @ionut-arm do you know if this is hash algorithm independent or do we actually need to use the specific identifiers for SHA1, SHA256 etc.?

ionut-arm: As far as I know/understand, this identifier is supposed to match up to the type of the key that follows in `SubjectPublicKeyInfo`. If you look at [the RFC](https://datatracker.ietf.org/doc/html/rfc8017#appendix-A.1), it does mention [an OID](http://oid-info.com/get/1.2.840.113549.1.1.1) (`pkcs-1 1`) that applies to the `RSAPublicKey` format. This is [the same one](https://github.com/Devolutions/picky-rs/blob/master/picky-asn1-x509/src/oids.rs#L54) that you get here from `picky-asn1-x509` as well, so it should be good. ([this](https://jpassing.com/2021/11/30/using-pem-to-encode-public-keys/) blog post also finds the same.) I think this is correct!

tss-esapi/src/abstraction/public.rs
THS-on: done

THS-on:
I've now added tests for RSA and ECC.

ionut-arm (approved):
Thanks, LGTM! You decided not to test with OpenSSL as well?

THS-on:
> You decided not to test with OpenSSL as well?

No because I think checking that the raw ASN.1 structures are correct is good enough.

Superhepper (approved):
Thanks for your work.

There is something about forcing dependencies upon users that not strictly needs this functionality that bothers me. But not enough to not approve of this PR. 

But I think we need to improve on this further. By moving this into Public and add these conversions as 'feature' that you can enable in your Cargo.toml.

ionut-arm:
> There is something about forcing dependencies upon users that not strictly needs this functionality that bothers me. But not enough to not approve of this PR.

I'd suggest that for v8 we split out the `abstractions` part of the crate into its own crate, because it feels like it should have a life of its own. That will also help a lot with these dependencies, and other things like API stability.

ionut-arm:
Hey, yes, I'll line up a release, but I need to get #342 in before. Hopefully some time in the following days we can get it out. Do you have a hard deadline?

THS-on:
> Do you have a hard deadline?

No I don't, so if it takes a few weeks it is fine for me.

PR #344, opened 2022-05-23
Fix dangling pointer issue

ionut-arm:
This fixes an issue with the handling of the nonce field when opening
authentication sessions. The ESYS layer expects either a pointer to a
valid nonce of at least 16 bytes, or NULL if no nonce is provided. Our
handling, however, passed down an invalid pointer that was referencing a
now-defunct structure.

What happened, both before #340 and after, was that the `Nonce` input
was moved into a separate scope, either a `match` or a lambda function,
then converted to `TPM2B_NONCE`. A reference to this `TPM2B_NONCE` was
taken and converted to `*const TPM2B_NONCE`, which was passed outside of
the scope. The pointer, therefore, ended up referencing a structure that
was dropped at the end of that inner scope.

To ensure memory safety, we need to keep ownership of the `TPM2B_NONCE`
while the call is being made.

PR #345, opened 2022-05-25
Port #344 security fix

ionut-arm:
Port the security fix related to the nonce dangling pointer during auth
session creation, and fix the other issues flagged by the compiler.

### DIFF
--- a/tss-esapi/src/context.rs
+++ b/tss-esapi/src/context.rs
@@ -56,7 +56,7 @@ pub struct Context {
     ),
     /// TCTI context handle associated with the ESYS context.
     /// As with the ESYS context, an optional Mbox wrapper allows the context to be deallocated.
-    tcti_context: TctiContext,
+    _tcti_context: TctiContext,
     /// Handle manager that keep tracks of the state of the handles and how they are to be
     /// disposed.
     handle_manager: HandleManager,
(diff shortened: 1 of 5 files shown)


tss-esapi/src/context.rs
```diff
@@ -56,7 +56,7 @@ pub struct Context {
     ),
     /// TCTI context handle associated with the ESYS context.
     /// As with the ESYS context, an optional Mbox wrapper allows the context to be deallocated.
-    tcti_context: TctiContext,
+    _tcti_context: TctiContext,
```
wiktor-k, line 59: :thinking: What's the reason for this change? I looked at #344 but I don't see any renaming... or am I looking at wrong PR?

wiktor-k (approved):
LGTM :+1:

tss-esapi/src/context.rs
ionut-arm: The main reason is "the new compilers complain about this, so I needed to make the change to get the CI to pass" 😄

ionut-arm: (doesn't like it because `_tcti_context` is not directly used anywhere, but we need to keep ownership to it because it's used by the ESAPI context to communicate)

Superhepper (approved):

PR #347, opened 2022-06-01
Expand documentation around TransientKeyContext

ionut-arm:
Expanding the documentation of structures present in the interface of
`TransientKeyContext` to allow developers to more easily understand
their contents, or to at least know to which sections of the spec to
refer to.

cc @bootstrap-prime

### DIFF
--- a/tss-esapi/src/abstraction/transient/mod.rs
+++ b/tss-esapi/src/abstraction/transient/mod.rs
@@ -56,7 +56,9 @@ pub enum KeyParams {
         scheme: RsaScheme,
         /// Public exponent of the key
         ///
-        /// If set to 0, it will default to 2^16 - 1
+        /// If set to 0, it will default to 2^16 - 1.
+        ///
+        /// (Note that the default value for [`RsaExponent`] is 0)
         pub_exponent: RsaExponent,
     },
     Ecc {
@@ -71,6 +73,9 @@ pub enum KeyParams {
 ///
 /// The `public` field represents the public part of the key in plain text,
 /// while `private` is the encrypted version of the private key.
+///
+/// For information on public key formats, see the documentation of [`PublicKey`].
+/// The private part of the key should be treated as an opaque binary blob.
 ///
 /// # Warning
 ///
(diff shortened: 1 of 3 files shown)


tss-esapi/src/utils/mod.rs
```diff
@@ -272,9 +272,20 @@ pub fn create_unrestricted_signing_ecc_public(
         .build()
 }
 
+/// Container for public key values
 #[derive(Debug, Clone, Serialize, Deserialize, Zeroize, PartialEq, Eq)]
 pub enum PublicKey {
+    /// RSA public modulus (see 27.5.3.4 in Architecture spec)
```
Superhepper, line 278: It is a little bit problematic to refer to the specification in this matter. Because no where are we stating what version of the specification we use. So in my opinion it would be better to quote of the specification instead.

```
/// RSA public modulus
///
/// # Details
/// * Trusted Platform Module Library, Part 1: Architecture, Family “2.0”, Level 00 Revision 01.59*
/// >
/// >
```

I do not know if rust doc supports quotation of the type used in markdown. But the above was just an example of a better way to do it my opinion.

tss-esapi/src/abstraction/transient/mod.rs
```diff
@@ -106,11 +118,8 @@ pub struct ObjectWrapper {
 /// The `TransientKeyContext` makes use of a root key from which the other, client-controlled
 /// keyes are derived.
 ///
-/// Currently, only functionality necessary for RSA key creation and usage (for signing,
-/// verifying signatures, encryption and decryption) is implemented. The RSA SSA
-/// asymmetric scheme with SHA256 is used for all created and imported signing keys.
-/// The RSA OAEP asymmetric scheme with SHA256 is used for all created and imported
-/// signing/encryption/decryption keys.
+/// The main goal of this abstraction is to make public key cryptography more accessible,
```
Superhepper, line 121: There is nothing wrong with the information in this section but it is something about how it is formulated... It is like it is expressing a vision rather then saying what it is actually does.

tss-esapi/src/utils/mod.rs
Superhepper: Yeah, either way is fine.

Superhepper (approved):

wiktor-k (approved):
LGTM

PR #348, opened 2022-06-10
Adds Parameters tests

Superhepper:
Adds tests for PublicParameters and some of the components it depends on.


ionut-arm (approved):
Thanks for the extra tests 🙏🏻

PR #355, opened 2022-07-02
Fixes problems reported by clippy in rust 1.62

Superhepper:
This commit fixes the problems reported by clippy in Rust 1.62.

### DIFF
--- a/tss-esapi/src/abstraction/pcr/data.rs
+++ b/tss-esapi/src/abstraction/pcr/data.rs
@@ -101,7 +101,7 @@ impl PcrData {
     }
 }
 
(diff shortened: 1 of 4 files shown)


wiktor-k:
In general this change looks very good :+1: 

I'd just maybe suggest a better commit message. Currently it seems like clippy introduced some problems while it's rather the reverse. Maybe something like "Fix code problems reported by clippy on Rust 1.62" or "Make code more idiomatic" or "Rust-1.62 compliant" or... you know what I mean :)

hug-dev (approved):

ionut-arm (approved):

PR #357, opened 2022-07-08
Add nv increment

sarnautov:


tss-esapi/src/context/tpm_commands/non_volatile_storage.rs
```diff
@@ -157,7 +158,35 @@ impl Context {
...
+        &mut self,
+        auth_handle: NvAuth,
+        nv_index_handle: NvIndexHandle,
+    ) -> Result<()> {
+        let ret = unsafe {
+            Esys_NV_Increment(
+                self.mut_context(),
+                AuthHandle::from(auth_handle).into(),
+                nv_index_handle.into(),
+                self.optional_session_1(),
```
ionut-arm, line 176: I think (?) this should be `required_session_1`, but I think we're very inconsistent in using it, and I'm not sure, at this point, what benefit it brings once we move to the new error code system in v8.0.0. Perhaps this should be something we need to scrap in that release.
(Sorry for the spam, it's not really something required for this PR, just caught sight of it now).

tss-esapi/src/context/tpm_commands/non_volatile_storage.rs
```diff
@@ -157,7 +158,35 @@ impl Context {
         }
     }
 
-    // Missing function: NV_Increment
+    /// Increment monotonic counter index
+    ///
+    /// # Details
+    /// This method is used to increment monotonic counter
+    /// in the TPM.
```
ionut-arm, line 165: We're also a bit inconsistent with documentation, but the methods in this file are so similar that it might be worth adding a doc at the module level instead, to show how to use some of the methods.
(Again, not needed for this PR, just a mental note)

ionut-arm (approved):
👍🏻 Thanks for the new method.

Superhepper (approved):
Thanks for the changes.

 I think it looks good the only thing I would like to see is perhaps some examples in the documentation but none of the other NV context methods seem to have that so I guess it can be forgiven.

tss-esapi/src/context/tpm_commands/non_volatile_storage.rs
Superhepper: I have added an issue for the missing code examples in the NV context methods. #359

PR #363, opened 2022-08-16
Conditionally expose `Esys_TR_GetTpmHandle`

mdcarr941:
Exposed the Esys_TR_GetTpmHandle function as a new `Context` method called `tr_get_tpm_handle`.

This is conditionally compiled if `has_esys_tr_get_tpm_handle` is defined, which is the case if the `generate-bindings` feature is enabled and the TSS library the bindings are generated for has version >= 2.4.0.

### DIFF
--- a/tss-esapi/src/context/general_esys_tr.rs
+++ b/tss-esapi/src/context/general_esys_tr.rs
@@ -85,6 +85,22 @@ impl Context {
         Ok(())
     }
 
+    #[cfg(has_esys_tr_get_tpm_handle)]
+    /// Retrieve the `TpmHandle` stored in the given object.
+    pub fn tr_get_tpm_handle(&mut self, object_handle: ObjectHandle) -> Result<TpmHandle> {
+        use crate::{constants::tss::TPM2_RH_UNASSIGNED, tss2_esys::Esys_TR_GetTpmHandle};
+        let mut tpm_handle = TPM2_RH_UNASSIGNED;
+        ReturnCode::ensure_success(
+            unsafe {
+                Esys_TR_GetTpmHandle(self.mut_context(), object_handle.into(), &mut tpm_handle)
+            },
+            |ret| {
+                error!("Error when getting TPM handle from ESYS handle: {}", ret);
+            },
+        )?;
+        TpmHandle::try_from(tpm_handle)
+    }
+
     // Missing function: Esys_TR_Serialize
     // Missing function: Esys_TR_Deserialize
 }
--- a/tss-esapi/tests/integration_tests/context_tests/general_esys_tr_tests.rs
+++ b/tss-esapi/tests/integration_tests/context_tests/general_esys_tr_tests.rs
@@ -375,4 +375,73 @@ mod test_tr_from_tpm_public {
         assert_eq!(expected_name, actual_name);
         assert_eq!(expected_data, actual_data);
     }
+
+    #[cfg(has_esys_tr_get_tpm_handle)]
+    #[test]
+    fn test_tr_get_tpm_handle() {
+        use tss_esapi::handles::TpmHandle;
+
+        let nv_index_tpm_handle = NvIndexTpmHandle::new(0x01500024).unwrap();
+        remove_nv_index_handle_from_tpm(nv_index_tpm_handle, Provision::Owner);
+
+        let mut context = create_ctx_without_session();
+
+        // closure for cleaning up if a call fails.
+        let cleanup = |context: &mut Context,
+                       e: tss_esapi::Error,
+                       handle: NvIndexHandle,
+                       fn_name: &str|
+         -> tss_esapi::Error {
+            // Set password authorization
+            context.set_sessions((Some(AuthSession::Password), None, None));
+            context
+                .nv_undefine_space(Provision::Owner, handle)
+                .expect("Failed to call nv_undefine_space");
+            panic!("{} failed: {}", fn_name, e);
+        };
+
+        // Create nv public.
+        let nv_index_attributes = NvIndexAttributesBuilder::new()
+            .with_owner_write(true)
+            .with_owner_read(true)
+            .build()
+            .expect("Failed to create owner nv index attributes");
+
+        let nv_public = NvPublicBuilder::new()
+            .with_nv_index(nv_index_tpm_handle)
+            .with_index_name_algorithm(HashingAlgorithm::Sha256)
+            .with_index_attributes(nv_index_attributes)
+            .with_data_area_size(32)
+            .build()
+            .unwrap();
+        ///////////////////////////////////////////////////////////////
+        // Define space
+        //
+        // Set password authorization when creating the space.
+        context.set_sessions((Some(AuthSession::Password), None, None));
+        let nv_index_handle = context
+            .nv_define_space(Provision::Owner, None, nv_public)
+            .expect("Failed to call nv_define_space");
+        ///////////////////////////////////////////////////////////////
+        // Get the TPM handle from the NV index handle object handle.
+        //
+        // Set password authorization
+        let actual = context
+            .tr_get_tpm_handle(nv_index_handle.into())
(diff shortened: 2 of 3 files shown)


tss-esapi/src/context/general_esys_tr.rs
```diff
@@ -85,6 +85,26 @@ impl Context {
         Ok(())
     }
 
+    #[cfg(has_esys_tr_get_tpm_handle)]
+    /// Retrieve the `TpmHandle` stored in the given object.
+    pub fn tr_get_tpm_handle(&mut self, object_handle: ObjectHandle) -> Result<TpmHandle> {
+        use crate::tss2_esys::Esys_TR_GetTpmHandle;
+        use tss_esapi_sys::TPM2_HANDLE;
+        let mut tpm_handle: TPM2_HANDLE = Default::default();
```
mdcarr941, line 93: I didn't see a way to create a default `TpmHandle`, so I resorted to using the underlying type here. If there's a way to do this using only `TpmHandle`, that would certainly be an improvement.

tss-esapi/src/context/general_esys_tr.rs
```diff
@@ -85,6 +85,26 @@ impl Context {
...
+                Esys_TR_GetTpmHandle(
+                    self.mut_context(),
+                    object_handle.into(),
+                    &mut tpm_handle)
+            },
+            |ret| {
+                error!("Error when getting TPM handle from ESYS handle: {}", ret);
+            },
+        )?;
+        Ok(TpmHandle::try_from(tpm_handle)?)
```
Superhepper, line 105: You do not need to do that.

Just let it return what ever comes out of the ```try_from```.

tss-esapi/tests/integration_tests/context_tests/general_esys_tr_tests.rs
```diff
@@ -375,4 +375,77 @@ mod test_tr_from_tpm_public {
...
+        //
+        // Set password authorization
+        let actual = context.tr_get_tpm_handle(nv_index_handle.into())
+            .map_err(|e| cleanup(&mut context, e, nv_index_handle, "tr_get"))
+            .expect("Failed to get TPM handle");
+        ///////////////////////////////////////////////////////////////
+        // Remove undefine the space
+        //
+        // Set password authorization
+        context.set_sessions((Some(AuthSession::Password), None, None));
```
Superhepper, line 441: Is this really necessary?

Superhepper (commented):
Thanks for the changes!

I just had a few comments.

tss-esapi/src/context/general_esys_tr.rs
Superhepper: I am not sure. You could probably find some ```null``` that you can initialize it to and then make sure it has changed.

Superhepper: Maybe ```TPM_RH_UNASSIGNED``` but I am not sure.

tss-esapi/src/context/general_esys_tr.rs
mdcarr941: Derp, that's a function and its inverse. I'm so used to converting errors in my projects. Fixed in 2e87dcb.

tss-esapi/src/context/general_esys_tr.rs
```diff
@@ -85,6 +85,26 @@ impl Context {
         Ok(())
     }
 
+    #[cfg(has_esys_tr_get_tpm_handle)]
+    /// Retrieve the `TpmHandle` stored in the given object.
```
Superhepper, line 89: We really need better documentation for these context methods. But it is not something that you need to change here. I have created an issue for that #364.

tss-esapi/tests/integration_tests/context_tests/general_esys_tr_tests.rs
mdcarr941: Yes, the test fails with TSS error code 0x0007000B ("esapi:A parameter has a bad value") if I remove this line.

Superhepper: Hrrm that is strange. But leave it if it makes the tests pass.

mdcarr941:
I'm not sure why the [ubuntu tests](https://github.com/parallaxsecond/rust-tss-esapi/runs/7873503589?check_suite_focus=true) are failing. I haven't modified any of the files listed in the error output, so I'm at a loss as to why my changes are causing this.

Superhepper:
They are failing due to clippy lint errors probably

tss-esapi/src/context/general_esys_tr.rs
Superhepper: Yeah I think that it will work because it is an internal value to the TPM that should never exist outside of it. So it is possible to add a check to see if the value has changed and if it has not then perhaps return an error.

Superhepper:
The DCO error you see that you have not signed off your commits. If you have properly configured git with name and email then use the ```git commit -s``` in order to created signed commits.


If you want to sign the commits you have created in this PR so far then just do a ```git rebase -i HEAD~5``` and chose to edit all the commits. Then you could probably do a ```git commit -s --amend```

wiktor-k:
In general this looks good but I think it would be nice if you squashed all commits into one and added DCO line (either through `git commit --signoff` or manually).

ionut-arm:
Sorry for the delay, was out and about :)

Reviewing this now, but I had an idea: what if instead of one set of bindings we commit and hold onto a range, for example by having a separate folder for bindings for 2.3.3, 2.4.6, ...? We can match the local version of the library to the correct bindings at build time.

ionut-arm (approved):
Thanks, looks good to me, just the CI stuff needs fixing

tss-esapi/src/context/general_esys_tr.rs
ionut-arm: Agree, though I have no preference either way whether that check should be done - don't think we specifically verify for potential bugs in TSS everywhere we could.

tss-esapi/tests/integration_tests/context_tests/general_esys_tr_tests.rs
ionut-arm: I'm just curious here: what call failed with that error code?

Superhepper:
@mdcarr941 Is there something wrong with your latest changes? Seems that the modifications you made are gone.

tss-esapi/tests/integration_tests/context_tests/general_esys_tr_tests.rs
mdcarr941: It was the call to `nv_undefine_space`.

tss-esapi/src/context/general_esys_tr.rs
mdcarr941: I think that kind of general error checking should be done in the `TpmHandle::try_from` method. Currently that method is returning an Ok result when given `TPM2_RH_UNASSIGNED`, perhaps we should modify it in a separate branch?

mdcarr941:
I messed up the rebase and the branch I pushed did not include all the changes. I've since force pushed a different commit that has everything in it. Sorry for all the back and forth.

Superhepper:
> I messed up the rebase and the branch I pushed did not include all the changes. I've since force pushed a different commit that has everything in it. Sorry for all the back and forth.

Nice now you just need to format your code correctly. And all checks will pass. I usually let my editor auto format for me on save. Using 'cargo fmt'.

wiktor-k:
> I messed up the rebase and the branch I pushed did not include all the changes. I've since force pushed a different commit that has everything in it. Sorry for all the back and forth.

No worries and thanks for your work! As @Superhepper mentioned we're just a `cargo fmt && git commit -a --amend --no-edit` away from merging this :)

Superhepper (approved):

PR #365, opened 2022-08-16
Improvement of FFI data zeroizing.

Superhepper:
- Adds a new InPlaceFfiDataZeroizer trait that types can implement in order to provide a zeroize capabilities for their associated FFI data.
- Implements the new InPlaceFfiDataZeroizer for several types in order to improve zerozation whe calling the create context method.
- Adds CreateInputHandler type to manage the input parameters the create context method and to zeroize the FFI data when dropped.
- Adds CreateOutputHandler to manage the FFI output parameters and when converted into CreateKeyResult properly zeroize FFI data allocated by the TSS.

### DIFF
--- a/tss-esapi/src/ffi_data.rs
+++ b/tss-esapi/src/ffi_data.rs
@@ -1,0 +1,20 @@
+// Copyright 2021 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+use crate::traits::InPlaceFfiDataZeroizer;
+use mbox::MBox;
+use std::ops::DerefMut;
+
+/// Function for taking ownership of data that has been
+/// allocated with C memory allocation functions in TSS while also
+/// zeroizing the memory before freeing it.
+#[allow(dead_code)]
+pub(crate) fn to_owned_with_zeroized_source<T, U>(data_ptr: *mut T) -> T
+where
+    T: Copy,
+    U: InPlaceFfiDataZeroizer<T>,
+{
+    let mut ffi_data = unsafe { MBox::from_raw(data_ptr) };
+    let owned_ffi_data: T = *ffi_data;
+    U::zeroize_ffi_data_in_place(ffi_data.deref_mut());
+    owned_ffi_data
+}
--- a/tss-esapi/src/ffi_data.rs
+++ b/tss-esapi/src/ffi_data.rs
@@ -7,7 +7,6 @@ use std::ops::DerefMut;
 /// Function for taking ownership of data that has been
 /// allocated with C memory allocation functions in TSS while also
 /// zeroizing the memory before freeing it.
-#[allow(dead_code)]
 pub(crate) fn to_owned_with_zeroized_source<T, U>(data_ptr: *mut T) -> T
 where
     T: Copy,
--- a/tss-esapi/src/structures/tagged/symmetric.rs
+++ b/tss-esapi/src/structures/tagged/symmetric.rs
@@ -1,3 +1,5 @@
+use zeroize::Zeroize;
+
 // Copyright 2021 Contributors to the Parsec project.
 // SPDX-License-Identifier: Apache-2.0
 use crate::{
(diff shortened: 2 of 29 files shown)


tss-esapi/src/ffi_data.rs
```diff
@@ -0,0 +1,19 @@
+// Copyright 2021 Contributors to the Parsec project.
```
ionut-arm, line 1:
```suggestion
// Copyright 2022 Contributors to the Parsec project.
```

tss-esapi/src/ffi_data.rs
```diff
@@ -0,0 +1,19 @@
...
+pub(crate) fn to_owned_with_zeroized_source<T, U>(data_ptr: *mut T) -> T
+where
+    T: Copy,
+    U: InPlaceFfiDataZeroizer<T>,
+{
+    let mut ffi_data = unsafe { MBox::from_raw(data_ptr) };
+    let owned_ffi_data: T = *ffi_data;
+    U::zeroize_ffi_data_in_place(ffi_data.deref_mut());
+    owned_ffi_data
+}
```
ionut-arm, line 19: Hmm, this feels like something MBox should be able to do... 😞

tss-esapi/src/structures/tagged/symmetric.rs
```diff
@@ -1,10 +1,13 @@
+use zeroize::Zeroize;
+
```
ionut-arm, line 2: The import is before the copyright o_O

ionut-arm (commented):
Thanks, looks quite nice!

One thing, though, shouldn't we implement something like zeroizing-on-drop using this new trait for stuff like [`PrivateKeyRsa`](https://docs.rs/tss-esapi/7.1.0/tss_esapi/structures/struct.PrivateKeyRsa.html)?

Superhepper:
> Thanks, looks quite nice!
> 
> One thing, though, shouldn't we implement something like zeroizing-on-drop using this new trait for stuff like [`PrivateKeyRsa`](https://docs.rs/tss-esapi/7.1.0/tss_esapi/structures/struct.PrivateKeyRsa.html)?

We absolutely should. I just did it for the items in the ```create``` context method to provide an example. If we think this is a way to go then I will add more commits that implements things for the other structs as well.

tss-esapi/src/ffi_data.rs
Superhepper: I agree though the MBox crate does not seem to be maintained any more.

Superhepper:
@ionut-arm, @wiktor-k 

Do you think that this draft shows the way we should take?

ionut-arm:
> Do you think that this draft shows the way we should take?

I think I'm happy with the approach, but I'll want to review it again, just to make sure I get all the details. I'm also mulling some changes related to our use of `MBox`, I'll write that up in an issue.

ionut-arm:
Damn, forgot about reviewing it again - please move it to "ready for review", I'll add it to my agenda and we can get it merged soon.

Superhepper:
I was not really looking for a review just yet more like if you approved of the way things was handled.

ionut-arm:
I've looked at it again - I think one question I have is why do stuff like `impl InPlaceFfiDataZeroizer<TPMT_SYM_DEF_OBJECT> for SymmetricDefinitionObject` instead of just `impl InPlaceFfiDataZeroizer for TPMT_SYM_DEF_OBJECT`? Even though the FFI type isn't owned by this crate, the trait is so it should work. It could then act essentially as a constructor for those FFI types.

Superhepper:
> I've looked at it again - I think one question I have is why do stuff like `impl InPlaceFfiDataZeroizer<TPMT_SYM_DEF_OBJECT> for SymmetricDefinitionObject` instead of just `impl InPlaceFfiDataZeroizer for TPMT_SYM_DEF_OBJECT`? Even though the FFI type isn't owned by this crate, the trait is so it should work. It could then act essentially as a constructor for those FFI types.

I did it this way because I thought it would break the 'Orphan rule' to implement the trait I defined for a type from another crate. But I realise this might not be the case.

PR #368, opened 2022-08-30
Changes execute_with_nullauth_session to not use extra sessions.

Superhepper:
- Makes sure that execute_with_nullauth_session does not use
  any sessions already associated with the context when calling
  start_auth_session.


Superhepper:
I do not think this is a good idea. Adding this would prevent anyone from using an outer session when creating the null auth session at that might not be desirable.

PR #375, opened 2022-09-23
Removes from slice conversions for buffer types.

### LINKED ISSUE #353: The buffer types should not implement ```TryFrom<&[u8]>```.
Several of the different buffer types are implementing ```TryFrom<&[u8]>``` including the the ```named_field_buffer_type``` macro that is called by ```buffer_type``` macro to generate some of the buffer types.

The problem is that the slice is not not consumed so it is not being converted which goes against what ```TryFrom``` is supposed to do. The buffer types should probably add a new public method ```from_bytes```. It would parhaps also make sense to change the buffer type api ```value``` to ```as_bytes```.

Superhepper:
This PR fixes #353.

- Replaces the ```value``` API with a ```as_bytes``` API for buffer types. As this name makes more sense in this context.
- Replaces the implementation of ```TryFrom<&[u8]>``` for buffer types with a ```from_bytes``` API.

### DIFF
--- a/tss-esapi/src/structures/buffers.rs
+++ b/tss-esapi/src/structures/buffers.rs
@@ -168,6 +168,38 @@ pub mod digest {
             Ok(result)
         }
     }
+
+    impl From<[u8; 20]> for Digest {
+        fn from(mut value: [u8; 20]) -> Self {
+            let value_as_vec = value.to_vec();
+            value.zeroize();
+            Digest(value_as_vec.into())
+        }
+    }
+
+    impl From<[u8; 32]> for Digest {
+        fn from(mut value: [u8; 32]) -> Self {
+            let value_as_vec = value.to_vec();
+            value.zeroize();
+            Digest(value_as_vec.into())
+        }
+    }
+
+    impl From<[u8; 48]> for Digest {
+        fn from(mut value: [u8; 48]) -> Self {
+            let value_as_vec = value.to_vec();
+            value.zeroize();
+            Digest(value_as_vec.into())
+        }
+    }
+
+    impl From<[u8; 64]> for Digest {
+        fn from(mut value: [u8; 64]) -> Self {
+            let value_as_vec = value.to_vec();
+            value.zeroize();
+            Digest(value_as_vec.into())
+        }
+    }
 }
 
 pub mod ecc_parameter {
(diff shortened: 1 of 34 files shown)


wiktor-k (approved):
I like it as it's more readable IMO :+1: 

~~Just out of curiosity is there some guideline that you've followed (e.g. discouraging using TryFrom with references or something like this)?~~

Edit: read the relevant ticket (#353) and it has the explanation. Thanks!

tss-esapi/src/structures/buffers.rs
```diff
@@ -157,17 +155,49 @@ pub mod digest {
...
         }
     }
+
+    impl From<[u8; 20]> for Digest {
+        fn from(mut value: [u8; 20]) -> Self {
+            let value_as_vec = value.to_vec();
+            value.zeroize();
+            Digest(value_as_vec.into())
+        }
+    }
```
ionut-arm, line 176: Perhaps these could be changed into a macro + 4 calls to it? I'm sure other buffers could benefit from it as well.

tss-esapi/src/structures/buffers.rs
```diff
@@ -23,8 +23,24 @@ macro_rules! named_field_buffer_type {
...
-            pub fn value(&self) -> &[u8] {
-                &self.0
+            pub fn from_bytes(bytes: &[u8]) -> Result<Self> {
+                Self::ensure_valid_buffer_size(bytes.len(), "bytes(&[u8])")?;
+                Ok($native_type(bytes.to_vec().into()))
+            }
+
+            /// Returns the content of the buffer type as
+            /// a slice of bytes.
+            pub fn as_bytes(&self) -> &[u8] {
```
ionut-arm, line 33: Would it make sense to have a `From<$native_type> for Vec<u8>` implemented? It would avoid a lot of `.as_bytes().to_vec()`

ionut-arm (approved):
Thanks! Just a couple of minor comments, feel free to ignore

tss-esapi/src/structures/buffers.rs
Superhepper: Though it is a little bit risky to do stuff like that because then that vector wont be zeroized when dropped and I wonder if the user really wants that and we do implement ```deref```.

ionut-arm: I mean, the user becomes responsible for the data - we already offer access to it, they can clone it and do the same things, this would just make it easier. We can perhaps put a warning on the API about it.

PR #376, opened 2022-10-08
Changes print of errors to hex format.

### LINKED ISSUE #369: Error message from context methods should output returncode in hex.
Context methods that calls ESAPI functions prints an error message upon errors. The return code is included in those messages but it is just a number. To make it easier to look them up in the specification they should printed in hex like they are written in the specification.

Superhepper:
This fixes #369 by changes the printing of return code values to hex in the context methods when the ESAPI functions called returns an error.

### DIFF
--- a/tss-esapi/src/context.rs
+++ b/tss-esapi/src/context.rs
@@ -101,7 +101,7 @@ impl Context {
                 )
             },
             |ret| {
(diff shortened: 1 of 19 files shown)


ionut-arm (approved):
Thanks, this will be useful!!

wiktor-k (approved):
👍 Thanks!

PR #377, opened 2022-10-09
Adds validation of session attributes.

### LINKED ISSUE #330: SessionAttributes and SessionAttributesMask should return error when bit 3 and 4 are set.
The specification says that

*Trusted Platform Module Library, Family "2.0", Revision 01.59, Section 8.4*
 >| 4:3 | Reserved | shall be CLEAR

For some of the attributes an error is thrown when the reserved bits are set but not for session attributes.

Superhepper:
Adds validation of SessionAttributes and SessionAttributesMask that checks if the reserved bits are set because the specification specifically says they shall be clear.

This will fix #330 .


wiktor-k (approved):
Looks good indeed! 👍

PR #378, opened 2022-11-12
Improvement of FFI data zeroizing(version 2).

Superhepper:
* Adds FfiDataZeroize trait and implements it for several TSS types.
* Adds CreateInputHandler type to manage the input parameters the create context method and to zeroize the FFI data when dropped.
* Adds CreateOutputHandler to manage the FFI output parameters and when converted into CreateKeyResult properly zeroize FFI data allocated by the TSS.


tss-esapi/src/ffi/data_zeroize.rs
```diff
@@ -0,0 +1,372 @@
...
+implement_ffi_data_zeroizer_trait_for_buffer_type!(TPM2B_DIGEST);
+implement_ffi_data_zeroizer_trait_for_buffer_type!(TPM2B_DATA);
+implement_ffi_data_zeroizer_trait_for_buffer_type!(TPM2B_ECC_PARAMETER);
+implement_ffi_data_zeroizer_trait_for_named_field_buffer_type!(TPM2B_ENCRYPTED_SECRET, secret);
+implement_ffi_data_zeroizer_trait_for_named_field_buffer_type!(TPM2B_ID_OBJECT, credential);
+implement_ffi_data_zeroizer_trait_for_buffer_type!(TPM2B_IV);
+implement_ffi_data_zeroizer_trait_for_buffer_type!(TPM2B_MAX_BUFFER);
+implement_ffi_data_zeroizer_trait_for_buffer_type!(TPM2B_MAX_NV_BUFFER);
+implement_ffi_data_zeroizer_trait_for_buffer_type!(TPM2B_PRIVATE);
+//implement_ffi_data_zeroizer_trait_for_buffer_type!(TPM2B_PUBLIC);
```
ionut-arm, line 104: Needed?

tss-esapi/src/ffi.rs
```diff
@@ -0,0 +1,27 @@
...
+/// The owned version of the FFI data.
+pub(crate) fn to_owned_with_zeroized_source<T>(ffi_data_ptr: *mut T) -> T
+where
+    T: FfiDataZeroize + Copy,
+{
+    let mut ffi_data = unsafe { MBox::from_raw(ffi_data_ptr) };
+    let owned_ffi_data: T = *ffi_data.deref();
+    ffi_data.ffi_data_zeroize();
+    owned_ffi_data
+}
```
ionut-arm, line 27: I keep forgetting - why do we need to copy here instead of just returning the data that was behind the pointer? Is it not possible to take proper ownership of it?

ionut-arm (commented):
Thanks for the new implementation, I like this one more than v1! Do you have a preference among the two?

tss-esapi/src/ffi.rs
Superhepper: > I keep forgetting - why do we need to copy here instead of just returning the data that was behind the pointer? Is it not possible to take proper ownership of it?

The MBox has the into_inner function which would allow us to immediately take ownership of the data and free the source. Unfortunately it does not allow us to zeroize the source. So instead here copy the data to the stack then zeroize the mermory pointed to by the MBox(i.e. the source).

tss-esapi/src/ffi/data_zeroize.rs
Superhepper: > Needed?

No it is not needed it is implemented further down.

Superhepper:
> Thanks for the new implementation, I like this one more than v1! Do you have a preference among the two?

Not really both works for me.

tss-esapi/src/ffi.rs
ionut-arm: I think this is where I always get confused, because I expect `MBox` to take ownership by just wrapping the source pointer into an owned object, instead of copying it and freeing the source.

Superhepper: Yes, this might not be the best way of doing it. Maybe we should implement ```TryFrom<MBox<T>>``` for some of our types.

ionut-arm:
I think I'm happy with how this looks, are you looking to implement anything else here?

Superhepper:
> I think I'm happy with how this looks, are you looking to implement anything else here?

I was thinking of writing some simple units tests and maybe have a look at implementing TryFrom for some of the MBox and use the zeroize functions in there instead.

tss-esapi/src/ffi.rs
Superhepper: We can use TryFrom with MBox but it does not solve the problem of zeroizing the source allocated by tpm2-tss before freeing it. So I think this is currently the way forward. Will add the handling of other sensitive data in other PRs.

tss-esapi/src/ffi/data_zeroize.rs
```diff
@@ -0,0 +1,444 @@
...
+            let mut actual = $tss_type {
+                size: 10,
+                $buffer_field_name: fill_with_ones(),
+            };
+            actual.ffi_data_zeroize();
+            assert_eq!(expected.size, actual.size);
+            assert_eq!(expected.$buffer_field_name, actual.$buffer_field_name);
+        }
+    };
+}
```
ionut-arm, line 394: Excellent, thanks!

ionut-arm (approved):
Based on previous reviews, looks good to me! I'm wondering if the rationale and the thought process behind why you've done it like this is worth documenting in `data_zeroize.rs` or somewhere.

tss-esapi/src/ffi/data_zeroize.rs
Superhepper: Yes, I will add that when I have the time. Probably sometime next year.

wiktor-k (approved):
It's a good way forward. I've read the diff in full and don't see any outstanding problems. (Sorry for the delay in reviewing... quite a lot on my plate recently :/ )

tss-esapi/src/ffi.rs
Superhepper: ```TryFrom``` for ```MBox``` did not really improve things. So I am happy with this for now.

Superhepper:
@ionut-arm @wiktor-k

I have added a new commit with some rationale behind why it is done this way please read it an see if I have written something crazy.

ionut-arm (approved):
👍🏻 LGTM, thank you for the great effort!!

PR #379, opened 2022-11-12
Fixes build problems related to compilers and editions.

Superhepper:
- Changes bool conversion to u8 by using the solution suggested by Clippy when it reported this issue.
- Changed MSRV to 1.56

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
       - uses: actions/checkout@v2
       - name: Check formatting
         run: cargo fmt --all -- --check
-  # Check that it builds with the Minimum Supported Rust Version      
+  # Check that it builds with the Minimum Supported Rust Version
   msrv:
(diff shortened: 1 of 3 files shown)


ionut-arm (approved):
👍🏻

wiktor-k (approved):
I think this is OK but I'd split these two changes into separate commits (address clippy, update the compiler) or even just mention that the compiler has been updated in the commit message.

Thanks :+1:

Superhepper:
If I would do that. Then one of the commits wouldn't build.

wiktor-k:
🤔 I thought it depends on the ordering. If you fix the clippy issue first it should be OK for the old compiler too.

Superhepper:
> 🤔 I thought it depends on the ordering. If you fix the clippy issue first it should be OK for the old compiler too.

No because. The old compiler issue is related to a dependency of dependency where they between two patch releases have updated the edition which causes old compilers to not work. So that wont build even if we fix the clippy problem.

PR #380, opened 2022-11-15
tcti: adds support for unix sockets

baloo:
TSS added support for unix sockets in 3.2.0, this commit bring support for configuration of unix sockets to talk to either swtpm or mssim.

Upstream PR: https://github.com/tpm2-software/tpm2-tss/pull/2212

### DIFF
--- a/tss-esapi/build.rs
+++ b/tss-esapi/build.rs
@@ -12,6 +12,11 @@ fn main() {
     let supported_tss_version =
         VersionReq::parse("<4.0.0, >=2.3.3").expect("Failed to parse supported TSS version");
 
+    let has_simulator_unix_socket = VersionReq::parse(">=3.2.0").unwrap();
+    if has_simulator_unix_socket.matches(&tss_version) {
+        println!("cargo:rustc-cfg=has_simulator_unix_socket")
+    }
+
     let hierarchy_is_esys_tr_req = VersionReq::parse(">=3.0.0").unwrap();
     if hierarchy_is_esys_tr_req.matches(&tss_version) {
         println!("cargo:rustc-cfg=hierarchy_is_esys_tr")
--- a/tss-esapi/build.rs
+++ b/tss-esapi/build.rs
@@ -12,11 +12,6 @@ fn main() {
     let supported_tss_version =
         VersionReq::parse("<4.0.0, >=2.3.3").expect("Failed to parse supported TSS version");
 
-    let has_simulator_unix_socket = VersionReq::parse(">=3.2.0").unwrap();
-    if has_simulator_unix_socket.matches(&tss_version) {
-        println!("cargo:rustc-cfg=has_simulator_unix_socket")
-    }
-
     let hierarchy_is_esys_tr_req = VersionReq::parse(">=3.0.0").unwrap();
     if hierarchy_is_esys_tr_req.matches(&tss_version) {
         println!("cargo:rustc-cfg=hierarchy_is_esys_tr")
--- a/tss-esapi/src/tcti_ldr.rs
+++ b/tss-esapi/src/tcti_ldr.rs
@@ -272,13 +272,16 @@ fn validate_from_str_tcti() {
         })
     );
 
-    let tcti = TctiNameConf::from_str("mssim:path=/foo/bar").unwrap();
-    assert_eq!(
-        tcti,
-        TctiNameConf::Mssim(TpmSimulatorConfig::Unix {
-            path: "/foo/bar".to_string(),
-        })
-    );
+    #[cfg(has_simulator_unix_socket)]
+    {
+        let tcti = TctiNameConf::from_str("mssim:path=/foo/bar").unwrap();
+        assert_eq!(
+            tcti,
+            TctiNameConf::Mssim(TpmSimulatorConfig::Unix {
+                path: "/foo/bar".to_string(),
+            })
+        );
+    }
 
     let tcti = TctiNameConf::from_str("swtpm:port=1234,host=168.0.0.1").unwrap();
     assert_eq!(
@@ -298,13 +301,16 @@ fn validate_from_str_tcti() {
         })
     );
 
-    let tcti = TctiNameConf::from_str("swtpm:path=/foo/bar").unwrap();
-    assert_eq!(
-        tcti,
-        TctiNameConf::Swtpm(TpmSimulatorConfig::Unix {
-            path: "/foo/bar".to_string(),
-        })
-    );
+    #[cfg(has_simulator_unix_socket)]
+    {
+        let tcti = TctiNameConf::from_str("swtpm:path=/foo/bar").unwrap();
+        assert_eq!(
+            tcti,
+            TctiNameConf::Swtpm(TpmSimulatorConfig::Unix {
+                path: "/foo/bar".to_string(),
+            })
+        );
+    }
 
     let tcti = TctiNameConf::from_str("device:/try/this/path").unwrap();
     assert_eq!(
--- a/tss-esapi/src/tcti_ldr.rs
+++ b/tss-esapi/src/tcti_ldr.rs
@@ -379,9 +379,14 @@ fn validate_from_str_device_config() {
     assert_eq!(config.path, PathBuf::from("/dev/tpm0"));
 }
 
-/// Configuration for an Mssim TCTI context
+/// Configuration for an Mssim or Swtpm TCTI context
 ///
 /// The default configuration will point to `localhost:2321`
+///
+/// Examples:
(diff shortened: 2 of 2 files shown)


Superhepper:
Another thing is you should probably add some kind of error when some one is trying to use unix sockets with a version tss that does not support it. By adding your own configuration item in '/rust-tss-esapi/tss-esapi/build.rs'.

tss-esapi/build.rs
```diff
@@ -12,6 +12,11 @@ fn main() {
     let supported_tss_version =
         VersionReq::parse("<4.0.0, >=2.3.3").expect("Failed to parse supported TSS version");
 
+    let hierarchy_is_esys_tr_req = VersionReq::parse(">=3.2.0").unwrap();
```
Superhepper, line 15: Yes, exactly like this! But you should probably rename that so it makes sense.
```rust
has_simulator_unix_socket_ req = VersionReq::parse(">=3.2.0").unwrap();
if has_simulator_unix_socket_ req.matches(&tss_version) {
    println!("cargo:rustc-cfg=has_simulator_unix_socket")
}
```

baloo: haha :D sorry

tss-esapi/src/tcti_ldr.rs
```diff
@@ -356,35 +383,58 @@ fn validate_from_str_device_config() {
...
+            if let Some(path) = path_pattern
+                .captures(config_str)
+                .and_then(|c| c.get(2))
+                .map(|m| m.as_str())
+            {
+                return Ok(TpmSimulatorConfig::Unix {
+                    path: path.to_string(),
+                });
+            }
+        }
```
ionut-arm, line 436: Hmm, this means that the Unix socket has "priority" (i.e., if you have `,port=1234,host=127.0.0.1,path=/tmp/sock`, it'll use the Unix socket configuration), but the TCP socket is treated as default, if I understand correctly. This should probably be documented somewhere, and also tested.

Is this also an explicit choice upstream, or is it at most coincidental?

ionut-arm (commented):
Hey, thanks for the patch!! :]

tss-esapi/src/tcti_ldr.rs
baloo: (sorry I missed the notification)

It was just easier for me to implement that way, the upstream just parse it so that the last of `host` or `path` has priority and nullify the other.

ionut-arm: Sorry, I drifted away as well...

That sounds like a reasonable argument for doing it this way, I think my only request would be to add a couple of tests for this ordering, and maybe a line in the doc of `TpmSimulatorConfig`.

baloo: @ionut-arm just to be clear, you expect me to replicate the behavior of upstream, correct?

ionut-arm: Ah, I misread that, apologies, thought your approach was identical to upstream. Hmm, well this is frustrating, would've been nice if the upstream project exposed functionality to parse this so we could all align.

But, since this is about a malign configuration, I wouldn't put too much effort in trying to implement the same treatment everywhere. Users really shouldn't expect an arguably wrong input to be parsed in some particular way.

Nah, just go ahead and document it, no need to change anything. Sorry again for the confusion.

tss-esapi/src/tcti_ldr.rs
```diff
@@ -379,9 +379,14 @@ fn validate_from_str_device_config() {
     assert_eq!(config.path, PathBuf::from("/dev/tpm0"));
 }
 
-/// Configuration for an Mssim TCTI context
+/// Configuration for an Mssim or Swtpm TCTI context
 ///
 /// The default configuration will point to `localhost:2321`
+///
+/// Examples:
+///  - Use of a TCP connection: `swtpm:host=localhost.port=2321`
```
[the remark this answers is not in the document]
baloo: typo :D

tss-esapi/src/tcti_ldr.rs
```diff
@@ -352,39 +379,67 @@ fn validate_from_str_device_config() {
...
+        /// Defaults to `localhost`
+        host: ServerAddress,
+        /// Port used by the server at the address given in `host`
+        ///
+        /// Defaults to `2321`
+        port: u16,
+    },
+    #[cfg(has_simulator_unix_socket)]
+    /// Connects to tpm over Unix socket
+    Unix { path: String },
```
wiktor-k, line 405: :thinking: Correct me if I'm wrong but this makes the `Unix` variant disappear when building on systems without unix sockets support.

Since this variant doesn't have any Unix specific types I'm wondering if that doesn't make writing clients to `rust-tss-esapi` harder. That is if a crate that used `tss-esapi` crate would have to do the detection too (that is writing a build script that propagates `#[cfg(has_simulator_unix_socket)]`).

Wouldn't it be better to actually expose the Unix variant always but only when *using* that variant return an error that this is unsupported?

(That said I'm not completely sure if that's a good approach but I think it's worth considering).

wiktor-k (approved):
In general this looks OK and I don't mind merging as is. One note I added though.

tss-esapi/src/tcti_ldr.rs
baloo: This is only for the swtpm which is for simulation / tests. And unix socket is probably even more niche.

Beside that, I'm usually not too comfortable with compile-time checks as the library you link to might be changed from one without support to one with and still not get support. But here, this will be deployed on a nixos so ... my opinion doesn't really matter here.

Up to you folks really.

baloo: (if we were to expose the unix flavor anyway that would trigger a configuration error sent from the C library which you already have to deal with).

Superhepper: I totally agree. This ```cfg``` should be removed. The comment should be updated with information that using this on platforms not supporting unix sockets will result in an error when calling ```TctiContext::initialize```. Then in the the ```initialize function we should check if unix sockets are supported otherwise return an InvalidParam error.

baloo: Tried by best to document the error case, let me know.

Superhepper: I think that is fine. 

I realize that I did not explain my self properly and I apologise for that. So let me explain more thoroughly how 
I meant. What I meant was that the configuration item that you had added would remain but but the check would be moved to ```TctiContyext::initialize``` and the error that would occur would be documented there.


Addition to the ```TctiContyext::initialize``` documentation:
```rust
/// * Errors
///  - Returns InvalidParam error if TpmSimulatorConfig specifies a unix socket and the tpm2-tss
///    version of the library does not support it.
```

The in the initialize function one would add (not sure if you can negate the 'has_unix_socket_support' in cfg_if using 'not' otherwise it should probably be changed so we check if unix socket support is missing):
```rust
cfg_if::cfg_if!{
    if #[cfg(has_unix_socket_support)] {
    } else {
        if let TctiNameConf::Mssim(TpmSimulatorConfig::Unix(_)) = name_conf {
            error!("tpm2-tss library does not support unix sockets.");
            return Err(Error::local_error(WrapperErrorKind::InvalidParam));
        }

        if let TctiNameConf::Swtpm(TpmSimulatorConfig::Unix(_)) = name_conf {
            error!("tpm2-tss library does not support unix sockets.");
            return Err(Error::local_error(WrapperErrorKind::InvalidParam));
        }
    }
}
```

And then the documentation for the enum item could then be something like:
```rust
    /// Connects to TPM over Unix socket
    ///
    /// If used with a library without unix socket support (< 3.2.0),
    /// an InvalidParam error will be returned upon TCTI initialization.
    Unix { path: String },
```

But as I said this is just an example of how I envisioned it and I am fine with merging it if you prefer it your way.

baloo: There problem with the `#[cfg(has_unix_socket_support)]` is that it looks up the version of `tpm2-tss` at buildtime of that crate. If further down the line we were to upgrade `tpm2-tss` to `>=3.2` it would still return `InvalidParam` and I'm not sure it's the best thing.

Superhepper: Yes, I think you are right. The only other option would be to add some kind of tpm2-tss version runtime check to the initialize function or make some kind of analysis of the error returned from ```Tss2_TctiLdr_Initialize``` to see if was caused by using unix sockets with a library that did not support it. But I think that is a thing we could do in another PR if we feel there is a need for it to improve the user experience.

Superhepper (approved):

PR #386, opened 2023-01-18
Update bindgen to 0.63.0

aplanas:
The optional bindgen dependency was set to a version that depends on an old version of clap, and also depends on the ansi_term crate, now unmaintained and causing audit complains. See [1] for an example.

This patch upgrade to bindgen 0.63.0 and drop the dependency to the affected crate.

[1] https://github.com/keylime/rust-keylime/pull/502

Signed-off-by: Alberto Planas <aplanas@suse.com>


aplanas:
I moved it down lo 0.60 from 0.63, as since 0.62 it requires rustc 1.57

aplanas:
> The code in there needs to be re-generated.

Should be done. I take more time as I needed to make changes in regenerate-binginds. Sadly I needed to drop the darwin one as I do not have access to one

Superhepper:
I think it is @ionut-arm that have done these upgrades in the past. Maybe he can help with that one.

aplanas:
> Yes, please, leave the Darwin ones as they are (perhaps ignore them somehow if they lead to failures), and I'll update after this is merged.

Sure, dropping the commit

ionut-arm:
So, I tried porting the `regenerate-bindings` script from here, I struggled to get `curl` to actually cross-compile, but then I realised we don't need curl or json-C - those are only used for the FAPI layer in `tpm2-tss`, and you can disable that by passing `--enable-fapi=no` when configuring `tpm2-tss`.

aplanas:
Thanks for taking some bits in your PR. Can I close this one?

ionut-arm:
You can if you want, and I'll open a new one to do the updates for the `main` branch as well (since we don't constantly rebase it on v7).

Issue #383, opened 2022-12-13
Retrieve EK


tabajara-scontain:
Hi all,

I'm having trouble retrieving the EK from a simulated TPM (SWTPM).
Every time I try to read the certificate I get and error: `WrapperError(InvalidHandleState)`

It seems function `tr_from_tpm_public` is being called twice, and `Esys_TR_FromTPMPublic` fetches the same nv index.
 
 The snippet I'm trying to execute is the one below. I also tried to run your [tests](https://github.com/parallaxsecond/rust-tss-esapi/blob/072bd208548f39ef428270527039de67cadede96/tss-esapi/tests/integration_tests/abstraction_tests/ek_tests.rs) to retrieve the EK changing the simulator from MSSIM to SWTPM, and it also fails. 
 
 ```rust
 let mut context = Context::new(TctiNameConf::from_str("swtpm:port=2321,host=localhost")?)?;
 
 let ek_certificate = &ek::retrieve_ek_pubcert(&mut context, AsymmetricAlgorithm::Rsa)?;
 ```
 
 Any ideas on what the problem can be?
 
 Setup:
 - tss-esapi: 7.1.0 
 - swtpm: 0.8.0
 - tpm2-tss: 3.2.1

ionut-arm:
Hey! Before I dive into it too much, could you try using the tip of `main` for `tss-esapi`, instead of `7.1.0`? We've had some changes in that area since the release, it might solve the problem. We'll worry about what that means release-wise afterwards.

tabajara-scontain:
Hi, I downloaded the source code from the main and executed the test below using SWTPM instead of MSSIM.

```
running 1 test
[2022-12-14T06:26:31Z ERROR tss_esapi::context::handle_manager] Handle(1078035592) is already open
test abstraction_tests::ek_tests::test_retrieve_ek_pubcert ... FAILED
```

**Sidenote**: when building rust-tss-esapi I got `Unsupported TSS version 3.2.1-rc0`. I just ignored it. Let me know if I shouldn't have.

ionut-arm:
Hmmm, ok, will try to reproduce - a bit weird, we already have CI jobs using `swtpm` and running that particular test.

ansasaki:
This looks related: https://github.com/tpm2-software/tpm2-tss/issues/2537

ansasaki:
[...]

PR #387, opened 2023-01-27
Allow use with v4 of tpm2-tss

ionut-arm:
Also adding a test job for Fedora rawhide, which uses tpm2-tss v4.0.0 (for now).

@ansasaki - does this seem reasonable?

### DIFF
--- a/tss-esapi-sys/build.rs
+++ b/tss-esapi-sys/build.rs
@@ -57,6 +57,7 @@ fn main() {
     }
 }
 
+#[allow(clippy::uninlined_format_args)]
 #[cfg(feature = "generate-bindings")]
 pub fn generate_from_system(esapi_out: PathBuf) {
     pkg_config::Config::new()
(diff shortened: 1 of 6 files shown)


wiktor-k (approved):
I'd say this looks very clean and minimal. A couple of ignored test cases to be investigated but gosh, I wish that all major version bumps were this easy to handle!

Good job @ionut-arm :+1:

ionut-arm:
> I'd say this looks very clean and minimal. A couple of ignored test cases to be investigated but gosh, I wish that all major version bumps were this easy to handle!

I don't think they need to be investigated, actually! See the latest comments on https://github.com/parallaxsecond/rust-tss-esapi/issues/383 - we'll just need to re-enable once the upstream problem goes away.

And yes, so far so good in terms of supporting v4 of tpm2-tss, it's mostly a testament to the work those folks have been doing to keep backwards compatibility for the ESAPI layer :)

ansasaki:
@ionut-arm Yes, this looks good, thank you for the prompt action!

wiktor-k:
> I don't think they need to be investigated, actually! See the latest comments on https://github.com/parallaxsecond/rust-tss-esapi/issues/383 - we'll just need to re-enable once the upstream problem goes away.

That's cool! I wonder if we need a FIXME comment in the source code or at least an action item on #383 to un-ignore these with later versions? Not sure about your memory but mine is excellent but very short :)

Superhepper:
I have just looked through it briefly but from what I have seen it looks pretty good.

Superhepper (approved):

Superhepper:
@ionut-arm 
We need to think carefully about if changing minimum rust version in that branch is a breaking change.

ionut-arm:
> @ionut-arm We need to think carefully about if changing minimum rust version in that branch is a breaking change.

Yeah, I know. I guess if you have an application, not a library, you can control those dependencies underneath `tss-esapi` and keep using 1.53. But for libraries that doesn't work, right? They'd have to do the same as us.

Superhepper:
Patch versions get updated like automatically but if we introduce it in a new minor version it would be possible lock your version of the library...

ionut-arm:
Oh, I just realised that we're not actually making any changes that break our crate, we're just saying we don't/can't support 1.53 anymore because of changes upstream of us. I think that's perfectly fine - we can't really control any of that. Regardless of whether we change the MSRV, people will hit that problem.

> Patch versions get updated like automatically but if we introduce it in a new minor version it would be possible lock your version of the library...

Not really, AFAIK after v1.0.0 the minor numbers behave the same way - you can force Cargo to use a specific version, but by default it will use the latest with an equal major number.

Ok, I think I feel better about this now - I will bump the minor and do a release ASAP, after I get an ACK that this sounds reasonable to you as well, @wiktor-k @Superhepper.

Superhepper:
Yeah sounds good

Superhepper:
Can we merge this to master as well so that I do not need to redo all the fixes for the build.

wiktor-k:
Yep, looks sensible and I agree with your reasoning @ionut-arm! :+1: 

Thanks!

PR #388, opened 2023-01-31
Fix for handles not being unique.

Superhepper:
Handles returned from tpm2-tss APIs does not need to be unique. This an attempt to fix problem #383.

This problem was also discussed in https://github.com/tpm2-software/tpm2-tss/issues/2537.

### DIFF
--- a/tss-esapi/src/context/handle_manager.rs
+++ b/tss-esapi/src/context/handle_manager.rs
@@ -40,10 +40,14 @@ impl HandleManager {
             return Err(Error::local_error(WrapperErrorKind::InvalidParam));
         }
 
-        if self.open_handles.contains_key(&handle) {
-            error!("Handle({}) is already open", ESYS_TR::from(handle));
-            return Err(Error::local_error(WrapperErrorKind::InvalidHandleState));
+        // The TSS might return the same handle, see #383
+        if let Some(stored_handle_drop_action) = self.open_handles.get(&handle) {
+            if handle_drop_action != *stored_handle_drop_action {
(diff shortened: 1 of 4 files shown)


ionut-arm:
Will need to port this to v7 as well. I can open another PR after I merge the current one.

aplanas:
This should be already merged in #390? Safe to close?

ionut-arm (approved):
👍🏻

ionut-arm:
> This should be already merged in #390? Safe to close?

Depends on whether @Superhepper prefers getting rid of the `unwrap` or leaving as is - I, for one, don't mind either way.

Superhepper:
This basically the same thing this is just to master.

tss-esapi/src/context/handle_manager.rs
```diff
@@ -41,9 +41,14 @@ impl HandleManager {
         }
 
         if self.open_handles.contains_key(&handle) {
-            error!("Handle({}) is already open", ESYS_TR::from(handle));
-            return Err(Error::local_error(WrapperErrorKind::InvalidHandleState));
+            // It is safe to call unwrap because the existance of the key has already been verified.
+            let stored_handle_drop_action = self.open_handles.get(&handle).unwrap();
```
wiktor-k, line 45: It's possible to avoid the unwrap by using `if let` just like in the other PR.

I trust you'll adjust it and I don't want to unnecessarily block the workflow so I'll ack the PR now :)

wiktor-k (approved):

PR #390, opened 2023-02-01
Fix handle management for static resources

ionut-arm:
Making "duplicate" handles acceptable, as per the convo in #383. Also fixing tests that are impacted by this, and documenting it.

cc @aplanas @nullr0ute @ansasaki


Superhepper (approved):

ionut-arm:
@Superhepper - hope you're ok with being a co-author for this commit 😜

Superhepper:
Yeah absolutely I have already reviewed it hehe.

ionut-arm:
Unfortunately the fix for the tests breaks in the environments where the TSS is an older version 🥲  need to figure out a fix there

tss-esapi/src/abstraction/nv.rs
```diff
@@ -63,6 +65,8 @@ fn get_nv_index_info(
 }
 
 /// Lists all the currently defined NV Indexes' names and public components
+///
+/// NOTE: This call _may_ close existing ESYS handles to the existing NV Indexes.
```
Superhepper, line 69: Why do you say "_may_"?

ionut-arm: Because I assume on "old" versions of tpm2-tss, it doesn't, and that's why I have to do `unwrap_or_else` - that handle is still open, since the TSS (probably) created new ESYS handles every time.

tss-esapi/src/context/handle_manager.rs
```diff
@@ -41,8 +41,12 @@ impl HandleManager {
         }
 
         if self.open_handles.contains_key(&handle) {
-            error!("Handle({}) is already open", ESYS_TR::from(handle));
-            return Err(Error::local_error(WrapperErrorKind::InvalidHandleState));
+            // It is safe to call unwrap because the existance of the key has already been verified.
+            let stored_handle_drop_action = self.open_handles.get(&handle).unwrap();
```
wiktor-k, line 45: Why not just fold this with the if as something like:

```rust
if let Some(stored_handle_drop_action) = self.open_handles.get(&handle) {
```

and avoid the `unwrap` altogether? :)

tss-esapi/src/context/handle_manager.rs
```diff
@@ -41,8 +41,12 @@ impl HandleManager {
...
+            if handle_drop_action != *stored_handle_drop_action {
+                error!("Handle drop action inconsistency");
+                return Err(Error::local_error(WrapperErrorKind::InconsistentParams));
+            }
         }
         let _ = self.open_handles.insert(handle, handle_drop_action);
```
wiktor-k, line 51: :thinking: is `let _` necessary? It seems `insert` returns an `Option` and it's not marked with `#[must_use]`... ah... this is outside of this change... sorry :)

wiktor-k (approved):
Looks good with one minor remark :+1:

tss-esapi/tests/integration_tests/abstraction_tests/nv_tests.rs
```diff
@@ -81,12 +80,15 @@ fn list() {
         .map(|(public, _)| public.nv_index())
         .any(|x| x == nv_index));
 
+    // Need to get the ESYS handle again, as it was closed by nv::list above
+    let owner_nv_index_handle = context
+        .tr_from_tpm_public(nv_index.into())
+        .unwrap_or_else(|_| owner_nv_index_handle.into());
```
aplanas, line 86: Replace `|_|` with `||`, there is no parameter in the closure if I am not wrong

tss-esapi/tests/integration_tests/abstraction_tests/nv_tests.rs
```diff
@@ -98,8 +100,12 @@ fn read_full() {
     // Now read it back
     let read_result = nv::read_full(&mut context, NvAuth::Owner, nv_index);
 
+    // Need to get the ESYS handle again, as it was closed by nv::read_full above
+    let owner_nv_index_handle = context
+        .tr_from_tpm_public(nv_index.into())
+        .unwrap_or_else(|_| owner_nv_index_handle.into());
```
aplanas, line 106: same?

ionut-arm:
> > Unfortunately the fix for the tests breaks in the environments where the TSS is an older version smiling_face_with_tear need to figure out a fix there
> 
> If you know which tests are affected maybe at least add a comment or error message for people encountering the issue (something along the `did you try updating your TSS? See issue #XYZ`). So that when it actually fails it's easier to diagnose what went wrong.

Yeah, perhaps I should've posted an update after I fixed those tests 😄 but yes, they're all working now, as per the last commit.

tss-esapi/tests/integration_tests/abstraction_tests/nv_tests.rs
ionut-arm: That's true for `Option`, but not for [`Result`](https://doc.rust-lang.org/std/result/enum.Result.html#method.unwrap_or_else). Don't think the tests would compile otherwise!

tss-esapi/src/context/handle_manager.rs
ionut-arm: Fixed :)

PR #392, opened 2023-02-02
Update bindgen

ionut-arm:
cc @aplanas


Superhepper:
Should you update the dependency in tss-esapi Cargo.toml as well?

Superhepper:
What version of tpm2-tss did you use to regenerate these files?

ionut-arm:
> What version of tpm2-tss did you use to regenerate these files?

2.40, I think - how much does it matter? Should I use something like 3.0.0 instead?

> Should you update the bindgen dependency in tss-esapi Cargo.toml as well?

The what?

ionut-arm:
Oh, wait, no, I used the version from the string in the script - `3.2.1`. I think? I'll need to check.

Superhepper:
Forget what I said about bindgen in tss-esapi I must have been dreaming. I thought we had a feature or build dependency in there as well.

Previously you have generated the bindings from the lowest version of tpm2-tss that we support.

aplanas (approved):

ionut-arm:
Ok, all set now.

Superhepper (approved):

wiktor-k (approved):
Looks good 👍 sorry for the delay... was traveling 😅

PR #394, opened 2023-03-11
Adds tests for WrapperErrorKind.

Superhepper:
- Adds tests that checks that the implementation of the Display trait for WrapperErrorKind is producing the expected messages.


wiktor-k (approved):
In general I'm ok with the change. The one wrinkle through is that some of the strings have first letter upper cased but some don't. This kind of inconsistency just hurts my eyes a little :)

Superhepper:
> In general I'm ok with the change. The one wrinkle through is that some of the strings have first letter upper cased but some don't. This kind of inconsistency just hurts my eyes a little :)

It hurt my eyes too. But I was not sure if I should go ahead and change it. But now that I know I am not alone I think I will do it hehehe.

wiktor-k:
Looks very nice. Thanks! 👍

ionut-arm (approved):
Thanks!

PR #406, opened 2023-05-18
Bump enumflags2 to 0.7.7

nullr0ute:
There's a RUSTSEC-2023-0035 CVE which is addressed in 0.7.7 so we should bump that to the minimum revision.


Superhepper (approved):
Thank you for the PR!

wiktor-k (approved):
LGTM :+1: thanks!

PR #407, opened 2023-05-18
Adds tests for vendor specific errors.

Superhepper:
- Adds tests that check that a TPM format zero vendor specific response code gets parsed and converted properly.

- Adds a test that is checking if the correct wrapper error is returned when the reserved bit is SET in the TPM format zero response code.

- Adds a test that is checking if the correct wrapper error is returned when the TPM 2.0 bit is CLEAR in the TPM format zero response code.


ionut-arm (approved):
Excellent, thanks!

wiktor-k (approved):

PR #410, opened 2023-06-04
Expanding UnMarshall + a few additions

ionut-arm:
Hi all,

This PR contains two commits:

* first one adds another method to the UnMarshall trait to allow it to unmarshall from the buffer at a given offset, and return the post-unmarshalling offset back; it also adds Marshall/UnMarshall for CommandCode and u32
* second one adds CommandTag as another StructureTag, as well as Marshall/UnMarshall for this new enum, and for the AttestationType enum

Obviously, also adding tests for the new stuff. The plan is to modify the remaining UnMarshall implementations to use unmarshall_offset as the "base" implementation, and unmarshall to just use that underneath.

Let me know if you'd like me to split out the 2nd commit - I've put them together in one PR since the 2nd depends on the 1st.

### DIFF
--- a/tss-esapi/src/constants/command_code.rs
+++ b/tss-esapi/src/constants/command_code.rs
@@ -163,22 +163,7 @@ impl Marshall for CommandCode {
         let mut buffer = vec![0; Self::BUFFER_SIZE];
         let mut offset = 0;
 
-        ReturnCode::ensure_success(
-            unsafe {
-                crate::tss2_esys::Tss2_MU_TPM2_CC_Marshal(
-                    (*self).into(),
-                    buffer.as_mut_ptr(),
-                    Self::BUFFER_SIZE.try_into().map_err(|e| {
-                        error!("Failed to convert size of buffer to TSS size_t type: {}", e);
-                        Error::local_error(WrapperErrorKind::InvalidParam)
-                    })?,
-                    &mut offset,
-                )
-            },
-            |ret| {
-                error!("Failed to marshal CommandCode: {}", ret);
-            },
-        )?;
+        self.marshall_offset(&mut buffer, &mut offset)?;
 
         let checked_offset = usize::try_from(offset).map_err(|e| {
             error!("Failed to parse offset as usize: {}", e);
@@ -188,6 +173,30 @@ impl Marshall for CommandCode {
         buffer.truncate(checked_offset);
 
         Ok(buffer)
+    }
+
+    fn marshall_offset(
+        &self,
+        marshalled_data: &mut [u8],
+        offset: &mut std::os::raw::c_ulong,
+    ) -> Result<()> {
+        ReturnCode::ensure_success(
+            unsafe {
+                crate::tss2_esys::Tss2_MU_TPM2_CC_Marshal(
+                    (*self).into(),
+                    marshalled_data.as_mut_ptr(),
+                    marshalled_data.len().try_into().map_err(|e| {
+                        error!("Failed to convert size of buffer to TSS size_t type: {}", e);
+                        Error::local_error(WrapperErrorKind::InvalidParam)
+                    })?,
+                    offset,
+                )
+            },
+            |ret| {
+                error!("Failed to marshal CommandCode: {}", ret);
+            },
+        )?;
+        Ok(())
     }
 }
 
--- a/tss-esapi/src/constants/command_code.rs
+++ b/tss-esapi/src/constants/command_code.rs
@@ -158,23 +158,6 @@ impl From<CommandCode> for TPM2_CC {
 impl Marshall for CommandCode {
     const BUFFER_SIZE: usize = std::mem::size_of::<TPM2_CC>();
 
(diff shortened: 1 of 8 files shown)


tss-esapi/src/constants/command_code.rs
```diff
@@ -150,3 +154,67 @@ impl From<CommandCode> for TPM2_CC {
...
     }
 }
+
+impl Marshall for CommandCode {
+    const BUFFER_SIZE: usize = std::mem::size_of::<TPM2_CC>();
+
+    /// Produce a marshalled [TPM2_CC]
+    fn marshall(&self) -> Result<Vec<u8>> {
+        let mut buffer = vec![0; Self::BUFFER_SIZE];
+        let mut offset = 0;
```
Superhepper, line 164: Should not marshall then behave the same way?

tss-esapi/src/traits.rs
```diff
@@ -15,4 +20,74 @@ pub trait Marshall: Sized {
 pub trait UnMarshall: Sized {
     /// Creates the type from marshalled data.
     fn unmarshall(marshalled_data: &[u8]) -> Result<Self>;
+
+    /// Creates the type from the marshalled data, and modifies
+    /// the `offset` to point to the first byte in the `marshalled_data`
+    /// buffer which was not used in the conversion.
+    fn unmarshall_offset(_marshalled_data: &[u8], _offset: &mut u64) -> Result<Self> {
+        unimplemented!();
```
wiktor-k, line 28: Unimplemented will panic. I wonder what's the reason for not leaving it empty (with no implementation)? Because other types are not implementing that yet?

At the end of the day I guess it would be good to provide default implementation to the `unmarshall` that just calls `unmarshall_offset(marshalled_data, 0)` :thinking:

wiktor-k (approved):
Looks OK :+1: although it's not my area of expertise :sweat_smile:

tss-esapi/src/constants/command_code.rs
ionut-arm: By this you mean having a `marshall_offset` function as well?

tss-esapi/src/traits.rs
wiktor-k: Okay, this sounds fair. Thanks for taking the time to explain!

tss-esapi/src/constants/command_code.rs
Superhepper: Exactly or maybe that was something that was intensionally left out?

ionut-arm: That's very true - I implemented this as I needed the unmarshall in particular, but it makes sense to make it "symmetrical" :) I'll add that as a commit on top of these two, stay tuned

ionut-arm: Added

Superhepper (approved):

ionut-arm:
I've had to update the `_offset` versions of the methods because they were hardcoded to `u64`, whereas the original `size_t` type is 32 bits on some platforms.

ionut-arm:
Also it seems the `log` crate has been bumped beyond our MSRV underneath our feet, so we have to decide whether we want to bump the MSRV or fix the `log` version.

cc @gowthamsk-arm

Superhepper:
> Also it seems the `log` crate has been bumped beyond our MSRV underneath our feet, so we have to decide whether we want to bump the MSRV or fix the `log` version.
> 
> cc @gowthamsk-arm

It would be nice if we can lock it to a specific version but I am also completely ok with bumping MSRV.

Superhepper (approved):

tss-esapi/src/constants/command_code.rs
```diff
@@ -150,3 +154,79 @@ impl From<CommandCode> for TPM2_CC {
         command_code.to_u32().unwrap()
     }
 }
+
+impl Marshall for CommandCode {
+    const BUFFER_SIZE: usize = std::mem::size_of::<TPM2_CC>();
+
+    /// Produce a marshalled [TPM2_CC]
+    fn marshall(&self) -> Result<Vec<u8>> {
```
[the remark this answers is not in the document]
ionut-arm: You mean one that redirects to the `_offset` method, and rely on that being separate between the types implementing it?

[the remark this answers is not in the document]
Superhepper: Sorry I always comment on things as they pop up in my head and not necessarily at the correct location which can be confusing. This comment probably belongs in the Marshall trait.

Yeah I mean the ```marshall``` code is basically the same for all different types. So with a few tweaks it should be possible to make a generic ```marshall``` function in the trait that calls the ```marshall_offset``` which makes the actual work.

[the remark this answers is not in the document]
Superhepper: In order to avoid code duplication.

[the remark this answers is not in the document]
ionut-arm: Okay, I've done that in a new commit!

PR #412, opened 2023-06-06
Fix codespell action

wiktor-k:
Previously the action did not checkout the repository and as such actually didn't do what it was supposed to.

This patch fixes all typos that accumulated over the time. Adds the checkout action and fixes the codespell version action to `v1`.

Additionally codespell settings have been moved to a file so that the `codespell` command line can be executed locally using the same settings as in the CI.

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,0 +1,3 @@
+[codespell]
+skip = .git,target,Cargo.lock
+ignore-words-list = crate,keypair,daa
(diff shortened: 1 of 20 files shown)


Superhepper:
Excellent! This will be very helpful especially for me.

Superhepper (approved):

ionut-arm (approved):
Thank you for taking care of this! 🙏🏻

PR #414, opened 2023-06-11
ek, ak abstractions: allow to specficy exact key type

THS-on:
This adds support other keys than RSA2048 and ECC Nist P256.

### DIFF
--- a/tss-esapi/src/abstraction/ek.rs
+++ b/tss-esapi/src/abstraction/ek.rs
@@ -2,11 +2,11 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::{
-    abstraction::{nv, IntoKeyCustomization, KeyCustomization},
+    abstraction::{nv, AsymmetricAlgorithmSelection, IntoKeyCustomization, KeyCustomization},
     attributes::ObjectAttributesBuilder,
     handles::{KeyHandle, NvIndexTpmHandle, TpmHandle},
     interface_types::{
-        algorithm::{AsymmetricAlgorithm, HashingAlgorithm, PublicAlgorithm},
+        algorithm::{HashingAlgorithm, PublicAlgorithm},
         ecc::EccCurve,
         key_bits::RsaKeyBits,
         resource_handles::{Hierarchy, NvAuth},
@@ -24,12 +24,20 @@ use std::convert::TryFrom;
 const RSA_2048_EK_CERTIFICATE_NV_INDEX: u32 = 0x01c00002;
 const ECC_P256_EK_CERTIFICATE_NV_INDEX: u32 = 0x01c0000a;
 
+// Source: TCG EK Credential Profile for TPM Family 2.0; Level 0 Version 2.3 Revision 2
+// Section 2.2.1.5 (High Range)
+const ECC_P384_EK_CERTIFICATE_NV_INDEX: u32 = 0x01c00016;
+const ECC_P521_EK_CERTIFICATE_NV_INDEX: u32 = 0x01c00018;
+const ECC_P256_SM2_EK_CERTIFICATE_NV_INDEX: u32 = 0x01c0001a;
+const RSA_3072_EK_CERTIFICATE_NV_INDEX: u32 = 0x01c0001c;
+const RSA_4096_EK_CERTIFICATE_NV_INDEX: u32 = 0x01c0001e;
+
 /// Get the [`Public`] representing a default Endorsement Key
 ///
 /// Source: TCG EK Credential Profile for TPM Family 2.0; Level 0 Version 2.3 Revision 2
 /// Appendix B.3.3 and B.3.4
 pub fn create_ek_public_from_default_template<IKC: IntoKeyCustomization>(
-    alg: AsymmetricAlgorithm,
+    alg: AsymmetricAlgorithmSelection,
     key_customization: IKC,
 ) -> Result<Public> {
     let key_customization = key_customization.into_key_customization();
@@ -65,7 +73,7 @@ pub fn create_ek_public_from_default_template<IKC: IntoKeyCustomization>(
     ];
 
     let key_builder = match alg {
-        AsymmetricAlgorithm::Rsa => PublicBuilder::new()
+        AsymmetricAlgorithmSelection::Rsa(key_bits) => PublicBuilder::new()
             .with_public_algorithm(PublicAlgorithm::Rsa)
             .with_name_hashing_algorithm(HashingAlgorithm::Sha256)
             .with_object_attributes(obj_attrs)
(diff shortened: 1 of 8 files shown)


tss-esapi/src/abstraction/ek.rs
```diff
@@ -18,18 +18,76 @@ use crate::{
...
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+pub enum AsymmetricAlgorithmSelection {
+    Rsa2048,
+    Rsa3072,
+    Rsa4096,
+    EccP256,
+    EccP384,
+    EccP521,
+    EccP256Sm2,
+    Null,
```
ionut-arm, line 45: I'm curious - did you leave this in as an artifact of the fact that we already had null, and it seemed more complicated to get rid of it?

THS-on: Yes this is a left over. I could change the `From<AsymmetricAlgorithm>` to `TryFrom<AsymmetricAlgorithm>` and then drop the `Null`, because for these functions it does not really make sense.

THS-on: I now removed the `Null` case.

THS-on:
@ionut-arm the Original `AsymmetricAlgorithm` is consumed by also by the AK abstraction layer. So I now moved `AsymmetricAlgorithmSelection` into `algorithms` and then changed the other functions also to it.

The major change introduced is that now create_ak, also requires a key algorithm specified.

ionut-arm (commented):
Thanks for the new and improved patch!

tss-esapi/src/interface_types/algorithm.rs
```diff
@@ -335,6 +336,60 @@ impl TryFrom<TPMI_ALG_ASYM> for AsymmetricAlgorithm {
...
+/// Use this instead of [AsymmetricAlgorithm], which will be deprecated at some point.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+pub enum AsymmetricAlgorithmSelection {
+    Rsa2048,
+    Rsa3072,
+    Rsa4096,
+    EccP256,
+    EccP384,
+    EccP521,
+    EccP256Sm2,
```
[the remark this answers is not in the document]
wiktor-k: I agree. It seems using parameterized variants would simplify this code.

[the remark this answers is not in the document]
THS-on: Done, yes that made the code simpler. Thanks for the suggestion :+1:

tss-esapi/src/interface_types/algorithm.rs
```diff
@@ -335,6 +336,28 @@ impl TryFrom<TPMI_ALG_ASYM> for AsymmetricAlgorithm {
...
 
+/// Enum representing the asymmetric algorithm interface type with specific properties.
+/// Use this instead of [AsymmetricAlgorithm], which will be deprecated at some point.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum AsymmetricAlgorithmSelection {
+    Rsa(RsaKeyBits),
+    Ecc(EccCurve),
+}
+
+impl TryFrom<AsymmetricAlgorithm> for AsymmetricAlgorithmSelection {
```
ionut-arm, line 347: Could you add some public-facing documentation here that the conversion makes some assumptions about what RSA size/ECC curve you want?

ionut-arm (approved):
Thanks for all the changes! LGTM now, just left a nit comment, but otherwise happy to merge.

Also, I think we can merge despite the MSRV-related failure, we'll have to decide what to do with `log`.

tss-esapi/src/interface_types/algorithm.rs
THS-on: Done. I also removed the deprecation notice for `AsymmetricAlgorithm`, because it is just a direct mapping of the TSS esapi enum.

wiktor-k (approved):
LGTM :+1:

Superhepper:
I am not against this. 

But is not possible to achieve the same thing using the KeyCustomization trait? And if it is should we really have two ways of doing the same thing?

tss-esapi/src/interface_types/algorithm.rs
```diff
@@ -335,6 +337,32 @@ impl TryFrom<TPMI_ALG_ASYM> for AsymmetricAlgorithm {
...
+        match value {
+            AsymmetricAlgorithm::Rsa => Ok(AsymmetricAlgorithmSelection::Rsa(RsaKeyBits::Rsa2048)),
+            AsymmetricAlgorithm::Ecc => Ok(AsymmetricAlgorithmSelection::Ecc(EccCurve::NistP256)),
+            AsymmetricAlgorithm::Null => {
+                Err(Error::local_error(WrapperErrorKind::UnsupportedParam))
+            }
+        }
+    }
+}
+
```
Superhepper, line 365: I think it would be nice with a conversion from ```AsymmetricAlgorithmSelection``` to a ```NvIndexTpmHandle``` so that one can avoid those constants in ek.rs.

tss-esapi/src/interface_types/algorithm.rs
```diff
@@ -335,6 +337,32 @@ impl TryFrom<TPMI_ALG_ASYM> for AsymmetricAlgorithm {
     }
 }
 
+/// Enum representing the asymmetric algorithm interface type with specific properties.
+///
+/// # Details
+/// Use this instead of [AsymmetricAlgorithm].
```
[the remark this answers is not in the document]
THS-on: I moved it now into the abstraction model.

tss-esapi/src/interface_types/algorithm.rs
THS-on: `AsymmetricAlgorithmSelection` is also used for AKs and in that context the conversion to an EK certificate index does not really make sense, but indeed it would make the conversion nicer.

Superhepper: hrrm no that is true. In order to make that work you have to create an extra type (something like ```EkCertificateNvIndex```) and then implement the desired conversions for it. But I am not sure if it is worth the trouble.

Superhepper (approved):

THS-on:
@ionut-arm @Superhepper would it possible to backport this to 7.X? 

This is currently the main blocker for us to enable ECC fully in the rust Keylime agent, see: https://github.com/keylime/rust-keylime/issues/513

Superhepper:
Yeah, I am not sure if I have the time to do it any time soon but it is just a matter of writing a PR with changes for the 7.y.x branch instead.

Superhepper:
> @ionut-arm @Superhepper would it possible to backport this to 7.X?
> 
> This is currently the main blocker for us to enable ECC fully in the rust Keylime agent, see: [keylime/rust-keylime#513](https://github.com/keylime/rust-keylime/issues/513)

I tried to adapt your code to get it to work under the 7.x.y branch. https://github.com/parallaxsecond/rust-tss-esapi/pull/546
please try it out and see if it works or if it causes some weird issues.

PR #416, opened 2023-06-30
Implement `AsRef<[u8]>` for all buffer types

wiktor-k:
This allows using type generated by `named_field_buffer_type` macro in all places where `AsRef<[u8]>` arguments are accepted and also returning them via `impl` return types which hide the underlying implementation type.

### DIFF
--- a/tss-esapi/src/interface_types/algorithm.rs
+++ b/tss-esapi/src/interface_types/algorithm.rs
@@ -292,8 +292,8 @@ impl TryFrom<TPMI_ALG_SYM_MODE> for SymmetricMode {
 /// Enum representing the asymmetric algorithm interface type.
 ///
 /// # Details
-/// Use [AsymmetricAlgorithmSelection] instead where possible.
-/// This corresponds to TPMI_ALG_ASYM.
+/// Use [crate::abstraction::AsymmetricAlgorithmSelection] instead where possible.
+/// This corresponds to `TPMI_ALG_ASYM`.
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
 pub enum AsymmetricAlgorithm {
     Rsa,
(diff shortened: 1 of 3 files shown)


wiktor-k:
Not sure why the doc for unrelated stuff broke now but I think I fixed it :sweat_smile:

Superhepper (commented):
I do not see any update to any tests. Will they automatically be touching this addition to the code?

wiktor-k:
I'll add some that illustrate the use case I have in mind...

Superhepper (approved):

tss-esapi/src/interface_types/algorithm.rs
```diff
@@ -292,8 +292,8 @@ impl TryFrom<TPMI_ALG_SYM_MODE> for SymmetricMode {
 /// Enum representing the asymmetric algorithm interface type.
 ///
 /// # Details
-/// Use [AsymmetricAlgorithmSelection] instead where possible.
-/// This corresponds to TPMI_ALG_ASYM.
+/// Use [crate::abstraction::AsymmetricAlgorithmSelection] instead where possible.
+/// This corresponds to `TPMI_ALG_ASYMq.
```
ionut-arm, line 296:
```suggestion
/// This corresponds to `TPMI_ALG_ASYM`.
```

ionut-arm (approved):

tss-esapi/src/interface_types/algorithm.rs
wiktor-k: Ugh, that was pretty embarrassing :sweat_smile: I wonder why didn't `cargo doc` complain about broken markup though... :thinking: 

One way or another -> I've applied that fix to the commit and rebased everything accordingly. Thanks for your eagle eyes! :eagle:

PR #419, opened 2023-07-15
Move MSRV to 1.60.0

ionut-arm:
Our dependency on the `log` crate means that currently the crate does not compile with our MSRV. It makes more sense for us to bump the MSRV and stay aligned with updates of `log`.


wiktor-k (approved):
It's kind of crazy to update because of the log crate but lets check the facts: 1.60 was released on Apr. 7, 2022 (https://blog.rust-lang.org/2022/04/07/Rust-1.60.0.html). We use  the same version in Sequoia (https://gitlab.com/sequoia-pgp/sequoia#requirements-and-msrv) and we are *very* conservative in this aspect.

So I'm very much in favor! 👍

Superhepper (approved):

PR #420, opened 2023-07-24
Fixes remaining tests for Error types.

Superhepper:
This PR restructures the testing of the error tests while also adds tests for trait implementations that are not derived.


tss-esapi/src/error/return_code/tpm/format_zero.rs
```diff
@@ -85,7 +85,7 @@ impl std::fmt::Display for TpmFormatZeroResponseCode {
         match self {
             TpmFormatZeroResponseCode::Error(e) => e.fmt(f),
             TpmFormatZeroResponseCode::Warning(e) => e.fmt(f),
-            TpmFormatZeroResponseCode::VendorSpecific(_) => write!(f, "Vendor specific error"),
+            TpmFormatZeroResponseCode::VendorSpecific(_) => write!(f, "Vendor specific error."),
```
wiktor-k, line 88: Why not print the vendor specific u16 in here too? (maybe formatting in hex?). When bad stuff happens I like to have as much info as possible :)

wiktor-k:
```suggestion
            TpmFormatZeroResponseCode::VendorSpecific(code) => write!(f, "Vendor specific error: 0x{code:x}."),
```

(note: I didn't test it :sweat_smile: )

tss-esapi/tests/integration_tests/error_tests/return_code_tests/tpm_tests/tpm_format_zero_tests/tpm_format_zero_warning_tests.rs
```diff
@@ -230,3 +230,44 @@ fn test_display_implementation() {
...
+    test_error_number_method!(TpmFormatZeroWarning::ReferenceS1);
+    test_error_number_method!(TpmFormatZeroWarning::ReferenceS2);
+    test_error_number_method!(TpmFormatZeroWarning::ReferenceS3);
+    test_error_number_method!(TpmFormatZeroWarning::ReferenceS4);
+    test_error_number_method!(TpmFormatZeroWarning::ReferenceS5);
+    test_error_number_method!(TpmFormatZeroWarning::ReferenceS6);
+    test_error_number_method!(TpmFormatZeroWarning::NvRate);
+    test_error_number_method!(TpmFormatZeroWarning::Lockout);
+    test_error_number_method!(TpmFormatZeroWarning::Retry);
+    test_error_number_method!(TpmFormatZeroWarning::NvUnavailable);
```
wiktor-k, line 272: :thinking: Hmm... too bad there's no easy way to [enumerate all enum variants](https://stackoverflow.com/q/21371534). This would take a lot of this boilerplate code out (I just had to open the source code to see if all variants were being covered).

wiktor-k (approved):
It looks good in general :+1:

tss-esapi/src/error/return_code/tpm/format_zero.rs
Superhepper: Because usually all the the API calls looks like this:

```rust
        ReturnCode::ensure_success(
            unsafe {
                // Call to some ESYS API
            },
            |ret| {
                error!("Error when performing <some esys api>: {:#010X}", ret);
            },
        )?;
```

So I am not so sure the return code will be printed any way. I am not sure the return code without layer information will add anything to that.

tss-esapi/tests/integration_tests/error_tests/return_code_tests/tpm_tests/tpm_format_zero_tests/tpm_format_zero_warning_tests.rs
wiktor-k: I'd add it only during testing (with `#[cfg_attr(test, ...)]`) and make strum a dev dependency... not sure about the full solution though so I don't want to push you in that direction.

ionut-arm (approved):
Thanks, looks good to me! 👍🏻

tss-esapi/tests/integration_tests/error_tests/return_code_tests/tpm_tests/tpm_format_zero_tests/tpm_format_zero_warning_tests.rs
Superhepper: I t would be interesting and certainly something I would like to look into in another PR.

Superhepper: It is unfortunately not that easy. Because the integration tests tests everything as if the thing tested are an actual crate. So in order to to do this we would probably have to add a ```integration-tests``` feature and then ccompile it using ```--features integration-tests``` when run it in ci.

PR #423, opened 2023-07-27
Adds strum as a integration test requirement.

Superhepper:
-Adds integration-test feature in order to conditionally add
 dependencies.

-Updates the TpmFormatZeroWarning so it is possible to iterate over all
 the items when creating integration tests.


Superhepper:
Was it something like this that you had in mind @wiktor-k ?

wiktor-k:
Yeah the test code got significantly smaller. There's a slight cost of setup that I think we'd pay only once and then all similar tests can just reuse what you did here. 👍 

I see that the CI complains about MSRV thought :/

Superhepper:
> I see that the CI complains about MSRV thought :/

That was just because I thought I was going to be clever and update a dependency.

wiktor-k (approved):
I think it looks good. 👍 

I wonder how many tests could be improved using this pattern now 😁

ionut-arm (approved):
This looks like a pretty useful pattern, might be good to deploy for other cases too.

PR #424, opened 2023-07-29
Fixes problem with lint warnings in different versions of Rust.

Superhepper:
- The crate supports several different versions of rust and in some of these versions the Clippy lint tool introduces new warnings that are not present in previous versions of rust. This causes a warning when compiling. This has been fixed in this commit by using the `rustversion` crate to only add the lint if the crate is being built with a version of Rust that has it.

### DIFF
--- a/tss-esapi-sys/Cargo.toml
+++ b/tss-esapi-sys/Cargo.toml
@@ -16,6 +16,7 @@ links = "tss2-esys"
 bindgen = { version = "0.63.0", optional = true }
 pkg-config = "0.3.18"
 target-lexicon = "0.12.0"
+rustversion = "1.0.14"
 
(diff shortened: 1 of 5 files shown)


wiktor-k:
Looks good. 👍 

 I think it would be good to have an actual CI job that'd use the oldest version of Rust compiler that we support thus verifying that this change does what it claims to do.

(Or maybe it is already implemented... can't quickly check on mobile 😔)

Superhepper:
I have already made a MSRV job. It is one of the checks that must be ok before you can merge a PR.

ionut-arm (approved):
👍🏻

wiktor-k (approved):
LGTM 👍

PR #427, opened 2023-08-21
Bump picky- crates

ionut-arm:
A recent fix [1] in the picky-asn1-der crate highlights an issue that could lead to faulty results handed off by the abstractions in the tss-esapi crate. Specifically, mishandling of ASN.1 Integers could result in invalid encodings of various pieces of data such as elliptic curve points.

[1] https://github.com/Devolutions/picky-rs/pull/209/files


Superhepper (approved):

ionut-arm:
Unfortunately this won't work with the current MSRV of v7 (1.57.0). Would it make sense to bump that to 1.60.0, or do you reckon that's a breaking change?

Superhepper:
For master I think it is a non issue to bump MSRV to 1.60

ionut-arm:
> For master I think it is a non issue to bump MSRV to 1.60

True, though I'm hoping to bump it for the `7.x.y` branch, as I'd need to release a new version for Parsec

Superhepper:
That is slightly more complicated. Because that has the potential of breaking stuff if we only make a minor release bump. My suggestion, that I know for a lot of reason is not ideal is to do the following.

1. Merge the stuff that you need into the Major Version 7 branch and make a new release of that branch.
2. Create a Major Version 8 branch from the latest commit on the Major Version 7 branch. And then in that add a commit that updates the MSRV to the desired value and make a release of the Major Version 8 branch.

Ideally I would like for the the current master to be start of the Major Version 8 branch. But that might not be possible.

ionut-arm:
Mentioning the same thing as in the other PR: [tentative recommendation](https://github.com/rust-lang/api-guidelines/discussions/231) for Rust APIs seems to be that minor bumps are ok for MSRV changes. So we should be good to make a v7 update just for the MSRV.

Also, it seems we should be more thorough in checking our MSRV, either with a lock file or with a compiler option to keep every dependency at minimum level.

wiktor-k (approved):
Ugh, why it complains about `rustix` while I can't find it in the project: `cargo tree --all-features --target=all | grep rustix` :/

I'm OK with the change though :+1:

PR #431, opened 2023-09-10
Improve error messages of Public{Ecc,Rsa}ParametersBuilder

brandsimon:
I am not sure, if the `'restricted' and 'is_decrypt_key'` is wanted, since it is used that way in computing languages (but not in human language).

### DIFF
--- a/tss-esapi/src/structures/tagged/public/ecc.rs
+++ b/tss-esapi/src/structures/tagged/public/ecc.rs
@@ -155,12 +155,12 @@ impl PublicEccParametersBuilder {
                     return Err(Error::local_error(WrapperErrorKind::InconsistentParams));
                 }
             } else {
-                error!("Found symmetric parameter, expected it to be Null nor not set at all because 'restricted' and 'is_decrypt_key' are set to false");
+                error!("Symmetric parameter was not set but 'restricted' and 'is_decrypt_key' are set to true");
                 return Err(Error::local_error(WrapperErrorKind::ParamsMissing));
             }
         } else if let Some(symmetric) = self.symmetric {
             if !symmetric.is_null() {
-                error!("Found symmetric parameter, expected it to be Null nor not set at all because 'restricted' and 'is_decrypt_key' are set to false");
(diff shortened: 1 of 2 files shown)


brandsimon:
Tbh I am not sure if the logic here is correct.
I am trying to port a shell-script to rust with no success at all. It loads an object into the TPM (protected by a primary-endorsement-key) and signs something with it.

I'd appreciate if someone can have a look, the shell script, rust-code and output are here:
https://gist.github.com/brandsimon/b6c898a5a5021cf9757756a8319b741e
The public and private portion can be created with this script: (choose PCR 7):
https://gitlab.com/cryptographic_id/cryptographic-id-rs/-/blob/v0.2.2/usr/bin/cryptographic_id_add_initramfs

ionut-arm:
Hey! Thanks for the patch. Looking at your output and Rust code, it seems you're getting `TPM_RC_SCHEME`, which (given [the documentation](https://trustedcomputinggroup.org/wp-content/uploads/TCG_TPM2_r1p59_Part3_Commands_pub.pdf) of TPM2_Create) would mean you might be using the wrong ECC scheme. For [our EK abstraction](https://github.com/parallaxsecond/rust-tss-esapi/blob/main/tss-esapi/src/abstraction/ek.rs#L101) we use `EccScheme::Null` - perhaps you can try with that?

brandsimon:
@ionut-arm 
Thank you very much! I think your hint was a huge step into the right direction.

Superhepper:
There are a lot fo magical things tpm2-tools does in the background that one unfortunately have to do manually when porting from a script into a program. So if one wants to find out the root of the problem one needs to dig through the tpm2-tool code and have the specification ready at hand. But as @ionut-arm mentioned and from my own experience a mismatch in scheme between either a key and operation one tries to perform is not unusual.

wiktor-k (approved):
I think it looks good :+1: . Fix the spelling though :sweat_smile:

brandsimon:
@wiktor-k 
I change a typo against a typo 😆

@Superhepper 
Thank you, I found a lot in the tpm2-tools (like correct algorithms). I still struggle a bit, but I don't thick it could be related to the logic in PublicEccParametersBuilder anymore. I will ask in the slack channel if I have more questions.

PR #433, opened 2023-09-13
Fix MSRV-related issues

ionut-arm:
Fixing the MSRV in our CI involves making sure that the crate can build under that version of the Rust compiler, given some choice of dependencies that are also compatible with the MSRV. A major issue is that many crates bump the MSRV in patch or minor releases that then get picked up automatically if no lock file is present.

This commit adds a lock file for our current MSRV, and pins the MSRV for clippy as well.


.clippy.toml
```diff
@@ -0,0 +1 @@
+msrv = "1.30.0"
```
Superhepper, line 1: Well this was pleasant surprise. But isn't it a bit low?

wiktor-k: Yep, looks suspiciously low.

tss-esapi/tests/all-ubuntu.sh
```diff
@@ -15,6 +15,9 @@ set -euf -o pipefail
 #################################################
 if [[ ! -z ${RUST_TOOLCHAIN_VERSION:+x} ]]; then
 	rustup override set ${RUST_TOOLCHAIN_VERSION}
+	# Use the frozen Cargo lock to prevent any drift from MSRV being upgraded
+	# underneath our feet.
+	cp tests/Cargo.lock.frozen ../Cargo.lock
```
wiktor-k, line 20: :thinking: why not just store Cargo.lock there instead of copying? So that people don't commit it by accident?

.clippy.toml
ionut-arm: Oops! 😅 My mistake, copy-pasted without changing.

tss-esapi/tests/all-ubuntu.sh
ionut-arm: Because that Cargo lock would then apply to all builds, e.g. when packaging the crate, or when we do development. And we don't want that, we just want to check the MSRV.
I might be wrong though?

wiktor-k: :thinking: I don't know... it just somehow reminded me of this recent blogpost: https://blog.rust-lang.org/2023/08/29/committing-lockfiles.html

ionut-arm: > Since there isn't a universal answer to these situations, we felt it was best to leave the choice to developers and give them information they need in making a decision.

Ugh, that's sometimes the worst route.

It might be worth committing the Cargo.lock and setting up dependabot to keep track of updates (like [native-pkcs11](https://github.com/google/native-pkcs11) does), but I'd have to leave that to some other time and close this in the meanwhile.

Superhepper (approved):

wiktor-k (approved):

tss-esapi/tests/all-ubuntu.sh
ionut-arm: I'll look into making that lock file a permanent fixture for our builds :) hopefully this week

PR #437, opened 2023-09-28
Added integration test for pcr_extend

mkulke:
I didn't find a test for extending PCRs in the repo, it would be useful also as a reference

### DIFF
--- a/tss-esapi/tests/integration_tests/abstraction_tests/pcr_tests.rs
+++ b/tss-esapi/tests/integration_tests/abstraction_tests/pcr_tests.rs
@@ -1,11 +1,60 @@
 // Copyright 2021 Contributors to the Parsec project.
 // SPDX-License-Identifier: Apache-2.0
 use crate::common::create_ctx_without_session;
+use sha2::{Digest, Sha256};
 
 use tss_esapi::{
+    handles::PcrHandle,
     interface_types::algorithm::HashingAlgorithm,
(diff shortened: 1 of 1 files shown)


Superhepper:
Well this is nice. But I think this is in the wrong place. ```pcr_extend``` is an actual Esys API so it should be a context method. 

So I think a better place for this test would be [here](https://github.com/parallaxsecond/rust-tss-esapi/blob/main/tss-esapi/tests/integration_tests/context_tests/tpm_commands/integrity_collection_pcr_tests.rs).

mkulke:
> Well this is nice. But I think this is in the wrong place. `pcr_extend` is an actual Esys API so it should be a context method.
> 
> So I think a better place for this test would be [here](https://github.com/parallaxsecond/rust-tss-esapi/blob/main/tss-esapi/tests/integration_tests/context_tests/tpm_commands/integrity_collection_pcr_tests.rs).
> 
> Or maybe att it as an example in the documentation for that method if it is not properly documented.

Ah indeed, thx for clarifying. the linked test seems to cover pcr_extend. from glancing over it, I assumed it was testing some advanced tpm ceremony, but it's more or less the same test. I'll change this into an example similar to the hmac one and then you can decide whether it's a helpful addition.

Superhepper:
Is this still in active development or should this be closed?

Superhepper:
Closed due to non activity.

PR #439, opened 2023-09-29
Bump bindgen and update bindings

gowthamsk-arm:
Signed-off-by: Gowtham Suresh Kumar <gowtham.sureshkumar@arm.com>


tss-esapi-sys/src/bindings/x86_64-unknown-linux-gnu.rs
```diff
@@ -10798,20 +10774,6 @@ extern "C" {
         nonceTPM: *mut *mut TPM2B_NONCE,
     ) -> TSS2_RC;
 }
-extern "C" {
```
Superhepper, line 10801: There is something strange with these generated files. What version of TSS are these created from and what version do we intend to support in this branch?

tss-esapi-sys/src/bindings/x86_64-unknown-linux-gnu.rs
```diff
@@ -3498,13 +3474,13 @@ impl Default for TPMS_AUTH_RESPONSE {
 }
 pub type TPMI_AES_KEY_BITS = TPM2_KEY_BITS;
 pub type TPMI_SM4_KEY_BITS = TPM2_KEY_BITS;
-pub type TPMI_CAMELLIA_KEY_BITS = TPM2_KEY_BITS;
```
Superhepper, line 3501: This kind of name change is a breaking change because these constants and names are exposed through the crate.

tss-esapi-sys/src/bindings/x86_64-unknown-linux-gnu.rs
gowthamsk-arm: Just ran the `regenerate-bindings.sh` script which has `TPM2_TSS_VERSION="2.3.3"`. Is there some external dependency that needs to be considered before generating the script?

gowthamsk-arm:
The PR does not update the [x86_64-unknown-darwin.rs](https://github.com/parallaxsecond/rust-tss-esapi/blob/7.x.y/tss-esapi-sys/src/bindings/x86_64-unknown-darwin.rs) binding :( 

@ionut-arm mentioned that updating the bindings for this target is challenging as Mac doesn't even let you install the Rust toolchain locally for cross-compilation. Do you have any suggestions on how this can be done?

gowthamsk-arm:
The [x86_64-unknown-linux-gnu.rs](https://github.com/parallaxsecond/rust-tss-esapi/pull/439/files/b8cba8c0a5bea61349e876f6bcd42e386fb31b92#diff-c02c0bd01f8117ebbd920e0724ed5c8f6be325f04ea9b9601e678869b4910b9e) already was having some differences compared to the other 3 binding files. 

- `TPMI_CAMELLIA_KEY_BITS` is only present in the x86 linux binding file. The other bindings use  `TPMI_TPM2_CAMELLIA_KEY_BITS`. 
- `Esys_TR_GetTpmHandle`, `Esys_TRSess_GetAuthRequired` and `Esys_GetSysContext` are only present for the x86 linux bindings.

Was this done intentionally by any chance?

tss-esapi-sys/src/bindings/x86_64-unknown-linux-gnu.rs
ionut-arm: This is unfortunate, it might've been a mistake when we last updated these, but in 2.3.3 these functions are not present. Perhaps we ran with 2.4.0? Given that we'll be doing a minor update on v7, maybe we can bump to the latest v2 of tpm2-tss, 2.4.6. What do you think?

ionut-arm:
@wiktor-k @Superhepper - any thoughts on moving to v2.4.6 of tpm2-tss?

ionut-arm:
> it would require a major version bump.

Why? I'm hoping we'd only have additions to the bindings, nothing removed (or if it would be removed, for example to rename from `TPMI_CAMELLIA_KEY_BITS` to `TPMI_TPM2_CAMELLIA_KEY_BITS` we can probably keep both), which would be compatible with a minor version bump.

Superhepper:
If it is only an addition and it still works with the old version even if we do not support it then I guess no major version change is required.

gowthamsk-arm:
So after generating the new bindings with the `2.4.6` version of the library, I noticed that the x86 bindings were already generated for `2.4.6` and the rest of the bindings were not updated. This explains the discrepancies between the bindings. 

The new patch updates the other bindings to `2.4.6`. As discussed, I have manually edited the bindings file to not introduce any breaking change. But one thing that is a bit tricky is `TPMI_TPM2_CAMELLIA_KEY_BITS`.  This has been changed to `TPMI_CAMELLIA_KEY_BITS` in the newer version. Currently I have just added the new definition and didn't change the struct which is using the old `TPMI_TPM2_CAMELLIA_KEY_BITS`. 
```
pub union TPMU_SYM_KEY_BITS {
    pub aes: TPMI_AES_KEY_BITS,
    pub sm4: TPMI_SM4_KEY_BITS,
    pub camellia: TPMI_TPM2_CAMELLIA_KEY_BITS,
    pub sym: TPM2_KEY_BITS,
    pub exclusiveOr: TPMI_ALG_HASH,
}
```

tgonzalezorlandoarm (approved):
LGTM

gowthamsk-arm:
Any idea why the ubuntu test is failing?
```
./configure: line 17494: JSON_C_LIBS: command not found
checking for JSON_C... no
/tmp/rust-tss-esapi/tss-esapi/tests/pkg-config: 4: /tmp/rust-tss-esapi/tss-esapi/tests/pkg-config: SYSROOT: not found
configure: error: Package requirements (json-c) were not met:

/tmp/rust-tss-esapi/tss-esapi/tests/pkg-config: 4: /tmp/rust-tss-esapi/tss-esapi/tests/pkg-config: SYSROOT: not found
No package 'json-c' found
```

This happens for tpm2-tss `2.4.6` cross compilation.
Should I use another docker base image for `ubuntucontainer`?

tgonzalezorlandoarm:
@Superhepper CI is passing now and I think all issues are resolved

gowthamsk-arm:
@ionut-arm @wiktor-k @Superhepper can I get a review, please? :)

Superhepper (approved):
LGTM

PR #445, opened 2023-10-03
Update MSRV 1.66.0

tgonzalezorlandoarm:
Signed-off-by: Tomás González <tomasagustin.gonzalezorlando@arm.com>


Superhepper (approved):
Thanks for the PR.

gowthamsk-arm (approved):
LGTM.

wiktor-k:
Hi, it seems the codespell found some issues, I think they are relatively minor and could just be fixed in a follow-up commit. With one exception: it seems `AcSend` should be spelled like that and maybe it should be added here: https://github.com/parallaxsecond/rust-tss-esapi/blob/main/.codespellrc#L3 (as an exception)

tgonzalezorlandoarm:
Hi @wiktor-k ! Yes, thanks for noticing. I have opened PR https://github.com/parallaxsecond/rust-tss-esapi/pull/448 for this, it has one approval already, could I get you to review that one as well ?

PR #459, opened 2023-10-16
Add definition for UndefineSpaceSpecial

nathanrossi:
Add a implementation for the UndefineSpaceSpecial function which is used to delete an NV index that has the TPMA_NV_POLICY_DELETE attribute set.

An example and test are included for this function, setting up the NV index and deleting it requires additional initialization of a policy digest as well as a policy session in order to perform the delete due to the policy requiring the NvUndefineSpaceSpecial command code defined.

### DIFF
--- a/tss-esapi/src/context/tpm_commands/non_volatile_storage.rs
+++ b/tss-esapi/src/context/tpm_commands/non_volatile_storage.rs
@@ -7,7 +7,7 @@ use crate::{
     structures::{Auth, MaxNvBuffer, Name, NvPublic},
     tss2_esys::{
         Esys_NV_DefineSpace, Esys_NV_Increment, Esys_NV_Read, Esys_NV_ReadPublic,
-        Esys_NV_UndefineSpace, Esys_NV_Write,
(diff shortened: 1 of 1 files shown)


tss-esapi/src/context/tpm_commands/non_volatile_storage.rs
```diff
@@ -226,7 +226,149 @@ impl Context {
...
+    /// Please beware that this method requires both a policy and
+    /// authorization session handle to be present.
+    ///
+    /// # Arguments
+    /// * `nv_auth` - The [Provision] used for authorization.
+    /// * `nv_index_handle`- The [NvIndexHandle] associated with
+    ///                      the nv area that is to be removed.
+    ///
+    /// # Example
+    /// ```rust
```
wiktor-k, line 244: A newline here would be nice :)

wiktor-k (approved):
This is not strictly my area of expertise but I've read the PR and it looks high-quality. Thanks for your time! :+1:

Superhepper (approved):
Thanks for the PR. LGTM.

PR #460, opened 2023-10-18
Add duplication example

Firstyear:
This adds an example of duplication.

As usual - it's broken. I can't seem to work out why, and I was literally copy-pasting as much as possible from this projects own test cases. So any assistance to get this sorted out and over the line would be great.


wiktor-k:
Yep, you're right Ionut. I think one example is right in the doctests of the duplicate function. If there are antly problems digging it out I'll take a look tomorrow morning.

tss-esapi/examples/duplication.rs
```diff
@@ -0,0 +1,427 @@
...
+                storage_key.out_public.clone(),
+            )
+        })
+        .unwrap();
+
+    // Now we can create a child key that we will be able to move along with the parent.
+    //
+    // ALWAYS FAILS!!!
+    // 0x0000018a TpmFormatOneResponseCode { error_number: Type, argument_number: Handle(1) }
+    /*
```
Superhepper, line 222: Is this the actual error you get when you try to display it? If you use master you should get a way more comprehensive error then that if you just take the exception and print it. It should tell you something about which layer that caused the error and so on.

tss-esapi/examples/duplication.rs
```diff
@@ -0,0 +1,427 @@
...
+
+    let storage_public = PublicBuilder::new()
+        .with_public_algorithm(PublicAlgorithm::Ecc)
+        .with_name_hashing_algorithm(HashingAlgorithm::Sha256)
+        .with_object_attributes(object_attributes)
+        // Use policy digest computed using the trial session
+        .with_auth_policy(policy_digest)
+        .with_ecc_parameters(
+            PublicEccParametersBuilder::new()
+                .with_ecc_scheme(EccScheme::Null)
```
Superhepper, line 181: The standard says

> If the object being created is a Storage Key and fixedParent is SET in the attributes of inPublic, then the 
> symmetric  algorithms  and  parameters  of  inPublic  are  required  to  match  those  of  the  parent.
> The algorithms  that  must  match  are  inPublic.nameAlg,  and  the  values  in  inPublic.parameters  that  select  > the symmetric  scheme.  If  inPublic.nameAlg  does  not  match,  the  TPM  shall  return  TPM_RC_HASH.If  the 
> symmetric scheme of the key does not match, the parent, the TPM shall return TPM_RC_SYMMETRIC. 
> The TPM shall not use different response code to differentiate between mismatches of the components of 
> inPublic.parameters. However, after this verification, when using the scheme to encrypt child objects, the 
> TPM ignores the symmetric mode and uses TPM_ALG_CFB.

And you created the primary key with ```.with_symmetric(SymmetricDefinitionObject::AES_128_CFB)``` I think you might need to add that here as well. But I am not sure.

tss-esapi/examples/duplication.rs
Firstyear: When displayed it's `⚠️  TSS Layer: TPM, Code: 0x0000018A, Message: The type of the value is not appropriate for the use (associated with handle number 1).`. Which is still pretty vague. I'm guessing it's something about the storage key and it's attributes that is preventing it from having a child key this type?

tss-esapi/examples/duplication.rs
Firstyear: Yeah, that caused error as well. 

Similar, if I try the `new_restricted_decryption_key` method of PublicEccParametersBuilder it fails too. I'm quite stumped here.  :S

Firstyear: For the record, I tried with swtpm as well, but that was thoroughly unhelpful as every log level just dumps large binary traces rather than anything human readable.

tss-esapi/examples/duplication.rs
Superhepper: Yes, something about how that key is created prevents it from being a storage key.

wiktor-k:
Btw, check out this [tiny example](https://docs.rs/tss-esapi/latest/src/tss_esapi/context/tpm_commands/duplication_commands.rs.html#42-296) (that should be working).

tss-esapi/examples/duplication.rs
Firstyear: Yeah for sure. But whats weird is this is using https://docs.rs/tss-esapi/7.4.0/tss_esapi/structures/struct.PublicEccParametersBuilder.html#method.new_restricted_decryption_key which claims to be valid for a storage key. So clearly there is a bug here in these params and I lack the experience and skills to solve it :(

Firstyear:
I think this example is pretty much the same as the one I copied from in the duplication tests (except that the test uses ecc not rsa https://github.com/parallaxsecond/rust-tss-esapi/blob/main/tss-esapi/tests/integration_tests/context_tests/tpm_commands/duplication_commands_tests.rs ). But it has a significant flaw, an that is that you are duplicating from a primary to the same primary. It's not actually testing or demonstrating moving between two separate keys.

Additionally, the example I'm trying to make is mean to be for simulating or demonstrating if you have a specific duplication key below a primary, rather than duplicating between the primaries. 

So that's part of why I want to get this example to work - Because it certainly looks like it's exposing some bugs or flaws in functions like `new_restricted_decryption_key` ( https://github.com/parallaxsecond/rust-tss-esapi/pull/460/files#diff-aee362f84ca982e9226800f9ca6cec55b4c53d5e846af0bb475beb7d109bc74dR180 ) which per the docs state: "Create parameters for a restricted decryption key (i.e. a storage key)" (https://github.com/parallaxsecond/rust-tss-esapi/blob/main/tss-esapi/src/structures/tagged/public/ecc.rs#L37)

wiktor-k:
Just out of curiosity have you tried using RSA keys instead of ECC to see if the example passes? EC keys may have additional requirements for restricted decryption (like KDF specification). I mean: just to take the `new_restricted_decryption_key` out of the test and see if you can progress. And if so, then fixing `new_restricted_decryption_key`.

Firstyear:
Okay, I sorted out the restricted decryption key issues - turns out it was my own silly fault! 

Anyway, just down to the policy failure with duplication now. Going to re-read your examples to check but I think they're correct.

Firstyear:
@wiktor-k Can I get a re-review on this? Still stuck here with the duplication key :(

wiktor-k:
> @wiktor-k Can I get a re-review on this? Still stuck here with the duplication key :(

Sorry for the delay, you know how it is in free-time open-source side-projects :sweat_smile: 

I'll try to take a look early next week and thanks for the ping! :pray:

Firstyear:
> > @wiktor-k Can I get a re-review on this? Still stuck here with the duplication key :(
> 
> Sorry for the delay, you know how it is in free-time open-source side-projects 😅

Very much so 😅

> I'll try to take a look early next week and thanks for the ping! 🙏

Thank you so much for your help!

tss-esapi/examples/duplication.rs
```diff
@@ -0,0 +1,481 @@
...
+            .map(|_| policy_auth_session)
+            */
+        })
+        .unwrap();
+
+    // ALWAYS FAILS!!!
+    // For some reason this has a policy error - I was copy pasting from the
+    // tests in duplication_commands_tests.rs
+    let public = storage_key.out_public.clone();
+    let (data, duplicate, secret) = context_1
```
wiktor-k, line 359: The only thing that differs is this separation. If both of these blocks are executed in one go, that is the PolicySession setup is followed by the `ctx.duplicate` call the test succeeds and that's how the stuff in other examples is laid out.

Is that sufficient for you to continue or were you asking me to debug it further? (that is: were you asking why just the `policy_auth_session` value is not sufficient for the policy to succeed).

Firstyear: It seems I'd already refactored for that. I've updated the PR, but it still fails.

At this point I really actually feel quite lost and unable to solve this, so as much help as possible would be really appreciated :)

Firstyear:
@wiktor-k Okay I re-read your comments multiple times, I've been reading section 23 about duplication of the spec and I still can not get this to work. I'm completely stumped now what I have done wrong here. I think I really need you or someone else to send me a diff or PR on what to change because I think this is beyond my ability :(

tss-esapi/examples/duplication.rs
Superhepper: The only difference that I have noticed when I looked at other examples is they do not have this "middle" key but uses the primary key directly as a storage key.

Firstyear: But then what would "new parent name" be? A parent can't be duplicated iirc.

ionut-arm: I think I managed to figure out what's going wrong.

You don't need to set the command code, it's done automatically by a successful `PolicyDuplicationSelect`:

```
If the command succeeds, policySession→commandCode is set to TPM_CC_Duplicate.
```

But more importantly, you must only set the policy session on the context as the last step before executing the duplication command. At the moment you have an `load_external_public` in-between which then gets reflected in your policy digest. I'd recommend just running the duplication command under a `ctx.execute_with_session`.

I changed those two things and the example works for me now.

Firstyear: Holy shit, that did it! THANK YOU!!!!!

Firstyear:
@ionut-arm @wiktor-k @Superhepper Okay, after lots of help from all of you this example now works! Thank you all so much! 

I have rebased to main, squashed and signed off. I have also updated the duplication example to be "with_encrypted_duplication(true)" and for bonus, I added a second example which is duplication using inner wrappers only showing how to duplicate against TPM2_RH_NULL.

Superhepper (approved):
Excellent work!

ionut-arm (approved):
🚀

PR #461, opened 2023-10-25
Implement tr_serialize and tr_deserialize

brandsimon:
I don't know if the implementation with `buffer.len().try_into().map_err` and `std::slice::from_raw_parts` is the cleanest solution. Please let me know if you have ideas for cleaner solutions.
Regarding the tests, the `new_handle = old_handle + 1`, but I think checking the public part (which includes the public key) is the most relevant test for it.

Edit:
The test with the new context is needed. Currently it works with the old context, but not with a new one.

### DIFF
--- a/tss-esapi/src/context/general_esys_tr.rs
+++ b/tss-esapi/src/context/general_esys_tr.rs
@@ -1,16 +1,21 @@
 // Copyright 2021 Contributors to the Parsec project.
 // SPDX-License-Identifier: Apache-2.0
 use crate::{
+    constants::tss::TPM2_RH_UNASSIGNED,
     context::handle_manager::HandleDropAction,
+    ffi::to_owned_bytes,
     handles::ObjectHandle,
     handles::{handle_conversion::TryIntoNotNone, TpmHandle},
     structures::Auth,
     structures::Name,
-    tss2_esys::{Esys_TR_Close, Esys_TR_FromTPMPublic, Esys_TR_GetName, Esys_TR_SetAuth},
-    Context, Result, ReturnCode,
+    tss2_esys::{
+        Esys_TR_Close, Esys_TR_Deserialize, Esys_TR_FromTPMPublic, Esys_TR_GetName,
+        Esys_TR_Serialize, Esys_TR_SetAuth,
+    },
+    Context, Error, Result, ReturnCode, WrapperErrorKind,
 };
 use log::error;
-use std::convert::TryFrom;
+use std::convert::{TryFrom, TryInto};
 use std::ptr::null_mut;
 use zeroize::Zeroize;
 
@@ -372,7 +377,7 @@ impl Context {
     #[cfg(has_esys_tr_get_tpm_handle)]
     /// Retrieve the `TpmHandle` stored in the given object.
     pub fn tr_get_tpm_handle(&mut self, object_handle: ObjectHandle) -> Result<TpmHandle> {
-        use crate::{constants::tss::TPM2_RH_UNASSIGNED, tss2_esys::Esys_TR_GetTpmHandle};
+        use crate::tss2_esys::Esys_TR_GetTpmHandle;
         let mut tpm_handle = TPM2_RH_UNASSIGNED;
         ReturnCode::ensure_success(
             unsafe {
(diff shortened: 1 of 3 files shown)


tss-esapi/src/context/general_esys_tr.rs
```diff
@@ -388,6 +392,50 @@ impl Context {
...
+        let mut len = 0;
+        let mut buffer: *mut u8 = null_mut();
+        let mut result = vec![];
+        ReturnCode::ensure_success(
+            unsafe { Esys_TR_Serialize(self.mut_context(), handle.into(), &mut buffer, &mut len) },
+            |ret| {
+                error!("Error while serializing handle: {}", ret);
+            },
+        )?;
+        unsafe {
```
Superhepper, line 406: This feels like it should belong in a function in ```ffi.rs```.

(some psedu rust code):
```rust
pub(crate) fn to_owned_bytes(ffi_bytes_ptr: *mut u8, size: usize) -> Vec<u8> {
    let ffi_bytes = unsafe {MBox::<[u8]>::from_raw_parts(ffi_bytes_ptr, size) };
    return Vec::<u8>::from(ffi_bytes.as_ref())
}
```

And then the Context method would be something like:
```rust
    /// Store the `ObjectHandle` in a buffer
    pub fn tr_serialize(&mut self, handle: ObjectHandle) -> Result<Vec<u8>> {
        let mut len = 0;
        let mut buffer_ptr: *mut u8 = null_mut();
        ReturnCode::ensure_success(
            unsafe { Esys_TR_Serialize(self.mut_context(), handle.into(), &mut buffer_ptr, &mut len) },
            |ret| {
                error!("Error while serializing handle: {}", ret);
            },
        )?;
        Ok(to_owned_buffer(
                buffer_ptr,
                len.try_into().map_err(|e| {
                    error!("Failed to convert buffer len to usize: {}", e);
                    Error::local_error(WrapperErrorKind::InvalidParam)
                })?,
            )
        )
    }
```

tss-esapi/tests/integration_tests/context_tests/general_esys_tr_tests.rs
```diff
@@ -445,3 +447,33 @@ mod test_tr_from_tpm_public {
...
+                None,
+                None,
+            )
+            .unwrap()
+            .key_handle;
+        let data = context.tr_serialize(key_handle.into()).unwrap();
+        // Load handle in a new context
+        let mut new_context = create_ctx_with_session();
+        let new_handle = new_context.tr_deserialize(&data).unwrap().into();
+        assert_eq!(
```
Superhepper, line 474: Have you checked if the TPM just returns the same handle to you? (That would be sort of OK in this specific case but perhaps not what you wanted to test).

Superhepper: Disregard this I read your initial comment regarding this more closely.

Superhepper: This is a little bit dangerous and will not work for all tests. You should probably drop the first context before creating the second otherwise the creation of a context will be blocking forever.

Superhepper: *TSS ESAPI v1p0 r14, section 7.1 Esys_TR_Serialize()*
> NOTE: The auth value has to be set again after deserialization of the object.

I think you need to add the auth to the deserialized handle before reading public?

brandsimon:
@Superhepper 
Thank you very much. Your ffi code works great and looks clean.
I changed the test to persist the handle and drop the old context. If the persistent handle exists, it is cleared before usage.

tss-esapi/src/context/general_esys_tr.rs
```diff
@@ -388,6 +393,44 @@ impl Context {
         TpmHandle::try_from(tpm_handle)
     }
 
-    // Missing function: Esys_TR_Serialize
-    // Missing function: Esys_TR_Deserialize
+    /// Store the `ObjectHandle` in a buffer
```
ionut-arm, line 396: Could you please make these a bit more descriptive, as the name of the function would be somewhat confusing for a library user (since we don't expose the idea of a "TPM Resource" under that name). Perhaps something like:

```
Serialize the metadata of the object identified by `handle` into a new buffer.

This can subsequently be used to recreate the object in a new context in the future.
```

Also, similar request for the deserialize method

ionut-arm (commented):
Thank you for the patch!

tss-esapi/src/context/general_esys_tr.rs
brandsimon: @ionut-arm done

tss-esapi/src/context/general_esys_tr.rs
```diff
@@ -388,6 +393,48 @@ impl Context {
         TpmHandle::try_from(tpm_handle)
     }
 
-    // Missing function: Esys_TR_Serialize
-    // Missing function: Esys_TR_Deserialize
+    /// Serialize the metadata of the object identified by `handle` into a new buffer.
+    ///
+    /// This can subsequently be used to recreate the object in a new context in the future.
```
Superhepper, line 398: That is a good start!

But what I think he meant was that if you look at how the other context methods are documented. They have a description of the arguments to the method, a description of what the method returns, a description of errors that might be thrown (if any) and an example of how to use the method.

brandsimon: Thank you, done. :)

brandsimon:
@ionut-arm @Superhepper 
I added descriptions and examples.

Superhepper (approved):
LGTM!

Thank you for your hard work.

tss-esapi/tests/integration_tests/context_tests/general_esys_tr_tests.rs
```diff
@@ -445,3 +448,45 @@ mod test_tr_from_tpm_public {
...
+                None,
+                None,
+            )?
+            .key_handle;
+        let public = context.read_public(key_handle)?;
+        let persistent_handle =
+            context.evict_control(Provision::Owner, key_handle.into(), persistent)?;
+        let data = context.tr_serialize(persistent_handle)?;
+
+        std::mem::drop(context);
```
wiktor-k, line 484: Just `drop` will be sufficient (I think it's part of the prelude).

tss-esapi/src/context/general_esys_tr.rs
```diff
@@ -388,6 +393,147 @@ impl Context {
...
+        ))
+    }
+
+    /// Deserialize the metadata from `buffer` into a new object.
+    ///
+    /// This can be used to restore an object from a context in the past.
+    ///
+    /// # Arguments
+    /// * `buffer` - The buffer containing the data to restore the object.
+    ///              It can be created using `tr_serialize`.
```
wiktor-k, line 471: `tr_serialize` can be enclosed in square brackets. This has a minor benefit that it creates a clickable link in the documentation.

wiktor-k (approved):
I think it looks good 👍 I found only very minor things.

tss-esapi/src/context/general_esys_tr.rs
brandsimon: @wiktor-k Thank you, done

Superhepper (approved):

PR #462, opened 2023-10-25
Sbr/serialize2

brandsimon:


brandsimon:
Ups, I did not want to create a second PR here.

PR #464, opened 2023-10-26
Fix create_ak for ECC keys

ionut-arm:
Fixing the parameters for creating AKs in the Endorsement Hierarchy. The `count` value part of the `EccScheme` has been adjusted, and an empty `EccPoint` was added as unique identifier for the key.

### DIFF
--- a/tss-esapi/tests/integration_tests/abstraction_tests/ak_tests.rs
+++ b/tss-esapi/tests/integration_tests/abstraction_tests/ak_tests.rs
@@ -15,6 +15,7 @@ use tss_esapi::{
         session_handles::PolicySession,
     },
     structures::{Auth, Digest, PublicBuilder, SymmetricDefinition},
+    Error, WrapperErrorKind,
 };
 
 use crate::common::create_ctx_without_session;
@@ -51,7 +52,7 @@ fn test_create_ak_rsa_ecc() {
         None,
     )
     .unwrap();
-    if ak::create_ak(
+    if let Err(Error::WrapperError(WrapperErrorKind::InconsistentParams)) = ak::create_ak(
         &mut context,
         ek_rsa,
         HashingAlgorithm::Sha256,
@@ -59,16 +60,16 @@ fn test_create_ak_rsa_ecc() {
         SignatureSchemeAlgorithm::Sm2,
         None,
         None,
-    )
-    .is_ok()
-    {
-        // We can't use unwrap_err because that requires Debug on the T
-        panic!("Should have errored");
+    ) {
+    } else {
+        panic!(
+            "Should've gotten an 'InconsistentParams' error when trying to create an a P256 AK with an SM2 signing scheme."
+        );
     }
 }
 
 #[test]
-fn test_create_ak_ecc_ecc() {
+fn test_create_ak_ecc() {
     let mut context = create_ctx_without_session();
 
     let ek_ecc = ek::create_ek_object(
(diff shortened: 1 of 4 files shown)


wiktor-k (dismissed):
Looks good :+1: 

Is there a test case for `EcDaa` already? (so that both codepaths are tested).

ionut-arm:
I'm curious now why that `ecc_rsa` test was supposed to fail 🤔

wiktor-k:
> EDIT: nevermind, noticed the scheme... why does it work with swtpm, though....

I think a better explanation in the test would be a good idea. Also, instead of manual panic: [`#[should_panic]`](https://doc.rust-lang.org/book/ch11-01-writing-tests.html#checking-for-panics-with-should_panic) maybe even with `expected`...

Just throwing some random work ideas your way :sweat_smile:

Superhepper (approved):

tss-esapi/tests/integration_tests/abstraction_tests/ak_tests.rs
```diff
@@ -42,6 +42,7 @@ fn test_create_ak_rsa_rsa() {
 }
 
 #[test]
+#[should_panic]
```
Superhepper, line 45: I don't understand this panic. Should you not test that the correct error is generated instead?

wiktor-k: I guess Ionut did so because I suggested it in a previous comment. Sorry 🙇‍♂️! Maybe indeed `assert!(matches!(...` would be a better option 🤔

Superhepper: I mean for me ```#[should_panic]``` is used to test that the library panics under certain conditions. And we try very hard to not panic any where in the library so that should hopefully not happen.

Things that one can do to test is:
1. If the error in it self is not that import but rather that an error is returned at all then you can:
```rust
let result = ak::foo();
assert!(result.is_err(), "Function fo should have returned an error.");
```
This is not done any where in our code currently because the returned errors are a part of the API.

2. The specific error is important:
```rust
let result = ak::foo();
if let Err(actual_error) = result {
    assert_eq!(
        Error::WrapperError(WrapperErrorKind::InvalidParam),
        actual_error,
        "Foo did not produce the expected error."
    );
} else {
    panic!("Foo is expected to return an error");
}
```
This is done extensively when testing the TSS return code errors.

ionut-arm: Thanks for the pushback, should've thought this through 😅 

I'll tweak it and push back up.

Firstyear:
What should we do to unblock this pr?

Superhepper:
I do not think it is blocked I just think @ionut-arm has been busy and not been able to work on this.

Superhepper:
@ionut-arm Tell me if you need any help with this. I can fix the tests for you if you like.

tss-esapi/tests/integration_tests/abstraction_tests/ak_tests.rs
ionut-arm: Right, only took about 3 months, hope it works!

ionut-arm:
> @ionut-arm Tell me if you need any help with this. I can fix the tests for you if you like.

ok, finally got to this... @Firstyear - apologies for the long wait, I reckon this is ready to go (pending approval, of course)

Superhepper (approved):

Firstyear (approved):

hug-dev (approved):

PR #466, opened 2023-11-03
Add serde to Public and Private

Firstyear:
Add support for serde to Public and Private, going through the current Marshall and Unmarshall pathways. This pattern is quite "simple" so it could be possible to duplicate it to other structures, or macro them for types that implement Marshall and Unmarshall.

### DIFF
--- a/tss-esapi/src/structures/buffers/private.rs
+++ b/tss-esapi/src/structures/buffers/private.rs
@@ -2,8 +2,34 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::traits::impl_mu_standard;
+use crate::traits::{Marshall, UnMarshall};
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use tss_esapi_sys::_PRIVATE;
 
 buffer_type!(Private, ::std::mem::size_of::<_PRIVATE>(), TPM2B_PRIVATE);
 
 impl_mu_standard!(Private, TPM2B_PRIVATE);
+
+impl Serialize for Private {
+    /// Serialise the [Private] data into it's bytes representation of the TCG
+    /// TPM2B_PRIVATE structure.
+    fn serialize<S>(&self, serializer: S) -> std::result::Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let bytes = self.marshall().map_err(serde::ser::Error::custom)?;
+        serializer.serialize_bytes(&bytes)
+    }
+}
+
+impl<'de> Deserialize<'de> for Private {
+    /// Deserialise the [Private] data from it's bytes representation of the TCG
+    /// TPM2B_PRIVATE structure.
+    fn deserialize<D>(deserializer: D) -> std::result::Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let bytes = <Vec<u8>>::deserialize(deserializer)?;
+        Self::unmarshall(&bytes).map_err(serde::de::Error::custom)
+    }
+}
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,3 +1,3 @@
 [codespell]
 skip = .git,target,Cargo.lock
-ignore-words-list = acsend,crate,keypair,daa
+ignore-words-list = acsend,crate,keypair,daa,de,ser
--- a/tss-esapi/Cargo.toml
+++ b/tss-esapi/Cargo.toml
@@ -37,6 +37,7 @@ paste = "1.0.14"
 [dev-dependencies]
 env_logger = "0.9.0"
 sha2 = "0.10.1"
+serde_json = "^1.0.108"
 
 [build-dependencies]
 semver = "1.0.7"
--- a/tss-esapi/src/structures/tagged/public.rs
+++ b/tss-esapi/src/structures/tagged/public.rs
@@ -8,7 +8,7 @@ use crate::{
     attributes::ObjectAttributes,
     interface_types::algorithm::{HashingAlgorithm, PublicAlgorithm},
     structures::{Digest, EccPoint, PublicKeyRsa, SymmetricCipherParameters},
-    traits::{impl_mu_standard, Marshall},
+    traits::{impl_mu_standard, Marshall, UnMarshall},
     tss2_esys::{TPM2B_PUBLIC, TPMT_PUBLIC},
     Error, Result, ReturnCode, WrapperErrorKind,
 };
@@ -18,6 +18,7 @@ use keyed_hash::PublicKeyedHashParameters;
 use rsa::PublicRsaParameters;
 
 use log::error;
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use std::convert::{TryFrom, TryInto};
 use tss_esapi_sys::{TPMU_PUBLIC_ID, TPMU_PUBLIC_PARMS};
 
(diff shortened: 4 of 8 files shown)


tss-esapi/src/structures/buffers/private.rs
```diff
@@ -7,9 +7,36 @@ use crate::{
...
 buffer_type!(Private, ::std::mem::size_of::<_PRIVATE>(), TPM2B_PRIVATE);
 
+impl Serialize for Private {
+    fn serialize<S>(&self, serializer: S) -> std::result::Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let bytes = self.marshall()
+            .map_err(|err| {
+                serde::ser::Error::custom(err)
```
wiktor-k, line 21: Hmmm... annoying codespell... you'll need to add `ser` to `.codespellrc`

Superhepper:
I think it would be a good idea to make serde support as an optional feature.

Firstyear:
> I think it would be a good idea to make serde support as an optional feature.

Serde is already hard-required, so why make it optional?

tss-esapi/src/structures/buffers/private.rs
Firstyear: Will do :)

Superhepper:
Oh, you are right. That was unfortunate. Then it will not matter for this PR but rather something we should look into doing. Because it is nice in my opinion to make it possible for the user to opt out of non essential functionality.

wiktor-k:
Agreed with @Superhepper, serde should be optional but it's not an issue with this PR.

tss-esapi/src/structures/buffers/private.rs
```diff
@@ -7,9 +7,36 @@ use crate::{
...
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
 buffer_type!(Private, ::std::mem::size_of::<_PRIVATE>(), TPM2B_PRIVATE);
 
+impl Serialize for Private {
+    fn serialize<S>(&self, serializer: S) -> std::result::Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let bytes = self.marshall()
```
ionut-arm, line 19: Given that this just relies on `Marshall` underneath, wouldn't it make sense to just implement these two traits generically for all types `T: Marhsall + UnMarshall`?

tss-esapi/src/structures/buffers/private.rs
```diff
@@ -7,9 +7,36 @@ use crate::{
 };
 use std::convert::TryInto;
 use tss_esapi_sys::_PRIVATE;
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
 buffer_type!(Private, ::std::mem::size_of::<_PRIVATE>(), TPM2B_PRIVATE);
 
+impl Serialize for Private {
+    fn serialize<S>(&self, serializer: S) -> std::result::Result<S::Ok, S::Error>
```
ionut-arm, line 15: Could we add some doc lines here to explain that the serialization is using the TCG-defined serialization formats?

ionut-arm (commented):
Thanks for looking into implementing this!

tss-esapi/src/structures/buffers/private.rs
Firstyear: I suspect that if we make it generic in that way then you break users who have their own generics relying on Serialise/Deserialise since then we'd have a Serialise  + Marshall bound?

tss-esapi/src/structures/buffers/private.rs
Firstyear: Yep, no issue.

tss-esapi/src/structures/buffers/private.rs
ionut-arm: Since you can only implement traits for types defined in your crate, they'd only be able to implement it on their own types. I'm not really sure Marshall/UnMarshall make sense for types outside of our crates, tbh, and maybe we should document it as such. Also since this change will be going out in a major change, it won't break someone's code all of a sudden without an intentional update to their dependencies.

Firstyear: Okay, so is there anything actionable here then? I think the way it is now is fine, we could always macro it to save boiler plate I guess?

ionut-arm: I think it can go in as is, we'll maybe add the generic implementation down the line.

ionut-arm (commented):
Happy for this to go in, but maybe a couple of tests for (de)serialization would be good to catch any issues down the line! Also, the formatter seems to be complaining

Firstyear:
Looks like I stepped on a land-mine adding these tests.

Following the pattern of the marshall_unmarshall test I added these with a reflexive test using serde_json. serde_json was added as a dev-dependency.

This caused more than 1500 compiler errors - it turns out that once serde_json exists, this confuses the compiler about the partialEq trait in the macros "test_valid_conversion" used throughout the test suite. Because of this, assert_eq! fails to work because there are multiple possible satisfying types for the ::into calls.

An example is:

```
error[E0283]: type annotations needed
  --> /home/886b45ce-98b2-4d79-9069-c9ff25bc0232/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/macros/mod.rs:40:32
   |
36 |   macro_rules! assert_eq {
   |   ---------------------- in this expansion of `assert_eq!` (#2)
...
40 |                   if !(*left_val == *right_val) {
   |                                  ^^ cannot infer type
   |
  ::: tss-esapi/tests/integration_tests/constants_tests/algorithm_tests.rs:17:1
   |
17 | / macro_rules! test_conversion {
18 | |     ($tpm_alg_id:ident, $algorithm:ident) => {
19 | |         assert_eq!($tpm_alg_id, AlgorithmIdentifier::$algorithm.into());
   | |         --------------------------------------------------------------- in this macro invocation (#2)
20 | |         assert_eq!(
...  |
27 | |     };
28 | | }
   | |_- in this expansion of `test_conversion!` (#1)
...
32 |       test_conversion!(TPM2_ALG_CAMELLIA, Camellia);
   |       --------------------------------------------- in this macro invocation (#1)
   |
   = note: multiple `impl`s satisfying `u16: PartialEq<_>` found in the following crates: `core`, `serde_json`:
           - impl PartialEq for u16;
           - impl PartialEq<serde_json::Value> for u16;

```

I think this is beyond my ability to solve this, and probably needs some bigger work to clean up.

Firstyear:
In light of the above issue, what do you think I should do here @ionut-arm ?

Superhepper:
Have you tried something trivial like:
```rust
if !(u16::from(*left_val) == u16::from(*right_val)) {
```

And see if it helps?

Firstyear:
> Have you tried something trivial like:
> 
> ```rust
> if !(u16::from(*left_val) == u16::from(*right_val)) {
> ```
> 
> And see if it helps?

There are hundreds of locations that need this update. I'm not doing it in this PR.

Superhepper:
I can help  you fix some of them.

Firstyear:
Thanks to @Superhepper I was now able to add the serde tests :)

Firstyear:
Hey all, what else is needed here? We really want this merged so that we can use it for TPM support for Azure AD and other linux pam authentication subsystems.

Superhepper:
Not anything I think. I would just want CI to work again so this can without getting some unrelated errors.

Superhepper (approved):

Superhepper:
Try to rebase this on main now. The CI should be fine.

wiktor-k (dismissed):
Just a couple of tiny lints remain. :+1:

Firstyear:
All fixed :)

Superhepper (approved):

wiktor-k (approved):
Looks straightforward. Thanks!

Firstyear:
Thank you all so much!

PR #471, opened 2023-11-08
Improves handling ```size_t``` FFI values.

Superhepper:
- The size_t type that is used in all the MU APIs will be dependent on the compiler that ```bindgen``` uses to generate the bindings for the library. This attempts to make the crate less dependant on that value in the public APIs.

- Changes the Marshall and Unmarshall traits to use usize as input.


tss-esapi/src/ffi.rs
```diff
@@ -26,7 +27,60 @@ where
...
+///
+/// # Returns
+/// The owned bytes in the form of a `Vec<u8>` object.
 pub(crate) fn to_owned_bytes(ffi_bytes_ptr: *mut u8, size: usize) -> Vec<u8> {
     let ffi_bytes = unsafe { MBox::<[u8]>::from_raw_parts(ffi_bytes_ptr, size) };
     return Vec::<u8>::from(ffi_bytes.as_ref());
 }
+
+/// Type used for handling `size_t` variables
+pub(crate) struct FfiSizeType(crate::tss2_esys::size_t);
```
ionut-arm, line 45: I'm wondering - would we ever want to expose anything from this module to the outside world? If not, them maybe we can just tag the entire module `pub(crate)`, then we don't need to worry about having everything in here `pub(crate)` (?)

tss-esapi/src/interface_types/structure_tags.rs
```diff
@@ -78,54 +79,44 @@ impl TryFrom<TPMI_ST_ATTEST> for AttestationType {
 impl Marshall for AttestationType {
     const BUFFER_SIZE: usize = std::mem::size_of::<TPMI_ST_ATTEST>();
 
-    fn marshall_offset(
-        &self,
-        marshalled_data: &mut [u8],
-        offset: &mut std::os::raw::c_ulong,
-    ) -> Result<()> {
+    fn marshall_offset(&self, marshalled_data: &mut [u8], offset: &mut usize) -> Result<()> {
+        let ffi_buffer_size = FfiSizeType::try_from(marshalled_data.len())?;
```
ionut-arm, line 83: This is going to be such a common pattern - what do you think about implementing a macro in `ffi.rs` to go from `&mut [u8]` and `&mut usize` to `size_t`, via `FfiSizeType`? And you can just use that within the MU call.

ionut-arm (commented):
Thanks for the patch, looks good so far!

tss-esapi/src/interface_types/structure_tags.rs
Superhepper: I think the pattern is so common it would probably be possible the entire implementation of marshall and unmarshall in a macro.

tss-esapi/src/ffi.rs
Superhepper: Yes, the FFI module will probably only be used internally.

Superhepper: Actually it already is ...
https://github.com/parallaxsecond/rust-tss-esapi/blob/main/tss-esapi/src/lib.rs#L116

tss-esapi/src/interface_types/structure_tags.rs
Superhepper: I think I will look into if it is possible to just do a proc_macro
```rust
#[derive(marshall(TPM2B_DIGEST), unmarshall(TPM2B_DIGEST)]
```

tss-esapi/src/ffi.rs
Superhepper: When some of the ```pub(crate)``` I got some errors about leaking private stuff.

tss-esapi/src/structures/buffers/private.rs
```diff
@@ -1,69 +1,57 @@
...
-                error!("Failed to marshal Private: {}", ret);
-            },
-        )?;
+// impl Marshall for Private {
+//     const BUFFER_SIZE: usize = std::mem::size_of::<TPM2B_PRIVATE>();
 
-        Ok(())
-    }
-}
+//     /// Produce a marshalled [`TPM2B_PRIVATE`]
```
wiktor-k, line 14: I wonder if that commented out code can just be removed. If it's needed we have it in git anyway. (also the same for other commented out code here)

wiktor-k (approved):
I think in general it looks good and a lot of code got cleaner. I'm not a fan if leaving the commented-out code though but besides that :+1: Thanks!

Superhepper:
@wiktor-k 
Haha sorry I will remove that. I needed it as a reference when I write the macros.

tss-esapi/src/structures/buffers/private.rs
Superhepper: Fixed

tss-esapi/src/structures/buffers/public.rs
```diff
@@ -111,52 +112,40 @@ impl TryFrom<Public> for PublicBuffer {
 impl Marshall for PublicBuffer {
     const BUFFER_SIZE: usize = std::mem::size_of::<TPM2B_PUBLIC>();
 
-    /// Produce a marshalled [`TPM2B_PUBLIC`]
-    fn marshall(&self) -> Result<Vec<u8>> {
-        let mut buffer = vec![0; Self::BUFFER_SIZE];
-        let mut offset = 0;
-
+    fn marshall_offset(&self, marshalled_data: &mut [u8], offset: &mut usize) -> Result<()> {
```
ionut-arm, line 115: This one doesn't fit the macro patterns?

tss-esapi/src/traits.rs
```diff
@@ -51,65 +38,151 @@ pub trait UnMarshall: Sized {
...
 
-impl UnMarshall for u32 {
-    fn unmarshall_offset(
-        marshalled_data: &[u8],
-        offset: &mut std::os::raw::c_ulong,
-    ) -> Result<Self> {
-        let mut dest = 0_u32;
+/// Macro used to implement Marshall and Unmarshall for types that
+/// can be converted from native to TSS i.e. it cannot fail and are
+/// passed to MUAPI by value(i.e. pointer).
```
ionut-arm, line 128: `by value (i.e. pointer)` - is this correct?

tss-esapi/src/traits.rs
```diff
@@ -51,65 +38,151 @@ pub trait UnMarshall: Sized {
...
+            $native_type::try_from
+        );
+    };
+    ($native_type:ident, $tss_type:ident) => {
+        $crate::traits::impl_mu_complex!($native_type, $tss_type, $tss_type);
+    };
 }
+
+// Make the macros usable outside of the module.
+pub(crate) use impl_marshall_trait;
```
ionut-arm, line 181: Is this actually imported somewhere else?

ionut-arm (commented):
Thanks for the effort! I hate how similar, yet just different enough the implementations of Marshall and UnMarshall have to be...

tss-esapi/src/traits.rs
Superhepper: No, but I thought it could be useful if you for some reason only wanted to implement one of the traits or there was a case where the implementation did not fit the other macros.

tss-esapi/src/traits.rs
Superhepper: Fixed

tss-esapi/src/structures/buffers/public.rs
Superhepper: Fixed

wiktor-k (approved):

hug-dev (approved):

PR #474, opened 2023-11-25
Fixes some causes for type ambiguity in tests.

Superhepper:
- Many of tests uses macro to create the test code and in this test code it is common practice to use ```into``` or ```try_into``` conversions. But this can cause ambiguities later on if new types introduced implements traits that previously only was implemented by one type. So in order to avoid this kind of ambiguity some of the conversions have been changed. This only a first step of changes that probably needs to be done. This fixes the errors that occurs if one tries to add ```serde_json``` to the ```cargo.yaml``` which is probably needed in order to be able to write tests for #466.


wiktor-k (approved):
Titanic effort but I like it, especially that some stuff like `convert_to_u32_test` will now give better error messages :+1:

hug-dev (approved):

PR #476, opened 2023-11-28
Fixes problem with sign requiring a HashcheckTicket.

### LINKED ISSUE #475: No way to sign without hashcheck ticket
https://docs.rs/tss-esapi/latest/tss_esapi/struct.Context.html#method.sign requires a hashcheck ticket, however an unrestricted key should not *need* the hashcheck. Should there be two sign apis, one for restricted_sign and one for unrestricted? 

Or am I overlooking something obvious here?

Superhepper:
- This fixes #475 by making the HashcheckTicket that was previously required in the ```sign``` context method optional instead. If it is ```None``` it is then internally converted into the HashcheckTicket version of the `Null ticket` before being converted to the corresponding TSS type. This has the benefit of removing the need to use the TSS type in order to create a `Null ticket`.

### DIFF
--- a/tss-esapi/src/context/tpm_commands/signing_and_signature_verification.rs
+++ b/tss-esapi/src/context/tpm_commands/signing_and_signature_verification.rs
@@ -45,9 +45,10 @@ impl Context {
         key_handle: KeyHandle,
         digest: Digest,
         scheme: SignatureScheme,
-        validation: HashcheckTicket,
+        validation: Option<HashcheckTicket>,
     ) -> Result<Signature> {
         let mut signature_ptr = null_mut();
+        let validation_ticket = validation.unwrap_or_default().try_into()?;
         ReturnCode::ensure_success(
             unsafe {
                 Esys_Sign(
@@ -58,7 +59,7 @@ impl Context {
                     self.optional_session_3(),
                     &digest.into(),
                     &scheme.into(),
-                    &validation.try_into()?,
+                    &validation_ticket,
                     &mut signature_ptr,
                 )
             },
--- a/tss-esapi/src/context/tpm_commands/signing_and_signature_verification.rs
+++ b/tss-esapi/src/context/tpm_commands/signing_and_signature_verification.rs
@@ -40,15 +40,66 @@ impl Context {
     }
 
     /// Sign a digest with a key present in the TPM and return the signature.
+    ///
+    /// # Details
+    /// For signatures using a restricted key, a hashcheck must be provided. For unrestricted keys, this may be None.
+    ///
+    /// # Parameters
+    /// `key_handle` - Handle to the key be used for signing.
+    /// `digest`     - The digest that is going to be signed.
+    /// `scheme`     - The scheme to use if the scheme for the key referenced by the key handle is null.
+    /// `validation` - An optional [HashcheckTicket] that proof that the digest was created by the TPM.
+    ///                N.B. None will be treated as a "Null ticket".
+    /// # Example
+    ///
+    /// ```rust
+    /// # use tss_esapi::{Context, TctiNameConf,
+    /// #    interface_types::{
+    /// #        algorithm::{HashingAlgorithm, RsaSchemeAlgorithm},
+    /// #        key_bits::RsaKeyBits,
+    /// #        resource_handles::Hierarchy,
+    /// #    },
+    /// #    structures::{RsaScheme, RsaExponent},
+    /// #    utils::create_unrestricted_signing_rsa_public
+    /// # };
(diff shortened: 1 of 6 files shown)


Firstyear (commented):
I can't see the changes to `fn sign` here, were they committed?

tss-esapi/src/context/tpm_commands/signing_and_signature_verification.rs
```diff
@@ -45,9 +45,10 @@ impl Context {
         key_handle: KeyHandle,
         digest: Digest,
         scheme: SignatureScheme,
-        validation: HashcheckTicket,
+        validation: Option<HashcheckTicket>,
```
wiktor-k, line 48: I think it looks good but just for the sake of completeness, to avoid API breaks you could make `validation: impl Into<Option<HashcheckTicket>>` and below add a call to `.into()` before `.unwrap_or_default()`.

This would allow clients to call this function with `HashcheckTickets` like previously (no modification needed, no Some wrapping) but would also accept `None`.

An example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f509608362b096375d913dbed0a29a88

wiktor-k (approved):
This makes a lot of calling code cleaner. Thanks! :+1:

wiktor-k:
> I can't see the changes to `fn sign` here, were they committed?

I think you were looking for [this line of code](https://github.com/parallaxsecond/rust-tss-esapi/pull/476/files#diff-b349ca1f38d7200781cc5f9e14f2cd3cfefa3bc5ad07a6f406248ae2b9643ed4R43).

tss-esapi/src/context/tpm_commands/signing_and_signature_verification.rs
Superhepper: That is neat I added it.

wiktor-k (approved):

tss-esapi/src/context/tpm_commands/signing_and_signature_verification.rs
```diff
@@ -45,9 +45,10 @@ impl Context {
         key_handle: KeyHandle,
         digest: Digest,
         scheme: SignatureScheme,
-        validation: HashcheckTicket,
+        validation: impl Into<Option<HashcheckTicket>>,
```
ionut-arm, line 48: I guess since the docs on this method are fairly lightweight, there's no point in specifying what happens if one passes `None` here?

ionut-arm (approved):
Awesome, thanks!

tss-esapi/src/context/tpm_commands/signing_and_signature_verification.rs
Firstyear: It's probably better to say something here since we're looking. Maybe something like:

"For signatures using a restricted key, a hashcheck must be provided. For unrestricted keys, this may be None"

Superhepper: Yes, I should probably also update the documentation of the sign context method while I am at because it was seriously lacking.

Superhepper: Please have a look at the documentation that I added and see if it seems alright.

Firstyear (approved):
Looks good mate!

wiktor-k (dismissed):
LGTM :+1: thanks!

This got me thinking that maybe we should review the other parameters and allow `None`s there too :sweat_smile:

Superhepper:
> LGTM 👍 thanks!
> 
> This got me thinking that maybe we should review the other parameters and allow `None`s there too 😅

I am not so sure about that. The other parameters have actual ```Null``` options which the ticket didn't.

tss-esapi/src/structures/tickets.rs
```diff
@@ -124,6 +127,17 @@ pub struct HashcheckTicket {
     digest: Vec<u8>,
 }
 
+impl Default for HashcheckTicket {
+    /// The default for the Hashcheck ticket is the Null ticket.
+    fn default() -> Self {
+        Self {
+            tag: StructureTag::Hashcheck,
+            hierarchy: Hierarchy::Null,
+            digest: Vec::<u8>::new(),
```
wiktor-k, line 136: :thinking: This is minor but I think the `::<u8>` part is redundant.

wiktor-k (approved):
LGTM :+1:

Superhepper:
> Awesome, thanks!

Could you also add a review for this?

ionut-arm (approved):
LGTM, thanks for the documentation

PR #477, opened 2023-12-09
Adds support for non pkg-config tpm2-tss installs.

Superhepper:
This adds support for specifying an install path of tpm2-tss libraries and header files in order to make it possible to build crate without having to relay on pkg-config.

### DIFF
--- a/tss-esapi-sys/build.rs
+++ b/tss-esapi-sys/build.rs
@@ -1,114 +1,515 @@
 // Copyright 2021 Contributors to the Parsec project.
 // SPDX-License-Identifier: Apache-2.0
 
-#[cfg(feature = "generate-bindings")]
-use std::path::PathBuf;
-
-const MINIMUM_VERSION: &str = "3.2.2";
-
 fn main() {
     if std::env::var("DOCS_RS").is_ok() {
         // Nothing to be done for docs.rs builds.
         return;
     }
 
-    #[cfg(feature = "generate-bindings")]
-    {
-        let out_path = std::path::PathBuf::from(std::env::var("OUT_DIR").unwrap());
-        let esys_path = out_path.join("tss_esapi_bindings.rs");
-        generate_from_system(esys_path);
-    }
-
-    #[cfg(not(feature = "generate-bindings"))]
-    {
-        use std::str::FromStr;
-        use target_lexicon::{Architecture, OperatingSystem, Triple};
-
-        let target = Triple::from_str(&std::env::var("TARGET").unwrap())
-            .expect("Failed to parse target triple");
-        match (target.architecture, target.operating_system) {
-            (Architecture::Arm(_), OperatingSystem::Linux) => {}
-            (Architecture::Aarch64(_), OperatingSystem::Linux) => {}
-            (Architecture::X86_64, OperatingSystem::Darwin) => {}
-            (Architecture::X86_64, OperatingSystem::Linux) => {}
-            (arch, os) => {
-                panic!("Compilation target (architecture, OS) tuple ({}, {}) is not part of the supported tuples. Please compile with the \"generate-bindings\" feature or add support for your platform :)", arch, os);
-            }
+    cfg_if::cfg_if! {
+        if #[cfg(feature = "generate-bindings")] {
+            let installation = tpm2_tss::Installation::probe(true);
+            let out_dir = std::path::PathBuf::from(std::env::var("OUT_DIR").unwrap());
+            installation.generate_bindings(&out_dir.join("tss_esapi_bindings.rs"));
+        } else {
+            target::ensure_supported();
+            let _ = tpm2_tss::Installation::probe(false);
         }
-
-        pkg_config::Config::new()
-            .atleast_version(MINIMUM_VERSION)
-            .probe("tss2-sys")
-            .expect("Failed to find tss2-sys library.");
-        let tss2_esys = pkg_config::Config::new()
-            .atleast_version(MINIMUM_VERSION)
-            .probe("tss2-esys")
-            .expect("Failed to find tss2-esys library.");
-        pkg_config::Config::new()
-            .atleast_version(MINIMUM_VERSION)
-            .probe("tss2-tctildr")
-            .expect("Failed to find tss2-tctildr library.");
-        pkg_config::Config::new()
-            .atleast_version(MINIMUM_VERSION)
-            .probe("tss2-mu")
-            .expect("Failed to find tss2-mu library.");
-
-        println!("cargo:version={}", tss2_esys.version);
     }
 }
 
-#[cfg(feature = "generate-bindings")]
-pub fn generate_from_system(esapi_out: PathBuf) {
-    pkg_config::Config::new()
-        .atleast_version(MINIMUM_VERSION)
-        .probe("tss2-sys")
-        .expect("Failed to find tss2-sys library.");
-    let tss2_esys = pkg_config::Config::new()
-        .atleast_version(MINIMUM_VERSION)
-        .probe("tss2-esys")
-        .expect("Failed to find tss2-esys");
-    let tss2_tctildr = pkg_config::Config::new()
-        .atleast_version(MINIMUM_VERSION)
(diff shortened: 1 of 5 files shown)


wiktor-k:
Very interesting development!

By the way Github supports windows CI runners so maybe it would be useful to at least run `cargo build` there?

Superhepper:
Yeah this is just a test. I would really like to be able run the tests before committing anything.

wiktor-k:
I've just recently added it to my other crate, maybe you can take a look here: https://github.com/wiktor-k/ssh-agent-lib/blob/main/.github/workflows/ci.yml#L58

I'm using matrix there to build on several OSes at once but you can just use windows-latest to write a job for windows only.

You might have to use vcpkg for tss libs (I'm not an expert in this matter) like openssl bindings do: https://github.com/sfackler/rust-openssl/blob/master/.github/workflows/ci.yml#L95

Hope this helps!

Superhepper:
> I've just recently added it to my other crate, maybe you can take a look here: https://github.com/wiktor-k/ssh-agent-lib/blob/main/.github/workflows/ci.yml#L58
> 
> I'm using matrix there to build on several OSes at once but you can just use windows-latest to write a job for windows only.
> 
> You might have to use vcpkg for tss libs (I'm not an expert in this matter) like openssl bindings do: https://github.com/sfackler/rust-openssl/blob/master/.github/workflows/ci.yml#L95
> 
> Hope this helps!

Unfortunately tpm2-tss did not have a vcpkg. Will have to explore other options.

wiktor-k:
Oh, that's too bad. I see people are struggling with compilation on Windows: https://github.com/tpm2-software/tpm2-tss/issues/1828 but maybe you'll find a way :crossed_fingers:

Superhepper:
I have already compiled tpm2-tss on my computer(Windows) and got the tests of the crate to run. They fail because I have no simulator but they run.

tss-esapi-sys/build.rs
```diff
@@ -1,114 +1,487 @@
...
+                        .clang_arg(self.tss2_tctildr.include_dir_arg())
+                        .clang_arg(self.tss2_mu.include_dir_arg())
+                        .formatter(bindgen::Formatter::Rustfmt)
+                        .header(self.tss2_esys.header_file_arg())
+                        .header(self.tss2_tctildr.header_file_arg())
+                        .header(self.tss2_mu.header_file_arg())
+                        //See this issue: https://github.com/parallaxsecond/rust-cryptoki/issues/12
+                        .generate_comments(false)
+                        .blocklist_type("max_align_t")
+                        // Needed for windows
```
ionut-arm, line 132: What was the issue here?

tss-esapi-sys/build.rs
```diff
@@ -1,114 +1,487 @@
...
+        /// # Arguments
+        /// env_var - The value of the environment variable that contains the installation path.
+        ///
+        /// # Returns
+        /// A tuple containing the validated installation path and the version associated with it.
+        fn ensure_valid_installation_path(env_var: String) -> (PathBuf, String) {
+            let install_path = PathBuf::from(env_var);
+            if !install_path.is_dir() {
+                panic!(
+                    "`{}` specifies a path `{}`, that does not exist",
```
ionut-arm, line 194: Isn't this somewhat unrepresentative of the check before - if `env_var` points at a file instead of a dir, this panic will be triggered.

tss-esapi-sys/build.rs
```diff
@@ -1,114 +1,487 @@
...
+                .expect("Failed to parse the content of the VERSION file.");
+            let min_version_req_str = format!(">={}", MINIMUM_VERSION);
+            let min_version_req = VersionReq::parse(&min_version_req_str)
+                .expect("Failed to parse minimum version requirement.");
+            if !min_version_req.matches(&version) {
+                panic!(
+                    "Minimum supported version is {}, found installed version {}",
+                    MINIMUM_VERSION, version_str
+                );
+            }
```
ionut-arm, line 225: Could you also add a bit in these expect strings to make it clear which version we mean? Something like `Failed to read tpm2-tss VERSION file`.

tss-esapi-sys/build.rs
```diff
@@ -1,114 +1,487 @@
...
+            lib_name: &str,
+            include_path: &Path,
+            with_header_files: bool,
+        ) -> Option<PathBuf> {
+            if !with_header_files {
+                return None;
+            }
+            let file_name = PathBuf::from(lib_name.replace('-', "_"));
+            let header_file = include_path.join(file_name.with_extension("h"));
+            if with_header_files && !header_file.is_file() {
```
ionut-arm, line 478: Is the `with_header_files` needed here, since you already check for it at the beginning?

ionut-arm (commented):
That's quite a complicated build file, I'm surprised something like it doesn't exist more generally..

Thanks for the patch!

tss-esapi-sys/build.rs
Superhepper: When you compile under windows ( I used msvc 2022) you get a lot of errors of structs that it some how thinks are needed but not available probably due to it importing "windows.h" or something. So I just blocked everything that caused an error.

tss-esapi-sys/build.rs
Superhepper: Ah yes! Nice catch it should prompt an error saying that the installation path is expected to be a directory.

tss-esapi-sys/build.rs
Superhepper: Absolutely

tss-esapi-sys/build.rs
Superhepper: No it is not necessary. I will fix that.

tss-esapi-sys/build.rs
Superhepper: I meant when you generate the bindings on a windows machine.

wiktor-k: > When you compile under windows ( I used msvc 2022) you get a lot of errors of structs that it some how thinks are needed but not available probably due to it importing "windows.h" or something.

I wonder if [`WIN32_LEAN_AND_MEAN`](https://devblogs.microsoft.com/oldnewthing/20091130-00/?p=15863) would help here... (note that I'm no C/C++ dev so this may be wildly wrong)...

tss-esapi-sys/build.rs
Superhepper: Fixed

tss-esapi-sys/build.rs
```diff
@@ -1,114 +1,515 @@
...
+                    install_path.to_string_lossy(),
+                    e
+                );
+            });
+
+            let min_version_req_str = format!(">={}", MINIMUM_VERSION);
+            let min_version_req = VersionReq::parse(&min_version_req_str).unwrap_or_else(|e| {
+                panic!(
+                    "[Internal Error]: Failed to parse minimum tpm2-tss library version \
+                     requirement. Error: `{}`. Please report this.",
```
ionut-arm, line 239: Love the "Please report this" bit, we should do it more.

tss-esapi-sys/build.rs
ionut-arm: Thanks!

tss-esapi-sys/build.rs
ionut-arm: Thank you!

tss-esapi-sys/build.rs
ionut-arm: Looks great, thank you

ionut-arm (approved):
👍🏻

wiktor-k:
It looks good on a quick look. Is there a plan to add a windows target that at least builds it and checks that it doesn't regress?

I've seen people commenting on how to build it and such a CI config would be one place to point them... :thinking:

Superhepper:
I definitely have it on my list of things I want to improve. But I have found no way of accomplishing it. So I am going to create a ticket for it and if some Windows savvy github actions enthusiast could take a look at it. It would be great.

wiktor-k (approved):

Issue #468, opened 2023-11-03
Remove clippy checks from test scripts


ionut-arm:
Our testing scripts currently include clippy checks. This means that when clippy finds something to complain about, all CI jobs trying to run tests also fail. Ideally only one job would fail (perhaps a common "format and static checks" job), and the tests would continue unhindered so we know if there is something functionally incorrect too.

PR #438, opened 2023-09-29
Bump bindgen and update bindings

gowthamsk-arm:
Signed-off-by: Gowtham Suresh Kumar <gowtham.sureshkumar@arm.com>

PR #478, opened 2023-12-16
Moves Clippy lint checks from tests scripts to a separate job.

Superhepper:
This fixes https://github.com/parallaxsecond/rust-tss-esapi/issues/468.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,7 +63,7 @@ jobs:
       - name: Build the container
         run: docker build -t fedoracontainer tss-esapi/tests/ --file tss-esapi/tests/Dockerfile-fedora
       - name: Run the tests
-        run: docker run -v $(pwd):/tmp/rust-tss-esapi -w /tmp/rust-tss-esapi/tss-esapi fedoracontainer dbus-run-session -- /tmp/rust-tss-esapi/tss-esapi/tests/all-fedora.sh
+        run: docker run -v $(pwd):/tmp/rust-tss-esapi -w /tmp/rust-tss-esapi/tss-esapi --env USE_FROZEN_LOCKFILE=1 fedoracontainer dbus-run-session -- /tmp/rust-tss-esapi/tss-esapi/tests/all-fedora.sh
 
   tests-fedora-rawhide:
     name: Fedora rawhide tests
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -96,3 +96,12 @@ jobs:
         run: docker build -t ubuntucontainer tss-esapi/tests/ --file tss-esapi/tests/Dockerfile-ubuntu
       - name: Check documentation
         run: docker run -v $(pwd):/tmp/rust-tss-esapi -w /tmp/rust-tss-esapi/tss-esapi -e RUSTDOCFLAGS="-Dwarnings" ubuntucontainer cargo doc --document-private-items --no-deps
+
+  # Check that there are no Clippy lint errors.
+  clippy:
+    name: Check Clippy lints
```
wiktor-k, line 102: I'm not sure if this solves the problem without `continue on error`: https://stackoverflow.com/a/73357322

Superhepper: No I do not think it either because I will probably need to run in a container. Because it is run with all features which means that it wants to generate bindings. I am not sure that we want that?

Superhepper:
@wiktor-k @ionut-arm 

I think we need get a change in for the Fedora version we run tests on in order for the CI tests to pass.

ionut-arm (approved):
Thanks for the fixes!!

ionut-arm:
> I think we need get a change in for the Fedora version we run tests on in order for the CI tests to pass.

That seems fair to me, Fedora 35 is more than 2 years old and this is for v8.

.github/workflows/ci.yml
wiktor-k: I mean that without `continue-on-error: true` failing clippy will abort all other jobs and as I understood in #438 @ionut-arm wanted to continue even on clippy failure. Thus "This fixes https://github.com/parallaxsecond/rust-tss-esapi/issues/468." is not correct IMHO.

ionut-arm: Oh, I see.. I thought it would continue running the others regardless. That would be ideal, but even if we just avoid everything failing because of clippy would also be good.

Superhepper: Are you sure it will abort all other jobs? I thought it would only abort the remaining steps in the same job.

wiktor-k: Hmm... I *was* sure when I wrote my comment as I had a similar issue elsewhere but now I'm double thinking it... I'll need to run a test or two to verify...

wiktor-k: Yep, you're right. I just verified that and it runs other jobs to completion even if the fastest one fails: https://github.com/wiktor-k/ci-test/actions/runs/7567980145

Sorry for the noise, resolving this discussion!

wiktor-k (approved):

PR #482, opened 2023-12-31
Fixes max size issue with SensitiveData.

### LINKED ISSUE #481: The max size of the 'SensitiveData' buffer is to big.
When calculating what size to use as max size for the SensitiveData buffer the following expression is used ```::std::mem::size_of::<TPM2B_SENSITIVE_DATA>()```. This is incorrect this will include the size value (an u16 value) when calculating the maximum size of the buffer. Which in turn means it would be possible to create a structure that contains more data then is allowed.

The following test shows the issue:
```rust
#[test]
fn test_conversion_from_max_size_buffer() {
    let data = vec![1u8; SensitiveData::MAX_SIZE];
    let sensitive_data = SensitiveData::try_from(data)
        .expect("It should be possible to convert maximum amount of data into SensitiveData.");
    TPM2B_SENSITIVE_DATA::try_from(sensitive_data)
        .expect("It should be possible to valid convert SensitiveData into TPM2B_SENSITIVE_DATA.");
}
```

That test will create the following error:
```
running 1 test
thread 'structures_tests::buffers_tests::sensitive_create_buffer_tests::test_conversion_from_max_size_buffer' panicked at tss-esapi/src/structures/buffers.rs:357:5:
range end index 258 out of range for slice of length 256
```

Superhepper:
This fixes #481

### DIFF
--- a/tss-esapi/build.rs
+++ b/tss-esapi/build.rs
@@ -38,6 +38,11 @@ fn main() {
         println!("cargo:rustc-cfg=has_tss_base_rc_values_52_to_53")
     }
 
+    let has_tpmu_sensitive_create_req = VersionReq::parse(">=4.0.0").unwrap();
+    if has_tpmu_sensitive_create_req.matches(&tss_version) {
+        println!("cargo:rustc-cfg=has_tpmu_sensitive_create")
+    }
+
     #[cfg(feature = "generate-bindings")]
     {
         let has_esys_tr_get_tpm_handle_req = VersionReq::parse(">=2.4.0").unwrap();
--- a/tss-esapi/src/structures/buffers.rs
+++ b/tss-esapi/src/structures/buffers.rs
@@ -354,11 +354,28 @@ pub mod public_key_rsa {
 }
 
 pub mod sensitive_data {
-    buffer_type!(
-        SensitiveData,
-        ::std::mem::size_of::<TPM2B_SENSITIVE_DATA>(),
-        TPM2B_SENSITIVE_DATA
-    );
+    // The specification says that the size of the buffer should be the size
+    // TPMU_SENSITIVE_CREATE structure. This does not exist in all the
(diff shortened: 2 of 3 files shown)


wiktor-k (dismissed):
So... if I understand the change, it's using sensitive data create on tss versions >= 4 and a workaround with manually calculating the size for lower versions?

Superhepper:
Yes you understand it correctly.

Calculating the size for versions >= 4.0.0 is just using the way that the specification says it should be done.

> *11.1.14 TPM2B_SENSITIVE_DATA*
>
> This buffer wraps the TPMU_SENSITIVE_CREATE structure.
>
> | Parameter  | Type        | Description |
> | ------------- | ----------- | ------------- |
> | size            |   UINT16  |                  |
> | buffer[size]{: sizeof(TPMU_SENSITIVE_CREATE)} | BYTE | symmetric data for a created object or the label and context for a derived object |

But in earlier versions that structure is not available so it will be manually calculated by removing the size of the size variable.

Actually using the buffer macro for TPM2B types that are not sized buffers is not _great_. Because for other types the size is fixed and any other size then the specified one should be an error while for sized buffers the size just needs to be lower or equal to the max size.

wiktor-k:
Thanks for the explainer. I think it's worth having your knowledge captured in some persistent media (like a comment here). 🙇‍♂️

wiktor-k (approved):

Superhepper:
> Thanks for the explainer. I think it's worth having your knowledge captured in some persistent media (like a comment here). 🙇‍♂️

Fixed it. Added a comment the summarize what I wrote above.

wiktor-k (approved):

Superhepper:
@ionut-arm Could you please have a look at this as well.

ionut-arm (approved):
Very nice, thanks!

PR #483, opened 2023-12-31
Fixes lint errors reported in rust 1.75.

Superhepper:


wiktor-k (approved):
Looks cleaner 👍 👌 

(Although the hex changes from `0xff` to `0xFF` look unnecessary at first).

Superhepper:
> Looks cleaner 👍 👌
> 
> (Although the hex changes from `0xff` to `0xFF` look unnecessary at first).

They are actually unnecessary so you are right there I was trying to fix various spelling errors while I was at it and the extension I used complained about the lower letter hex values.

ionut-arm (approved):
Thanks!

PR #486, opened 2024-01-18
ea_commands: add policy_authorize_nv

gomesj:
Although binding is already present, policy_authorize_nv was missing from the supported enhanced authorization commands.
Add a simple implementation to fill the gap.

### DIFF
--- a/tss-esapi/tests/integration_tests/common/mod.rs
+++ b/tss-esapi/tests/integration_tests/common/mod.rs
@@ -10,22 +10,24 @@ use std::{
 use tss_esapi::{
     abstraction::{cipher::Cipher, pcr::PcrData},
     attributes::ObjectAttributes,
-    attributes::{ObjectAttributesBuilder, SessionAttributesBuilder},
+    attributes::{NvIndexAttributesBuilder, ObjectAttributesBuilder, SessionAttributesBuilder},
     constants::SessionType,
+    handles::{NvIndexHandle, NvIndexTpmHandle},
     interface_types::{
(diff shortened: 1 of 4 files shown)


Superhepper:
Thank you for the PR. Could you please also add some tests for this Context method?

gomesj:
Good point, added a simple test case in my last push.

tss-esapi/tests/integration_tests/common/mod.rs
```diff
@@ -432,3 +434,45 @@ pub fn create_public_sealed_object() -> Public {
         .build()
         .expect("Failed to create public structure.")
 }
+
+#[allow(dead_code)]
```
wiktor-k, line 438: Is this really necessary?

wiktor-k (approved):
Looks good to me although I think @Superhepper meant a doctest in the Context function you added. (Note that they are slightly harder to write but you can take a look at other functions for inspiration).

Superhepper:
Well actually I meant both. The methods need to be documented a little better especially when it is a Context method. Unfortunately not all Context methods are documented properly at the moment. But there are a few good examples that you can have a look at (e.g. ```ecdh_key_gen```).

tss-esapi/tests/integration_tests/common/mod.rs
gomesj: I originally followed the example of most other functions in this file, but apparently it was not.

Superhepper (approved):

PR #490, opened 2024-02-02
Migrate from mbox to malloced

### LINKED ISSUE #371: Replacing `MBox`
[`Mbox`](https://github.com/kennytm/mbox) seems to have gone out of maintenance - specifically, there are lots of PRs open, and (serious-looking) issues that have no reply on them. We should decide whether we want to continue with `Mbox`, switch to a different crate doing the same thing, or if we want to implement our own functionality for the stuff that we need.

We're currently sticking with `Mbox` by default, but that might not be a great idea in the long run - I also expect it to be reported as unmaintained in the Cargo audit database at some point, it won't look good for users of `tss-esapi` if we keep using it.

An alternative crate that seems to be maintained and does similar things is [`malloced`](https://crates.io/crates/malloced). However, it seems to be less popular, at least for now.

The last option is for us to copy the functionality that we care about into our crate and use that. That might be the "safest" option, but it could also come with some overhead, especially if it turns out to be quite a big chunk of code. But I'd lean towards this - I think we've been somewhat limited in our use of `MBox`.

kcking:
Fixes #371.

### DIFF
--- a/tss-esapi/Cargo.toml
+++ b/tss-esapi/Cargo.toml
@@ -18,7 +18,7 @@ name = "hmac"
 [dependencies]
 bitfield = "0.13.2"
 serde = { version = "1.0.115", features = ["derive"] }
-mbox = "0.6.0"
+malloced = "1.3.1"
 log = "0.4.11"
 enumflags2 = "0.7.7"
 num-derive = "0.3.2"
--- a/tss-esapi/src/context.rs
+++ b/tss-esapi/src/context.rs
@@ -13,7 +13,7 @@ use crate::{
 };
 use handle_manager::HandleManager;
 use log::{debug, error};
-use mbox::MBox;
+use malloced::Malloced;
 use std::collections::HashMap;
 use std::ptr::null_mut;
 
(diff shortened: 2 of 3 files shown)


Superhepper:
Nice! This is something I have been looking at my self a couple of times.

Seems like the malloced crate have not been updated in 4 years.
mbox have not been updated in 3 years.

I wonder what other options are out there. I would like to see a library that gets maintained and could provide the functionality that we have implemented here. Which is basically from ffi to owned with a possiblity of adding a cleanup function.

kcking:
The last commit and release of `malloced` was just under 1 year ago. As there are no outstanding issues, there is no evidence the crate is unmaintained. Unfortunately `mbox` has a couple of potential soundness issues and the author is unresponsive. Perhaps `malloced` could be considered as a short-term alternative, leaving open the possibility later for a crate that supports cleanup natively.

Superhepper:
Yes, I would rather like a dependency that has no known issues and was updated `recently` then have a dependency to something that seams abandoned and with a maintainer that does not seem unresponsive. Maybe it would even be possible to add feature request to the malloced crate.

ionut-arm:
I'm definitely for this move, seems like a great alternative for `MBox`

Superhepper:
Could you please rebase this on the latest changes from main?

kcking:
> Could you please rebase this on the latest changes from main?

Done :)

ionut-arm (approved):

Superhepper (approved):
Thank you!

PR #492, opened 2024-02-04
Add AES encrypt/decrypt example

Firstyear:
This adds an example of how to use AES-128-CBC encryption and decryption with a TPM.

### DIFF
--- a/tss-esapi/examples/aes_encrypt_decrypt.rs
+++ b/tss-esapi/examples/aes_encrypt_decrypt.rs
@@ -1,0 +1,235 @@
+// Copyright 2024 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+
+/*
+ * This example demonstrates how to use AES for symmetric encryption and decryption
+ */
+
+use tss_esapi::{
+    Context, TctiNameConf,
+    attributes::ObjectAttributesBuilder,
+    interface_types::{
+        algorithm::{HashingAlgorithm, PublicAlgorithm, SymmetricMode},
+        key_bits::AesKeyBits,
+        reserved_handles::Hierarchy,
+    },
+    structures::{
+        CreatePrimaryKeyResult, Digest, InitialValue, MaxBuffer, PublicBuilder,
+        SymmetricCipherParameters, SymmetricDefinitionObject,
+    },
+};
+
+use std::convert::TryFrom;
+
+fn main() {
+    // Create a new TPM context. This reads from the environment variable `TPM2TOOLS_TCTI` or `TCTI`
+    //
+    // It's recommended you use `TCTI=device:/dev/tpmrm0` for the linux kernel
+    // tpm resource manager.
+    let mut context = Context::new(
+        TctiNameConf::from_environment_variable()
+            .expect("Failed to get TCTI / TPM2TOOLS_TCTI from environment. Try `export TCTI=device:/dev/tpmrm0`"),
+    )
+    .expect("Failed to create Context");
+
+    // This example won't go over the process to create a new parent. For more detail see `examples/hmac.rs`.
+
+    let primary = create_primary(&mut context);
+
+    // Begin to create our new AES symmetric key
+
+    let object_attributes = ObjectAttributesBuilder::new()
+        .with_fixed_tpm(true)
+        .with_fixed_parent(true)
+        .with_st_clear(false)
+        .with_sensitive_data_origin(true)
+        .with_user_with_auth(true)
+        .with_sign_encrypt(true)
+        .with_decrypt(true)
+        // Note that we don't set the key as restricted.
+        .build()
+        .expect("Failed to build object attributes");
+
+    let aes_params = SymmetricCipherParameters::new(SymmetricDefinitionObject::Aes {
+        key_bits: AesKeyBits::Aes128,
+        mode: SymmetricMode::Cbc,
+    });
+
+    let key_pub = PublicBuilder::new()
+        .with_public_algorithm(PublicAlgorithm::SymCipher)
+        .with_name_hashing_algorithm(HashingAlgorithm::Sha256)
+        .with_object_attributes(object_attributes)
+        .with_symmetric_cipher_parameters(aes_params)
+        .with_symmetric_cipher_unique_identifier(Digest::default())
+        .build()
+        .unwrap();
+
+    let (enc_private, public) = context
+        .execute_with_nullauth_session(|ctx| {
+            ctx.create(primary.key_handle, key_pub, None, None, None, None)
+                .map(|key| (key.out_private, key.out_public))
+        })
+        .unwrap();
+
+    // The data we wish to encrypt. Be aware that there is a limit to the size of this data
+    // that can be encrypted or decrypted (1024 bytes). In some cases you may need to encrypt a
+    // content encryption key (CEK), which can be decrypted and released and then used to decrypt
+    // the actual data in question outside of the TPM.
+    //
+    // TPMs also tend to be "slower" for encryption/decryption, so you should consider the
+    // CEK pattern for performance reasons.
+    let data_to_encrypt = "TPMs are super cool, you should use them!"
+        .as_bytes()
+        .to_vec();
+
+    eprintln!("{:?}", data_to_encrypt.len());
+
+    // Input data needs to always be a multiple of the AES block size, in this case which is 16
+    // bytes for AES-128-CBC. Normally you *MUST* implement a secure padding scheme such as pkcs7
+    // but in this example we will *manually* pad the data.
+
+    // WARNING: Manually implemented pkcs7 follows. This has not been audited. Don't use this
+    // in production.
+    const AES_BLOCK_SIZE: usize = 16;
+
+    let need_k_bytes = AES_BLOCK_SIZE - (data_to_encrypt.len() % AES_BLOCK_SIZE);
+    // PKCS7 always pads to remove ambiguous situations.
+    let need_k_bytes = if need_k_bytes == 0 {
+        AES_BLOCK_SIZE
+    } else {
+        need_k_bytes
+    };
+
+    let new_len = data_to_encrypt.len() + need_k_bytes;
+
+    let mut padded_data_to_encrypt = data_to_encrypt.to_vec();
+    padded_data_to_encrypt.resize(new_len, need_k_bytes as u8);
+
+    let padded_data_to_encrypt = MaxBuffer::try_from(padded_data_to_encrypt).unwrap();
+
+    // Padding always has to be added in pkcs7 to make it unambiguous.
+    assert_ne!(
+        data_to_encrypt.as_slice(),
+        padded_data_to_encrypt.as_slice()
+    );
+    // END WARNING
+
+    // AES requires a random initial_value before any encryption or decryption. This must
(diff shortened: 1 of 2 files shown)


Firstyear:
Please note there are two comments in the code that we should probably discuss. First is exposing the AES block size from the AesKeyBits enum. The second is if we should add pkcs7 padding to the crate to prevent people from "rolling their own".

ionut-arm (commented):
Thank you for the effort of putting these examples together!

tss-esapi/examples/aes_encrypt_decrypt.rs
```diff
@@ -0,0 +1,231 @@
...
+        .to_vec();
+
+    eprintln!("{:?}", data_to_encrypt.len());
+
+    // Input data needs to always be a multiple of AES_BLOCK_SIZE, so we implement PKCS7 padding
+    // to achieve this.
+
+    // REVIEW NOTE: Tss-esapi likely should expose these as constants from AesKeyBits::Aes128
+    // to prevent ambiguity!
+    const AES_BLOCK_SIZE: usize = 16;
```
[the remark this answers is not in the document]
Firstyear: Wait what? I thought that block size was tied to the AESKeyBit size? IE AES-128 requires a 128bit key and a 16 byte block, where AES-256 requires a 256 bit key and a 32 byte block? I'm happy to be wrong, but all my other experience with openssl etc has the above assumptions enshrined.

tss-esapi/examples/aes_encrypt_decrypt.rs
```diff
@@ -0,0 +1,231 @@
...
+
+    // REVIEW NOTE: Should we added PKCS7 padding as a function to MaxBuffer to prevent
+    // people needing to "roll their own"?
```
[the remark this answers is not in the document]
Firstyear: Yeah, I think that's a good idea. But I suspect that is it's own PR? 

Still even so, this example has highlighted a few places the api we have can be better :)

tss-esapi/examples/aes_encrypt_decrypt.rs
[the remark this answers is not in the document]
ionut-arm: See [NIST FIPS 197](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.197.pdf), first paragraph of section 1, first paragraph of section 3.1 ... Rijndael accepts different block sizes, but AES does not.

I'm not that great at digging through OpenSSL docs, the best I found was [this](https://www.openssl.org/docs/manmaster/man3/EVP_EncryptInit.html#blocksize-OSSL_CIPHER_PARAM_BLOCK_SIZE-unsigned-integer).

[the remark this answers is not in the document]
Firstyear: Ahhh yes, it's related to cipher selection not key. Maybe it should be on https://docs.rs/tss-esapi/latest/tss_esapi/interface_types/algorithm/enum.SymmetricMode.html or https://docs.rs/tss-esapi/latest/tss_esapi/interface_types/algorithm/enum.SymmetricAlgorithm.html 🤔

AES-GCM makes it tricky since it's block size is 1, but I think the way we have the parameters laid out we need the SymmetricMode + Key to really know what the block size is.

tss-esapi/examples/aes_encrypt_decrypt.rs
```diff
@@ -0,0 +1,231 @@
...
+    // any values.
+    let (encrypted_data, _initial_value) = context
+        .execute_with_nullauth_session(|ctx| {
+            let aes_key = ctx
+                .load(primary.key_handle, enc_private.clone(), public.clone())
+                .unwrap();
+
+            let decrypt = false;
+
+            ctx.encrypt_decrypt_2(
```
ionut-arm, line 139: You should add a comment explaining why you chose `encrypt_decrypt_2` instead of `encrypt_decrypt_2`. Also, we should do a better job of explaining that difference in our [docs](https://docs.rs/tss-esapi/7.4.0/tss_esapi/struct.Context.html#method.encrypt_decrypt_2).

tss-esapi/examples/aes_encrypt_decrypt.rs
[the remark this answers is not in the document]
ionut-arm: I'm not actually sure if in the context of TPMs the same applies. I've skimmed a bit through the spec, couldn't find much, I'll need to test it.

tss-esapi/examples/aes_encrypt_decrypt.rs
Firstyear: Huh? I don't know what you mean? encrypt_decrypt_2 == encrypt_decrypt_2

ionut-arm: Ooops, my bad! I meant `encrypt_decrypt_2` instead of `encrypt_decrypt`.

wiktor-k: Just FYI the older one has been deprecated: https://github.com/tpm2-software/tpm2-tools/issues/253

Superhepper: Yes, it's deprecated in the specification as well and only remains due to backward compatibility I think.

Firstyear: Given that I can't see encrypt_decrypt at all in Context, do I really need to document why I chose this?

wiktor-k: I think this is just an oversight of @ionut-arm so... I'd consider this thread resolved :)

tss-esapi/examples/aes_encrypt_decrypt.rs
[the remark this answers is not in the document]
Firstyear: I'm not even sure we'd need that since https://docs.rs/block-padding/0.3.2/block_padding/struct.Pkcs7.html already has .as_slice and we have https://docs.rs/tss-esapi/latest/tss_esapi/structures/struct.MaxBuffer.html#impl-TryFrom%3C%26%5Bu8%5D%3E-for-MaxBuffer 

So we could just swap the padding process here.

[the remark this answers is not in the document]
Firstyear: Actually, reading that API it seems to be really limited and complex.

We would need to take the array of bytes, then iterate over it as chunks exact with https://doc.rust-lang.org/std/slice/struct.ChunksExact.html to match the block size. 

We need this because we need to check the remainder case. If the remainder is empty, we need to append another block because pkcs7 always pads to avoid ambiguity.  So that means we can't just look at the vec and grow it if needed, we have to clone the full input everytime because blocks are effectively an array. 

And then a vec of generic array can't do to_slice, so we have to iterate *again* to make a new vec that we can try_from for max_buffer.

So unless I'm really overlooking something obvious, this seems like a really rough api to want to interface with.

Firstyear:
> Have you forgot to sign the commits?

Very likely :) But it wasn't ready to merge yet either.

Firstyear:
Okay, I've rebased an updated to latest main. I think that pkcs7/padding is "out of scope" for this crate, and I'm going to make an example of exposing the TPM via rust-crypto interfaces so we can inherit their padding options. I'll hopefully have this example "converted" shortly as a POC.

Firstyear:
This has finally come up again so I should be looping back to improve this shortly :)

tss-esapi/examples/aes_encrypt_decrypt_rust_crypto.rs
```diff
@@ -0,0 +1,312 @@
...
+    type BlockSize = cipher::consts::U16;
+}
+
+impl<'a> cipher::ParBlocksSizeUser for TpmEnc<'a> {
+    // 1024 / 16 bytes
+    type ParBlocksSize = cipher::consts::U16;
+}
+
+impl<'a> cipher::BlockBackend for TpmEnc<'a> {
+    fn proc_block(&mut self, mut block: cipher::inout::InOut<'_, '_, cipher::Block<Self>>) {
```
wiktor-k, line 286: I guess `InOut` needs to be added [here](https://github.com/parallaxsecond/rust-tss-esapi/blob/main/.codespellrc#L3) :roll_eyes:

mcr (approved):

tss-esapi/examples/aes_encrypt_decrypt.rs
```diff
@@ -0,0 +1,235 @@
...
+    assert_ne!(encrypted_data.as_slice(), padded_data_to_encrypt.as_slice());
+
+    // Decryption is the identical process with the "decrypt" flag set to true.
+    let (decrypted_data, _initial_value) = context
+        .execute_with_nullauth_session(|ctx| {
+            let aes_key = ctx
+                .load(primary.key_handle, enc_private.clone(), public.clone())
+                .unwrap();
+
+            let decrypt = true;
```
wiktor-k: I'm very happy that you've made this explicit and I'm wondering how whether we should be providing `encrypt_2` and `decrypt_2` as two separate functions just for the sake of users :sweat_smile: 

That's of course a totally different subject, nothing to do here, carry on! :rocket:

tss-esapi/examples/aes_encrypt_decrypt_rust_crypto.rs
```diff
@@ -0,0 +1,309 @@
...
+            };
+
+            let enc_data = aes_128_cbc_enc
+                .encrypt_padded_vec_mut::<cipher::block_padding::Pkcs7>(&data_to_encrypt);
+
+            Ok::<_, tss_esapi::Error>(enc_data)
+        })
+        .unwrap();
+
+    /*
```
wiktor-k: I think this entire section can be removed :thinking:

wiktor-k (dismissed):
This is a well-done example :ok_hand: 

I'm wondering though whether the RustCrypto trait implementation should be moved to the main crate. It seems quite useful and generic so instead of folks copying that part of the code they could use our rustcrypto feature and get these traits...

Pinging our RustCrypto guru @baloo :bear:

mcr:
> I'm wondering though whether the RustCrypto trait implementation should be moved to the main crate. It seems quite useful and generic so instead of folks copying that part of the code they could use our rustcrypto feature and get these traits...

Sure, but make a new PR.

mcr (changes requested):
Just the changes baloo asked for.

ionut-arm:
Could we merge this as-is and fix the RustCrypto example / add the RustCrypto functionality in separate PR(s)?

Firstyear:
Blocked by https://github.com/parallaxsecond/rust-tss-esapi/issues/646 at the moment, else I'm happy to split the PR up :)

Firstyear:
@ionut-arm Got it working under a different OS, so I've split this to two PR's.

ionut-arm (dismissed):
I'm ok with this going in as-is - I think the failed CI would require rebasing, but we can just push it through.

Firstyear:
The formatting was the only failure and was an easy fix.

ionut-arm (approved):

wiktor-k (approved):
Thanks for taking the time to make this happen! 🚀

Firstyear:
No problem at all, I'll update and rebase the other example too and we can discuss the traits there.

PR #489, opened 2024-02-01
Backport lint error fixes for rust 1.75

tgonzalezorlandoarm:
Signed-off-by: Tomás González <tomasagustin.gonzalezorlando@arm.com>

PR #493, opened 2024-02-05
Fixes CI problems with fedora rawhide.

Superhepper:


Superhepper:
This needs to be combined with #489 in a single commit in order to work.

wiktor-k:
> This needs to be combined with #489 in a single commit in order to work.

Hmm... is there a reason why this commit was not added to #489 directly? I see it has "Maintainers are allowed to edit this pull request." turned on.

tgonzalezorlandoarm:
I've just brought that commit over to the mentioned PR, everything is passing now - Thanks!

wiktor-k:
> I've just brought that commit over to the mentioned PR, everything is passing now - Thanks!

Do I get it right that we can close this PR now? (if so please press the button :bow: )

Superhepper:
I made one to test it in isolation to try and see if I missed something else then i forgot to add it to the other PR.

PR #502, opened 2024-02-17
Creates proper native type for TPMS_CONTEXT.

Superhepper:
- Moves TPMS_CONTEXT from utils into structures and creates a proper native type for it.
- Adds new types needed by the above mentioned type.
- Changes interface_types module names to the correct ones.


Superhepper:
I would like your thoughts on this draft. Like the structure, naming and documentation and such things.

tss-esapi/src/interface_types/data_handles.rs
```diff
@@ -0,0 +1,191 @@
...
+    Endorsement,
+}
+
+/// Enum representing the 'Persistent' data handles interface type.
+///
+/// # Details
+/// This corresponds to the TPMI_DH_PERSISTENT interface type.
+#[derive(Debug, Copy, Clone)]
+pub enum Persistent {
+    Persistent(PersistentTpmHandle),
```
wiktor-k, line 43: To be honest this one-variant enums look a bit weird... are they here because they need to wrap some TPM structure? Maybe a [newtype](https://doc.rust-lang.org/rust-by-example/generics/new_types.html) pattern would be more appropriate here :thinking:

tss-esapi/src/interface_types/data_handles.rs
```diff
@@ -0,0 +1,191 @@
...
+    // TODO: Handle Auth, that is vendor specific.
+}
+
+/// Enum representing the 'PCR' data handles interface type.
+///
+/// # Details
+/// This corresponds to the TPMI_DH_PCR interface type.
+#[derive(Debug, Copy, Clone)]
+pub enum Pcr {
+    Pcr(PcrTpmHandle),
```
wiktor-k, line 83: I've got similar concerns here...

wiktor-k (commented):
I think it looks mostly straightforward :ok_hand: 

I wonder if some enums don't need to be marked with [`#[non_exhausive]`](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute) if they still will be extended (like the vendor auth).

Also, btw, I wish we could [drop the copyright years](https://hynek.me/til/copyright-years/), just like [curl did](https://github.com/curl/curl/pull/10205) (but it's an orthogonal matter).

ionut-arm:
> Also, btw, I wish we could [drop the copyright years](https://hynek.me/til/copyright-years/), just like https://github.com/curl/curl/pull/10205 (but it's an orthogonal matter).

@gowthamsk-arm @tgonzalezorlandoarm - do you have any info or points of reference for this?

tgonzalezorlandoarm:
> > Also, btw, I wish we could [drop the copyright years](https://hynek.me/til/copyright-years/), just like [curl/curl#10205](https://github.com/curl/curl/pull/10205) (but it's an orthogonal matter).
> 
> @gowthamsk-arm @tgonzalezorlandoarm - do you have any info or points of reference for this?

I only know that TF-M only updates for major changes:
https://trustedfirmware-m.readthedocs.io/en/latest/contributing/contributing_process.html

wiktor-k:
> I only know that TF-M only updates for major changes: https://trustedfirmware-m.readthedocs.io/en/latest/contributing/contributing_process.html

It seems the guidelines are more generic or suited towards Gerrit powered workflows (as evidenced by "Add a [Change-Id](https://review.trustedfirmware.org/Documentation/user-changeid.html) to the commit message").

It'd be cool to clarify with them the copyright years.

The links I sent have a lot of other projects, eg this Linux Foundation guidance:

> Copyright The XYZ Authors.
> Copyright The XYZ Contributors.
> ...
> By using a common format, the project avoids having to maintain lists of names of the authors or copyright holders, years or ranges of years, and variations on the (c) symbol. This aims to minimize the burden on developers and maintainers as well as redistributors of the code...

tss-esapi/src/interface_types/data_handles.rs
Superhepper: I do not think there us any special reason for this. I think I just did it that way for consistency. Every other interface type is an enum so I just did it the same way with those ones.

Superhepper:
I will abandon this PR for one without module name changes and the tpm2-tss library updates, see  #507.

PR #504, opened 2024-02-24
Fixes interface types module names

Superhepper:
Changes the interface types module names for:
`resource_handles` -->`reserved_handles`
`dynamic_handles`-->`data_handles`


ionut-arm (approved):
Thanks for the change! It would be good for them to have an official definition somewhere, but yeah, these new options are clearer for users..

PR #508, opened 2024-03-13
Get random authvalues locally instead of from the TPM

tgonzalezorlandoarm:
Currently, the TPM itself is the root of trust for randomness in
    authvalues used for creating Primary Keys.
    This is susceptible to physical attacks over the TPM bus.
    
     * Get authvalues via the 'getrandom' crate
    
    getrandom retrieves random data from the (operating) system sources
    and assumes "that the system always provides high-quality
    cryptographically secure random data, ideally backed by hardware
    entropy sources", so the administrator of the platfrom should take
    this into account.
    
    Note: This change may slow down the tests as accessing random values
    from the OS instead of the TPM may be slower.


tgonzalezorlandoarm:
I will create the same PR in main once this is reviewed

gowthamsk-arm (approved):
LGTM.

Superhepper:
@tgonzalezorlandoarm 
I hope you will make a pr for main as well?

tgonzalezorlandoarm:
@Superhepper yes, done :+1:

PR #509, opened 2024-03-13
tss-esapi/transient: get_root_key_name

tgonzalezorlandoarm:
Expose the root key name of a TransientKeyContext.
This may be used for verifying previously generated keys.

PR #511, opened 2024-03-13
tss-esapi/transient: get_root_key_name

tgonzalezorlandoarm:
Expose the root key name of a TransientKeyContext. This may be used for verifying previously generated keys.


Superhepper (approved):

tgonzalezorlandoarm:
#509

PR #515, opened 2024-03-30
Disables the clippy multiple_crate_versions lint error.

Superhepper:
In Rust version 1.77 clippy has begun reporting
errors for when two versions of the same crate
exists among the dependencies. In this case
it is the `syn` crate for which there exists two
versions. The reason for this is that the
`zeroize_derive` crate updated the `syn` version
after making a Rust Edition bump. So the ones
using Rust Edition 2018 gets an older version
of `zeroize_derive` which uses `syn` with major
release version 1.

### DIFF
--- a/tss-esapi/Cargo.toml
+++ b/tss-esapi/Cargo.toml
@@ -2,7 +2,7 @@
 name = "tss-esapi"
 version = "8.0.0-alpha"
 authors = ["Parsec Project Contributors"]
-edition = "2018"
+edition = "2018"                                                       # TODO Update edition to 2021 - Remove the disabled lints that are due to using old edition.
 description = "Rust-native wrapper around TSS 2.0 Enhanced System API"
 readme = "README.md"
 keywords = ["tpm", "tss", "esys", "esapi"]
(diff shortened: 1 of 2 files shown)


wiktor-k (approved):
This looks okay but could you also add a comment about it here: https://github.com/parallaxsecond/rust-tss-esapi/blob/main/tss-esapi/Cargo.toml#L13 

Since I guess when someone will be changing the msrv they may not look at other files.

Thanks!

ionut-arm (dismissed):
Hmm, yes, maybe we should bump our edition...

Superhepper:
> This looks okay but could you also add a comment about it here: https://github.com/parallaxsecond/rust-tss-esapi/blob/main/tss-esapi/Cargo.toml#L13
> 
> Since I guess when someone will be changing the msrv they may not look at other files.
> 
> Thanks!

I added a comment to the edition section.

Superhepper:
> Hmm, yes, maybe we should bump our edition...

I added a new issue for this and attached it to the 8.0.0 stabilization.

wiktor-k (approved):

ionut-arm (approved):
🙏🏻

PR #523, opened 2024-04-15
implement bundled feature and improve windows support

uglyoldbob:
This pull request adds a feature bundled. It builds the tpm2-tss library from source after fetching with git.
The normal requirements for build the software apply.

The linux build needs all of the normal requirements for building tpm2-tss as described in their repository.

Windows builds require visual studio 2017, with the c2 clang option, visual studio build tools 10.0.17134.0. Openssl should be installed to C:\OpenSSL-v11-Win64

PR #524, opened 2024-05-29
WIP: Add windows CI

wiktor-k:
This is just an experiment to see how far can we go with building the crate on Windows using code from #523.

I'll try to push it as far as possible and when we're done we can merge that back to the target PR.


wiktor-k:
Okay, I guess that's it for today. The build complains about missing `v141_clang_c2` toolset and I can't find an easy way to install it.

I've made the following adjustments, that should be resolved:

  - changed the SDK minor version from `17134` to `17763` because of that one having an ISO [here](https://developer.microsoft.com/en-us/windows/downloads/sdk-archive/) and as such easily installable in GitHub actions. I had to also adjust tpm-tss2 to be less picky about the SDK version.
  - I've made the msbuild search for any version.

Well... that's it for today... see you later~ :wave:

uglyoldbob:
I think the v141_clang_c2 toolset might only exist in visual studio 2017.

Superhepper:
I think with the bundled feature it should be quite possible to build on windows now if the installation of OpenSSL is as expected.

wiktor-k:
Sadly, I don't have time right now to check this so I'll close the PR as stale. If anyone wants to take it then please be my guest.

Thanks for help and see you later! :wave:

PR #526, opened 2024-06-01
Makes the serde dependency optional.

### LINKED ISSUE #525: Make the serde dependency optional.
The serde dependency is actually not really needed for building and using the library and should be changed into a feature.

Superhepper:
This fixes #525 adding the `serde` feature which when enabled makes it possible to serialize
and deserialize the types that previously
supported it automatically.

### DIFF
--- a/tss-esapi/tests/coverage.sh
+++ b/tss-esapi/tests/coverage.sh
@@ -16,4 +16,4 @@ tpm2_startup -c -T mssim
 # Install and run tarpaulin #
 #############################
 cargo install cargo-tarpaulin
-cargo tarpaulin --features integration-tests --tests --out xml --exclude-files="tests/*,../*" -- --test-threads=1 --nocapture
+cargo tarpaulin --features "integration-tests serde" --tests --out xml --exclude-files="tests/*,../*" -- --test-threads=1 --nocapture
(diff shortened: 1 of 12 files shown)


tss-esapi/tests/coverage.sh
```diff
@@ -16,4 +16,4 @@ tpm2_startup -c -T mssim
 # Install and run tarpaulin #
 #############################
 cargo install cargo-tarpaulin
-cargo tarpaulin --features integration-tests --tests --out xml --exclude-files="tests/*,../*" -- --test-threads=1 --nocapture
+cargo tarpaulin --features integration-tests serde --tests --out xml --exclude-files="tests/*,../*" -- --test-threads=1 --nocapture
```
wiktor-k, line 19:
```suggestion
cargo tarpaulin --features integration-tests,serde --tests --out xml --exclude-files="tests/*,../*" -- --test-threads=1 --nocapture
```

Either a comma or enclose with quotes, either way, it's better to test this :sweat_smile:

wiktor-k (dismissed):
LGTM :+1: thanks! (I've been looking at cutting serde usage in my crates too so it's nice to see it here!)

tss-esapi/tests/coverage.sh
Superhepper: fixed

ionut-arm (dismissed):
Awesome stuff, thank you!

wiktor-k (approved):

ionut-arm (approved):

PR #527, opened 2024-06-02
Fixes Clippy lint errors reported in rust 1.78

Superhepper:
In version 1.78.0 of rust the clippy lint tool
report an error for structs that have a const
`new` function but lacks an implementation
of the `Default` trait.


ionut-arm (approved):
🙏🏻 thank you for putting this together!

hug-dev (approved):

PR #528, opened 2024-06-04
Add support for testing with opensuse

Firstyear:
This adds opensuse as an option for the docker integration tests. Additionally, it adds a .dockerignore file to improve build times and reduce image size.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,12 @@
+.git
+.github
+.gitignore
+.mypy_cache
+.vscode
+*.md
+**/__pycache__/**
+docs
+Makefile
+target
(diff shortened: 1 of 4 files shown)


wiktor-k (approved):
I think it looks good. It has some stuff for vendoring that we don't use but I'm not opposed to that.

(For the vendoring to work with GitHub cache [additional config would be needed](https://github.com/wiktor-k/currencies/blob/main/.github/workflows/docker.yml#L32-L34)).

Superhepper (approved):
Lgtm

ionut-arm (approved):
👍🏻

PR #529, opened 2024-06-05
20240604 506 create loaded

### LINKED ISSUE #506: Support Created Loaded
CreateLoaded is required for key derivation. We should add it to this library to allow KDF usage for applications.

Firstyear:
Fixes #506

Start on created loaded.

Create loaded appears to work in the base case, but when using it for key derivation, it appears to be failing with:

```
[2024-06-05T04:39:58Z ERROR tss_esapi::context::tpm_commands::object_commands] Error in creating derived key: 0x000002D2
thread 'context_tests::tpm_commands::object_commands_tests::test_create_loaded::test_create_loaded_tpm_alg_kdf1_sp800_108' panicked at tss-esapi/tests/integration_tests/context_tests/tpm_commands/object_commands_tests.rs:519:14:
Failed to create derivation parent.: TssError(Tpm(FormatOne(TpmFormatOneResponseCode { error_number: Scheme, argument_number: Parameter(2) })))
```

I'm pretty sure I'm making a mistake in the derive_parent_public PublicBuilder parameters that I'm setting. Per the spec, this does need to be keyed hash, but I think given the parameters I've set, I think that we need to set the KDF1SP800_108 as the scheme when we create this key - currently though I think that's not possible as KeyHashSchemes are just for Xor and Hmac, not derivation.

So any advice would be more than welcome on how to proceed next :)


tss-esapi/tests/integration_tests/context_tests/tpm_commands/object_commands_tests.rs
```diff
@@ -427,3 +427,154 @@ mod test_unseal {
...
+        // Create primary.
+        let primary_key_handle = context
+            .create_primary(Hierarchy::Owner, primary_pub, None, None, None, None)
+            .unwrap()
+            .key_handle;
+
+        // Create Derivation Parent
+        // - How to mark an object as a derivation parent? From what I read
+        // in the spec, a derivation parent is just when != (primary || storage)
+        let derive_parent_object_attributes = ObjectAttributesBuilder::new()
```
Superhepper, line 488: It is not that simple. It also depends on what kind of handle the parent is when you call `CreateLoaded`.

Look at `TCG TPM2, r1p59, Part3 Commands, Section 12.9.1`

Firstyear: Yes I did read that section. The problem I hit was that part 3 12.9.1 doesn't actually define a derivation parent - that's actually defined in part 1 4.19 as "loadable key used to derive other keys; a TPM_ALG_KEYEDHASH Parent Key" and part 1 25.1.5 table 24 states "Asymmetric keys and
symmetric keys with these attributes are Storage Parents, and hash objects with these attributes are
Derivation Parents. " where the attributes are sign=clear, decrypt=set, restricted=set. 

I think the comment I wrote here needs to be updated to reflect this, but I still think the attributes I used on the key were correct :)

Firstyear:
Okay, I have finally come back to this and after rebasing and experimenting I can not solve it. 

I still get the same scheme error, but everything in the specification says these parameters should be correct. So I am clearly not understanding something here. 

What's really interesting though is this request isn't even making it to swtpm - it's getting rejected somewhere in tss2. But after a ton of reading I can't find where. I think I'll ask the tpm2-tss project about it and see if they have some ideas.

Firstyear:
https://github.com/tpm2-software/tpm2-tss/issues/3124

Firstyear:
Alright, progress! It looks like you need to use KeyedHashAlgorithm Xor with KDF. I've done that now and have progressed, but now during the actual object derivation itself I hit an error due to insufficient bytes in the marshalled data. I'm going to re-read all the structures again to be sure but any advice would be great.

Firstyear:
So it looks the problem is that the design of the structures and enums never really took into account that Derive exists.

There are two places derive parameters can be used - in the unique field, or sensitive data. 

From the unique side the issue is that the marshalling doesn't correctly handle when a derive is present as it's a pair of length-values, not a single one. This results in the truncation and removal of context causing the RC_SIZE error. 

For sensitive data, the issue is that while sensitive data *should* be an enum to account for being both a buffer *or* a set of derive parameters, the code was designed so that sensitive data is always the buffer.

Finally to throw some more fuel into this, it looks like these isn't actually a way to marshall to TPMU_SENSITIVE_CREATE because it goes via Tss2_MU_TPMU_SENSITIVE_COMPOSITE_Marshal instead which doesn't support derive parameters. So I think without some nasty hand rolled marshalling, this may be a dead end.

Firstyear:
Okay, Implemented the hand-rolled marshalling and still hitting issues. For now, I might call this one dead.

PR #530, opened 2024-06-05
Fix start_auth_session resource leak

Zha0Chan:
The session handle returned by _start_auth_session_. If there is an error in subsequent function calls, it is best to release it immediately.

### DIFF
--- a/tss-esapi/src/abstraction/ak.rs
+++ b/tss-esapi/src/abstraction/ak.rs
@@ -129,7 +129,7 @@ pub fn load_ak(
             session_attributes_mask,
         )
         .or_else(|e| {
-            ctx.flush_context(SessionHandle::from(policy_auth_session).into())?;
+            context.flush_context(SessionHandle::from(policy_auth_session).into())?;
             Err(e)
         })?;
 
@@ -194,7 +194,7 @@ pub fn create_ak<IKC: IntoKeyCustomization>(
             session_attributes_mask,
         )
         .or_else(|e| {
-            ctx.flush_context(SessionHandle::from(policy_auth_session).into())?;
+            context.flush_context(SessionHandle::from(policy_auth_session).into())?;
             Err(e)
         })?;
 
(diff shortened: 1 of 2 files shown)


ionut-arm (dismissed):
Thanks for the patch, it makes sense to me, I guess we haven't noticed because those commands don't fail that often.

Superhepper (dismissed):
Nice catch

tss-esapi/src/abstraction/ak.rs
```diff
@@ -182,11 +187,16 @@ pub fn create_ak<IKC: IntoKeyCustomization>(
...
-        session_attributes_mask,
-    )?;
+    context
+        .tr_sess_set_attributes(
+            policy_auth_session,
+            session_attributes,
+            session_attributes_mask,
+        )
+        .or_else(|e| {
+            ctx.flush_context(SessionHandle::from(policy_auth_session).into())?;
```
Superhepper, line 197: There is no `ctx` in this scope.

tss-esapi/src/context.rs
```diff
@@ -305,7 +305,11 @@ impl Context {
             .with_decrypt(true)
             .with_encrypt(true)
             .build();
-        self.tr_sess_set_attributes(auth_session, session_attributes, session_attributes_mask)?;
+        self.tr_sess_set_attributes(auth_session, session_attributes, session_attributes_mask)
+            .or_else(|e| {
+                ctx.flush_context(SessionHandle::from(auth_session).into())?;
```
Superhepper, line 310: there is no `ctx` in this scope either.

Zha0Chan (commented):
Fixed some typos and compiled it successfully

Superhepper:
Have you signed the commit(s)? I think I saw the DCO check failed.

Superhepper (approved):

ionut-arm (approved):

PR #532, opened 2024-07-29
Fixes lint problems reported by Clippy in Rust 1.80.0

Superhepper:
- In Rust 1.80.0 the `size_of` function is added to the preamble which causes lint errors when used with full qualifiers in the code so for that specific function the lint error has been suppressed.

- Other unused qualifiers and documentation lint errors have been fixed.

- Fixed redundant code in build.rs

- Fixed a configuration flag not being used correctly.

### DIFF
--- a/tss-esapi/src/structures/nv/storage/public.rs
+++ b/tss-esapi/src/structures/nv/storage/public.rs
@@ -27,6 +27,7 @@ pub struct NvPublic {
 }
 
 impl NvPublic {
+    #[allow(unused_qualifications)]
     const MAX_SIZE: usize = std::mem::size_of::<TPMS_NV_PUBLIC>();
 
     pub fn nv_index(&self) -> NvIndexTpmHandle {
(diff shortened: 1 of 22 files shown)


tss-esapi/src/structures/nv/storage/public.rs
```diff
@@ -27,6 +27,7 @@ pub struct NvPublic {
 }
 
 impl NvPublic {
+    #[allow(unused_qualifications)]
```
ionut-arm, line 30: Does the compiler complain if you remove the "unused" qualifications?

ionut-arm (approved):
Thanks!

tss-esapi/src/structures/nv/storage/public.rs
Superhepper: I am not sure. The call is specified using the full qualifications but because 1.80 puts the size_of in the preamble the qualifications do not get used and that is what it complains about.

I will test it.

tmfink:
I need these changes to build with Rust 1.80.0

Superhepper:
Yes, we just need to find another reviewer to have a look at it and approve it before it can be merged.

wiktor-k (approved):
It's okay for me too but if you can check the thing that Ionut asked for it'd be 👌

tss-esapi/src/structures/nv/storage/public.rs
Superhepper: You cannot remove it completely because then you will get an error.
And there is no 'qualifications' lint so you cannot write ```#[allow(qualifications)]```.

PR #536, opened 2024-08-02
File encryption and decryption example ; dev container with c library build and tpm emulator

nicolaspernoud:
Hello,

Here is a pull request with those two features that seems useful...

Best regards.


ionut-arm (commented):
Hey, thanks for the example and the devcontainer! I'd been looking at devcontainers for a while hoping we could have something like this, glad someone put it together :)

One thing to note is that you'll need to add your text file as [an exception](https://github.com/codespell-project/actions-codespell?tab=readme-ov-file#parameter-exclude_file) to the spellchecking.

You probably also need to wait for #532 to be merged :]

Other than that, LGTM

tss-esapi/examples/symmetric_file_encrypt_decrypt.rs
```diff
@@ -0,0 +1,217 @@
...
+        .with_st_clear(false)
+        .with_sensitive_data_origin(true)
+        .with_user_with_auth(true)
+        .with_sign_encrypt(true)
+        .with_decrypt(true)
+        .build()
+        .expect("Failed to build object attributes");
+
+    let key_pub = PublicBuilder::new()
+        // This key is a AES key
```
Superhepper, line 45: Minor:
Should it be 'an AES key'?

Superhepper (dismissed):

tss-esapi/examples/symmetric_file_encrypt_decrypt.rs
nicolaspernoud: It should... i corrected the typo.

nicolaspernoud:
@ionut-arm and @Superhepper : i made the changes, thanks for your feedbacks !

nicolaspernoud:
I updated using github web "update" button : it merges instead of rebasing, but that should be ok.

wiktor-k:
I think the only missing bit is running `cargo fmt` locally and committing and pushing the result :thinking:

ionut-arm:
Unfortunately I don't think we can merge as is now - you'll need to rebase and get all your commits on top of @Superhepper's, otherwise we'll have his commit twice.

Superhepper (approved):

PR #537, opened 2024-08-03
WIP: adds a `signature::Signer` interface

baloo:
This brings an implementation of a [`signature::Signer`](https://docs.rs/signature/2.2.0/signature/trait.Signer.html) for keys stored on the TPM.

This is intend to make for easier re-use of this crate and to allow to:
 - Create certificates https://docs.rs/x509-cert/latest/x509_cert/builder/struct.CertificateBuilder.html
 - Create CSRs https://docs.rs/x509-cert/latest/x509_cert/builder/struct.RequestBuilder.html
 - Sign documents with S/MIME https://docs.rs/cms/latest/cms/builder/index.html

Here is an implementation of an SSH agent making use of that infrastructure: https://github.com/wiktor-k/ssh-agent-lib/pull/87


tss-esapi/src/abstraction/public.rs
```diff
@@ -3,126 +3,232 @@
...
 
-    fn try_from(value: Public) -> Result<Self, Self::Error> {
-        public_to_decoded_key(&value)
+    fn try_from(value: &Public) -> Result<Self, Self::Error> {
+        match value {
+            Public::Rsa {
+                unique, parameters, ..
+            } => {
+                let exponent = match parameters.exponent() {
+                    RsaExponent::ZERO_EXPONENT => BigUint::from(65537u32),
```
Superhepper, line 74: Can you make this into a constant instead with a name that makes it more clear what it is. Because it does not seem to be ```u16::MAX```.

tss-esapi/src/abstraction/public.rs
```diff
@@ -3,126 +3,232 @@
...
+    }
+}
+
+impl TryFrom<&TpmPublicKey> for RsaPublicKey {
+    type Error = Error;
+
+    fn try_from(value: &TpmPublicKey) -> Result<Self, Self::Error> {
+        match value {
+            TpmPublicKey::Rsa(modulus) => {
+                let exponent = BigUint::from(65537u32);
```
Superhepper, line 187: Same thing here

tss-esapi/src/abstraction/transient/signer.rs
```diff
@@ -0,0 +1,169 @@
+// Copyright 2019 Contributors to the Parsec project.
```
Superhepper, line 1: This is a completely new file so this should probably be 2024?

tss-esapi/src/interface_types/algorithm.rs
```diff
@@ -76,6 +76,52 @@ impl TryFrom<TPMI_ALG_HASH> for HashingAlgorithm {
     }
 }
 
+/// Provides the value of the digest used in this crate for the digest.
```
Superhepper, line 79: I am not sure this is the best place to put this. This file is mainly containing the TPMI_ALG stuff.

tss-esapi/src/structures/signatures.rs
```diff
@@ -143,3 +152,31 @@ impl TryFrom<TPMS_SIGNATURE_ECC> for EccSignature {
         })
     }
 }
+
+#[cfg(feature = "abstraction")]
+impl<C> TryFrom<EccSignature> for ecdsa::Signature<C>
```
Superhepper, line 157: If this is actually abstraction then put under the abstraction module.

tss-esapi/tests/integration_tests/abstraction_tests/public_tests.rs
```diff
@@ -70,44 +69,38 @@ mod public_rsa_test {
...
             .expect("Failed to convert Public structure to SubjectPublicKeyInfo (RSA).");
-        assert_eq!(key.algorithm, AlgorithmIdentifier::new_rsa_encryption());
-        match key.subject_public_key {
-            PublicKey::Rsa(key) => {
-                assert_eq!(key.public_exponent, default_exponent);
-                assert_eq!(key.modulus.as_unsigned_bytes_be(), RSA_KEY)
-            }
-            _ => panic!("PublicKey of SubjectPublicKeyInfo is not an instance for RSA"),
-        }
+        let default_exponent = BigUint::from(65537_u32);
```
Superhepper, line 84: Same thing here try to use a constant instead.

Superhepper (commented):
Sorry for reviewing a WIP though I found it a little bit interesting so I couldn't help my self. Feel free to disregard anything I have  commented on.

tss-esapi/tests/integration_tests/abstraction_tests/public_tests.rs
baloo: I don't want to expose the constant in the API, I will redefine it.

baloo:
> Sorry for reviewing a WIP though I found it a little bit interesting so I couldn't help my self. Feel free to disregard anything I have commented on.

Oh, no, thanks for reviewing it!
I sent it a bit early as this is work I've been putting off for almost a year now. I'm doing it mostly to get a sense whether this is a direction the project could go or not.

ionut-arm (commented):
Thank you for bringing the patch! The overall design looks good to me.

tss-esapi/src/abstraction/signatures.rs
```diff
@@ -0,0 +1,72 @@
...
+            FieldBytes::<C>::from_slice(r).clone(),
+            FieldBytes::<C>::from_slice(s).clone(),
+        )
+        .map_err(|_| Error::local_error(WrapperErrorKind::InvalidParam))?;
+        Ok(signature)
+    }
+}
+
+// Note: this does not implement `TryFrom<RsaSignature>` because `RsaSignature` does not carry the
+// information whether the signatures was generated using PKCS#1v1.5 or PSS.
```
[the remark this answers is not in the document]
Superhepper: No it is not really a problem as long as it does not causes any ambiguities in the conversions between ```TPMS_SIGNATURE_RSA``` and ```RsaSignature```.

tss-esapi/src/abstraction/transient/signer.rs
```diff
@@ -0,0 +1,170 @@
...
+    pub fn key_params_default() -> KeyParams
+    where
+        C: DigestPrimitive,
+        <C as DigestPrimitive>::Digest: AssociatedHashingAlgorithm,
+    {
+        Self::key_params::<C::Digest>()
+    }
+
+    /// Key parameters for this curve
+    pub fn key_params<D>() -> KeyParams
```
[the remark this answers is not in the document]
baloo: `C` describes the curve (could be `NistP256`, `NistP384`, `NistP521`, ...) those are all supported here. When using the signer (through the `Ecdsa` struct) you would specify which curve you're using.

This would pick the correct parameters to specify to the TPM when signing, the size of the object that comes back from signature, how to verify them, ...

This function just creates the TPM parameters related to this curve and the selected digest.

tss-esapi/src/abstraction/transient/signer.rs
```diff
@@ -0,0 +1,170 @@
...
+    generic_array::ArrayLength,
+    ops::Invert,
+    sec1::{FromEncodedPoint, ModulusSize, ToEncodedPoint},
+    subtle::CtOption,
+    AffinePoint, CurveArithmetic, FieldBytesSize, PrimeCurve, PublicKey, Scalar,
+};
+use signature::{DigestSigner, Error as SigError, KeypairRef};
+
+#[derive(Debug)]
+pub struct Ecdsa<'ctx, C>
```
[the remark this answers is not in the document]
baloo: I tried to add some, and add a code sample in the doc as well.

tss-esapi/src/abstraction/signatures.rs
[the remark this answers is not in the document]
baloo: I think it's fine like that, the comment was more for a future self why we would not have such a `TryFrom`

tss-esapi/tests/integration_tests/abstraction_tests/transient_key_context_tests.rs
```diff
@@ -875,3 +889,78 @@ fn get_random_from_tkc() {
...
+        .create_key(EcSigner::<NistP256>::key_params_default(), 0)
+        .expect("create private key");
+
+    let subject = Name::from_str("CN=tpm.example").expect("Parse common name");
+    let signer = EcSigner::<NistP256>::new(&mut ctx, tpm_km, None).expect("Create a signer");
+    let builder = RequestBuilder::new(subject, &signer).expect("Create certificate request");
+
+    let cert_req = builder
+        .build::<p256::ecdsa::DerSignature>()
+        .expect("Sign a CSR");
```
baloo, line 908: I added proper examples of what this interface is used for.

tss-esapi/tests/integration_tests/abstraction_tests/transient_key_context_tests.rs
```diff
@@ -875,3 +889,78 @@ fn get_random_from_tkc() {
...
+// NOTE(baloo): I believe this is a legitimate case, but for a reason I don't understand yet
+// the TPM will reply with `Error in creating derived key: 0x000002C3` when trying to create a
+// P-256 key with Sha3-256 associated hash.
+// This is combination defined by NIST:
+// https://csrc.nist.gov/CSRC/media/Projects/Cryptographic-Standards-and-Guidelines/documents/examples/P256_SHA3-256.pdf
+//
+// I can't find where in the TPM spec this would be an illegal pair, we'll ignore for now.
+#[ignore]
+#[test]
+fn sign_p256_sha3_256() {
```
baloo, line 946: This one is a bit puzzling to me. I can't find where in the spec the the P-256/SHA3-256 pair would be illegal. If anyone has an idea, I'd love to know.

tss-esapi/src/abstraction/hashing.rs
```diff
@@ -0,0 +1,50 @@
...
+
+use crate::interface_types::algorithm::HashingAlgorithm;
+
+/// Provides the value of the digest used in this crate for the digest.
+pub trait AssociatedHashingAlgorithm {
+    /// Value of the digest when interacting with the TPM.
+    const TPM_DIGEST: HashingAlgorithm;
+}
+
+#[cfg(feature = "sha1")]
```
wiktor-k, line 12: Hmm... how does this work if I don't see the feature declared in Cargo.toml? :thinking:

tss-esapi/tests/integration_tests/abstraction_tests/transient_key_context_tests.rs
ionut-arm: Could it be a limitation of the TPM you're using? From the [Structures spec](https://trustedcomputinggroup.org/wp-content/uploads/TCG_TPM2_r1p59_Part2_Structures_pub.pdf), table 8, it seems SHA3-256 is classified as Assigned rather than TCG standard. I assume this might mean that manufacturers are not required to implement it (?)

tss-esapi/src/abstraction/transient/signer.rs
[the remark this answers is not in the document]
ionut-arm: Sure, I think I was mostly confused by the existence of two nearly-identical methods before, and that the struct they're tacked to doesn't represent (just) a specific curve.

tss-esapi/tests/integration_tests/abstraction_tests/transient_key_context_tests.rs
baloo: I'm using swtpm (0.8.2) with libtpms (0.9.6) I expected it to play nicely. I haven't took time to look in libtpms yet.

tss-esapi/src/abstraction/hashing.rs
baloo: yeaaah, I learned this one recently too lol.
https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies
```
By default, this optional dependency implicitly defines a feature that looks like this:

[features]
gif = ["dep:gif"]
```

I'm abusing that mechanism here.

I actually need to split the features in the `Cargo.toml` to look more like:
```
rustcrypto = ["ecdsa", "elliptic-curve", "signature", "x509-cert"]
rustcrypto-full = ["rustcrypto", "p192", "p224", "p256", "p384", "p521", "rsa", "sha1", "sha2", "sha3", "sm2", "sm3"]
```

That way you can bring in rustcrypto + "p256" + "sha2" if you only care about those.

tss-esapi/tests/integration_tests/abstraction_tests/transient_key_context_tests.rs
ionut-arm: Hmm, from [this](https://github.com/stefanberger/libtpms/blob/4b89a028dd55a73d0c36ba2a1d8c93a624994870/src/tpm2/TpmProfile_Common.h#L217) it seems libtpms can block SHA3 at compile-time, perhaps that's what is going on here.

baloo: yeah, turns out it's more complicated than that.
https://github.com/stefanberger/libtpms/issues/206

No sha3 support for now.

tss-esapi/src/abstraction/transient/signer.rs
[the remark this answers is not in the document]
baloo: I did bring back the two nearly-identical methods btw.

baloo:
The two remaining failing tests should be fixed by https://github.com/parallaxsecond/rust-tss-esapi/pull/562 I'll rebase once that merges.

tss-esapi/src/abstraction/transient/signer.rs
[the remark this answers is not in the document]
baloo: The `key_params_default` just use the "default" hashing mechanism for a given curve (P256 would use sha256, P384 would use sha384, etc).

The hashing methods needs to line up otherwise the curve has a higher "security score" than the hashing mechanism it uses and it's a waste of space.

Superhepper (approved):
I think it looks fine though my experience with these things are quite limited so I mainly focused on the code structure and concepts.

One thing that will always be discussed is conversions from reference types. I think it is ok here because I do not see any `clone` or `to_owned` any where.

wiktor-k (approved):
I think it looks good, a couple of question-nits...

tss-esapi/src/abstraction/public.rs
```diff
@@ -2,127 +2,247 @@
 // SPDX-License-Identifier: Apache-2.0
```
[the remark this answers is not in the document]
baloo: wait? which header?

tss-esapi/src/abstraction/public.rs
```diff
@@ -2,127 +2,247 @@
...
+{
+    type Error = Error;
+
+    fn try_from(value: &TpmPublicKey) -> Result<Self, Self::Error> {
+        match value {
+            TpmPublicKey::Ecc { x, y } => {
+                let x = x.as_slice();
+                let y = y.as_slice();
+
+                // TODO: When elliptic_curve bumps to 0.14, we can use the TryFrom implementation instead
```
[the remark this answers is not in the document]
baloo: When that happens, the line 11 of this file will have to change (from `GenericArray` to `Array`), and I hoped this would be enough to be taken care of.

tss-esapi/src/abstraction/transient/signer.rs
```diff
@@ -0,0 +1,254 @@
...
+        let signature = context
+            .sign(
+                self.key_material.clone(),
+                key_params,
+                self.key_auth.clone(),
+                digest,
+            )
+            .map_err(SigError::from_source)?;
+        let TpmSignature::EcDsa(signature) = signature else {
+            todo!();
```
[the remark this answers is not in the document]
baloo: I missed this one, sorry

tss-esapi/src/abstraction/public.rs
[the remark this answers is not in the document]
baloo: (the header is on line 1, even if hidden from the diff)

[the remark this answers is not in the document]
wiktor-k: > (the header is on line 1, even if hidden from the diff)

Facepalm. Just... pretend you didn't see this :face_with_peeking_eye:

tss-esapi/src/abstraction/public.rs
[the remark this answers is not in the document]
baloo: When upgrading the `TryFrom` are not actually usable. Those are only useful for compressed points, those are uncompressed.

rucoder:
@baloo pkcs8::EncodePublicKey is included only when "rsa" feature is enabled however it is required for `impl TryFrom<&Public> for SubjectPublicKeyInfoOwned`  
```
                #[cfg(all(feature = "rustcrypto", feature = "p192"))]
                read_key!(EccCurve::NistP192, p192::NistP192);
                #[cfg(all(feature = "rustcrypto", feature = "p224"))]
                read_key!(EccCurve::NistP224, p224::NistP224);
                #[cfg(all(feature = "rustcrypto", feature = "p256"))]
                read_key!(EccCurve::NistP256, p256::NistP256);
                #[cfg(all(feature = "rustcrypto", feature = "p384"))]
                read_key!(EccCurve::NistP384, p384::NistP384);
                #[cfg(all(feature = "rustcrypto", feature = "p521"))]
                read_key!(EccCurve::NistP521, p521::NistP521);
                #[cfg(all(feature = "rustcrypto", feature = "sm2"))]

```

tss-esapi/src/abstraction/public.rs
```diff
@@ -2,127 +2,239 @@
...
-#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
-pub enum DecodedKey {
-    RsaPublicKey(RsaPublicKey),
-    EcPoint(EcPoint),
+use x509_cert::spki::SubjectPublicKeyInfoOwned;
+
+#[cfg(all(feature = "rustcrypto", feature = "rsa"))]
+use {
+    crate::structures::RsaExponent,
+    rsa::{pkcs8::EncodePublicKey, BigUint, RsaPublicKey},
```
rucoder, line 21: `rsa::{pkcs8::EncodePublicKey` is also required for elliptic curves not only for RSA

tss-esapi/src/abstraction/public.rs
```diff
@@ -2,127 +2,239 @@
...
+                        }
+                    };
+                }
+
+                #[cfg(all(feature = "rustcrypto", feature = "p192"))]
+                read_key!(EccCurve::NistP192, p192::NistP192);
+                #[cfg(all(feature = "rustcrypto", feature = "p224"))]
+                read_key!(EccCurve::NistP224, p224::NistP224);
+                #[cfg(all(feature = "rustcrypto", feature = "p256"))]
+                read_key!(EccCurve::NistP256, p256::NistP256);
```
rucoder, line 136: e.g. required here or the build failed

rucoder (changes requested):

tss-esapi/src/abstraction/public.rs
baloo: good point, thanks!

rucoder:
@baloo why EcSigner depends on TransientKeyContext? The only method is used there is  sign(). Besides `TransientKeyContext` is limited to RSA key as a primary. it would be better IMO either 
1. to have a constructor that that accepts `Context` and `KeyHandle` in TransientKeyContext/Builder. Currently fields are private
2.  Or maybe turn (partially) TransientKeyContext into a trait since EcSigner uses its methods (actually only sign() so far).

baloo:
This is a good point. I think I like the idea of making a trait but I'm not sure how to do it yet, `TransientKeyContext` and `Context::sign` are pretty dissimilar.

baloo:
I have a rough follow-up PR for the trait option: https://github.com/baloo/rust-tss-esapi/pull/1

baloo:
It was also missing the implementation for an RSA signer: https://github.com/baloo/rust-tss-esapi/pull/2

tss-esapi/src/structures/buffers.rs
```diff
@@ -220,6 +226,34 @@ pub mod digest {
             Digest(value_as_vec.into())
         }
     }
+
+    macro_rules! impl_from_digest {
+        ($($size:ty),+) => {
+            $(impl From<GenericArray<u8, $size>> for Digest {
+                fn from(value: GenericArray<u8, $size>) -> Self {
+                    Digest(value.as_slice().to_vec().into())
```
Superhepper, line 234: This is not zeroizing the digest data that you are converting from.

tss-esapi/src/abstraction/public.rs
```diff
@@ -2,127 +2,253 @@
...
+    )
+))]
+use pkcs8::EncodePublicKey;
+
+/// Default exponent for RSA keys.
+// Also known as 0x10001
+#[cfg(all(feature = "rustcrypto", feature = "rsa"))]
+const RSA_DEFAULT_EXP: u64 = 65537;
+
+impl<C> TryFrom<&Public> for PublicKey<C>
```
ionut-arm, line 43: Should this (and all the other similar `impl`s below) not be gated on the `rustcrypto` feature being enabled?

ionut-arm (commented):
Looks pretty good overall!

Are you intending to add signing support for RSA too?

ionut-arm:
Ah, one thing that I keep forgetting - could you please document the new feature (in the README)?

tss-esapi/src/structures/buffers.rs
baloo: good point.

tss-esapi/Cargo.toml
```diff
@@ -33,9 +34,23 @@ hostname-validator = "1.1.0"
...
+p384 = { version = "0.13.0", optional = true }
+p521 = { version = "0.13.3", optional = true }
+pkcs8 = { version = "0.10.2", optional = true }
+rsa = { version = "0.9", optional = true }
+sha1 = { version = "0.10.6", optional = true }
+sha2 = { version = "0.10.8", optional = true }
+sha3 = { version = "0.10.8", optional = true }
+sm2 = { version = "0.13.3", optional = true }
+sm3 = { version = "0.4.2", optional = true }
+digest = "0.10.7"
```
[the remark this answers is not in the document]
baloo: Fixed

tss-esapi/src/abstraction/public.rs
baloo: no, because it's all gated by `abstraction` which pulls `rustcrypto`, but I can rework the `cfg` flags.

tss-esapi/src/abstraction/hashing.rs
wiktor-k: `rustcrypto` and `rustcrypto-full` are fine for me.

Would you be so kind to adjust all of these to use the `rustcrypto-full` feature name? (maybe this stuff only bothers me :sweat_smile:

tss-esapi/src/abstraction/public.rs
```diff
@@ -2,127 +2,250 @@
...
+                Ok(public_key
+                    .to_public_key_der()
+                    .map_err(|_| Error::local_error(WrapperErrorKind::InvalidParam))?
+                    .decode_msg::<Self>()
+                    .map_err(|_| Error::local_error(WrapperErrorKind::InvalidParam))?)
+            }
+            #[allow(unused)]
+            Public::Ecc { parameters, .. } => {
+                macro_rules! read_key {
+                    ($curve:expr, $key_type:ty) => {
```
wiktor-k, line 129: Why do we need `$curve` if it's not used anywhere? (also the `allow(unused)` lint)

wiktor-k (dismissed):
I think this looks good. While I'm no expert in rust crypto I'm of the opinion that this is ready to be shipped to folks to try out in the wild.

tss-esapi/src/abstraction/hashing.rs
baloo: In this instance this is gated by a `#[cfg(feature = "rustcrypto")]` on the module itself, this just further restrict the cfg.

But the intent here is that as a library consumer you may only enable (and pull dependencies for) the specific algorithms you care for.

That is:
```
tss-espapi = { version = "...", features = ["rustcrypto", "sha1", "rsa"] }
```

Or just opt-in everything:
``
tss-espapi = { version = "...", features = ["rustcrypto-full"] }
```

tss-esapi/src/abstraction/public.rs
baloo: The `allow(unused)` is if you compile with `features = ["rustcrypto", "rsa"]` and opt-out of ECC, then the `read_key` macro gets unused.

tss-esapi/src/abstraction/hashing.rs
wiktor-k: Okay, if this works then it's really nifty. Thanks for explaining! :+1:

baloo:
anything else needed here?

wiktor-k (approved):

baloo:
@rucoder just so you know, there are a bunch of improvements in the upcoming release of `x509-cert` and we changed everything to a trait based profile.
I'm the author for the x509-cert builder (and would love feedback).

The trait-based approach proved much more flexible for us.
We expect the stable release of it in 3-4 months.

https://github.com/parallaxsecond/rust-tss-esapi/pull/563 should pull the pre-releases if you wanted to try that.
I published the `Cargo.lock` here: https://github.com/baloo/rust-tss-esapi/tree/baloo/make-credentials%2Block 
(you will need the `Cargo.lock` as we're migrating the ecosystem to `rand_core 0.9` the ecosystem is a construction site at the moment :D)

ionut-arm (approved):
It looks good enough to me, let's get it out there and get it tested in the wild

PR #538, opened 2024-08-06
Fix serde test features

Firstyear:


Superhepper (dismissed):

Superhepper (commented):
I wonder if it would be possible to achieve this using ```Resolver = 2``` instead in the cargo.toml?

Something like this:
> [package]
> name = "B"
> resolver = "2"
>
> [features]
> serde = []
> abstractions = []
> integration-tests = []
>
> [dev-dependencies]
> B = { path = ".", features = ["test-utils", "abstractions", "integration-tests"] }

I am not sure this is a good idea or not. But I think it is worth trying. Because I think it is a good idea to build all the tests as standard and if you want to be clever you can add some command line arguments.

Superhepper (approved):

Superhepper:
I will try to work on the Cargo configuration things in another PR.

wiktor-k (approved):
Looks good :+1: thanks!

PR #539, opened 2024-08-07
Fix `cargo test` compilation 

tmfink:
If "serde" or "abstraction" features were not set, then the tests would not compile

### DIFF
--- a/tss-esapi/README.md
+++ b/tss-esapi/README.md
@@ -27,6 +27,7 @@ The crate currently offers the following features:
 * `abstraction` (enabled by default) - provides a set of abstracted primitives
   on top of the basic Rust-native ESAPI API provided by the crate. This feature
   can be turned off to reduce the number of dependencies built.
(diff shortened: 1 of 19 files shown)


Superhepper:
This is related to the #538 .

Normally when I build the tests I just specify all features because I want to build everything but I think it is not very intuitive to have to specify features you do not even know exists when building. So I would really like for all features to be default. So in the PR I mentioned I suggested to change to resolver 2 to specify the features in dev-depedencies. And hopefully make all this markup unnecessary.

So it would be nice to see a PR that tries to solve the problem using the suggested method in 538 pr.

Superhepper:
This issue has been resolved in #543. I added you as co-author on that PR hope that works for you. I'll be closing this PR.

tmfink:
> This issue has been resolved in #543. I added you as co-author on that PR hope that works for you. I'll be closing this PR.

Sure--no worries.

PR #540, opened 2024-08-07
Migrate to 2021 edition

### LINKED ISSUE #516: Update Rust edition.
The 2018 rust edition is getting quite old and it is starting to cause some problems with dependencies so it is time to update it to 2021.

tmfink:
Resolves #516

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,2 +1,3 @@
 [workspace]
+resolver = "2"
 members = ["tss-esapi", "tss-esapi-sys"]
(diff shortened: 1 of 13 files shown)


Superhepper (dismissed):
I am all for it. If necessary we can make the 7.x branch work for older edition.

wiktor-k:
I'm also good :+1: but the TODO says we should take a look at the lints that were disabled... maybe that's something that you could look into @tmfink ? Thanks for your time! :bow:

ionut-arm (dismissed):
Happy with the change, but I also support @wiktor-k's query

PR #543, opened 2024-08-30
Fixes problem with running tests needing features to be specified.

Superhepper:
This PR builds upon the work made in #539 and #538 and fixes a problem with running tests required
several features to be specified in order for the
tests to compile.

This has been accomplish by doing the following:
1. Tests have been refactored in order to avoid dependency on features that may and may not be available.
2. Updated workspace Cargo.toml to use version 2 resolver.
3. Updated tss-esapi Cargo.toml to use the crate it self as dev-dependency with most of the features enabled.

### DIFF
--- a/tss-esapi/tests/integration_tests/context_tests/tpm_commands/integrity_collection_pcr_tests.rs
+++ b/tss-esapi/tests/integration_tests/context_tests/tpm_commands/integrity_collection_pcr_tests.rs
@@ -4,7 +4,6 @@ mod test_pcr_extend_reset {
     use crate::common::create_ctx_with_session;
     use std::convert::TryFrom;
     use tss_esapi::{
-        abstraction::pcr::PcrData,
         handles::PcrHandle,
         interface_types::algorithm::HashingAlgorithm,
         structures::{Digest, DigestValues, PcrSelectionListBuilder, PcrSlot},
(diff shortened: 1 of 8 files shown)


wiktor-k (approved):

tss-esapi/tests/integration_tests/context_tests/tpm_commands/integrity_collection_pcr_tests.rs
```diff
@@ -34,6 +33,12 @@ mod test_pcr_extend_reset {
         });
 
         // Needs to have the length of associated with the hashing algorithm
+        assert_eq!(read_pcr_selections.get_selections().len(), 2);
+        assert_eq!(
+            pcr_selection_list.get_selections(),
+            read_pcr_selections.get_selections()
+        );
```
ionut-arm, line 40: Are these `len` checks not redundant, given the other asserts? It might make debugging easier if the issue is a length discrepancy, I guess...

ionut-arm (approved):
Thanks for the cleanup! I only had one question, but it's not a blocker

tss-esapi/tests/integration_tests/context_tests/tpm_commands/integrity_collection_pcr_tests.rs
Superhepper: Maybe I just wanted to be really sure there were no missmatch.

PR #545, opened 2024-09-13
Adds a TBS TctiNameConf.

Superhepper:
When #477 got merged it became possible to build
using a path to the ```tpm2-tss``` installation
instead of depending on ```pkg-config```.

This made it possible to build under Windows. To
further increase the support for the windows
platform this commit moves the option for TBS
TCTI that is being introduced in #523 into a
separate commit.

This commit also updates the documentation
regarding building using an installation
folder.

Co-author: Thomas Thomas Epperson <thomas.epperson@gmail.com>


ionut-arm (approved):
Awesome stuff :)

wiktor-k:
Thanks for the PR. It's something that's really needed.

I tried building it on my Windows computer and after making the correct folder structure (setting up correct OpenSSL dirs and moving files around was a chore but doable) I still get the error: `couldn't find any v alid shared libraries matching ['clang.dll']...` (when doing `--generate-bindings` without that it complains about unsupported tuple).

I'm wondering how far should the instructions go explaining all steps required :thinking:

uglyoldbob:
@wiktor-k I should be able to test it on Saturday morning.

wiktor-k:
> I should be able to test it on Saturday morning.

Thank you! :pray: It would be greatly appreciated!

Superhepper:
I tested it with a test app. And it seems to work fine for me.

https://github.com/Superhepper/rust-tss-esapi-test-app

wiktor-k:
> I tested it with a test app. And it seems to work fine for me.
> 
> https://github.com/Superhepper/rust-tss-esapi-test-app

Okay, I'm going to approve it not to have the code bitrot. Still, if you'd be able to confirm it's working even afterwards @uglyoldbob it'd be nice. All in all I don't think this PR is particularly dangerous and it clearly improves things.

uglyoldbob:
Oh it looks like tpm2-tss was updated to supported visual studio 2019/2022. I was expecting the bundled feature to still be in this pull request. I'll have to take some more time later to do the support for that.

Superhepper:
No as I

> Oh it looks like tpm2-tss was updated to supported visual studio 2019/2022. I was expecting the bundled feature to still be in this pull request. I'll have to take some more time later to do the support for that.

No, as I mentioned. This was only for the TCTI part so that it will be possible to run things you build from an installation folder on windows as well. If you want I can try to adapt the `bundled` feature code as well?

PR #549, opened 2024-09-25
Fixes issue wrong max size in some sized buffer types.

### LINKED ISSUE #548: Conversion from tss-esapi native buffer type into the corresponding tss-esapi-sys TSS type can panic
Hello,

I identified native buffer types whose MAX_SIZE is larger than the buffer size of corresponding TSS types (TPM2B). This is an issue because the infallible conversion From<$native_type> for $tss_type is implemented.  When the native buffer is too big, the conversion method panics.

Affected buffer types are : 
* IdObject / TPM2B_ID_OBJECT
* SensitiveData / TPM2B_SENSITIVE_DATA

Example of code that triggers the issue : 
```rs
#!/usr/bin/env cargo-eval

//! ```cargo
//! [dependencies]
//! tss-esapi = { version = "7.5.1" }
//! ```

use tss_esapi::structures::IdObject;
use tss_esapi::tss2_esys::TPM2B_ID_OBJECT;
use std::hint::black_box;

fn main() {
    // IdObject::MAX_SIZE: usize = 256usize
    let id_object = IdObject::try_from(vec![0u8; IdObject::MAX_SIZE]).expect("this works");
    //
    // #[repr(C)]
    // pub struct TPM2B_ID_OBJECT {
    //     pub size: UINT16,
    //     pub credential: [BYTE; 132],
    // }
    //
    // The following conversion causes 
    // thread 'main' panicked at /home/vscode/.cargo/registry/src/github.com-1ecc6299db9ec823/tss-esapi-7.5.1/src/structures/buffers.rs:185:5:
    // range end index 256 out of range for slice of length 132
    let buffer_id_object: TPM2B_ID_OBJECT = id_object.into();
    black_box(buffer_id_object);
}
```

#### Security concerns

An unexpected panic can cause a program to abort unexpectedly, potentially leading to a denial of service (DoS) vulnerability. However in this context, it is unlikely that the affected buffers are untrusted, so I don't think there is much of a security concern.

Superhepper:
Some of the sized buffers had their buffer sizes
set as numbers. Even though this in some cases
were the correct numbers they were a little hard
to determine if they actually followed the size
specified in the standard. So this PR fixes #548
in the main branch by using the the calculations
specified in the standard for the buffer sizes.


tss-esapi/tests/integration_tests/structures_tests/buffers_tests/encrypted_secret_test.rs
```diff
@@ -0,0 +1,15 @@
...
+fn test_max_sized_ecc_parameter_conversions() {
+    let expected_secret = [0xffu8; EncryptedSecret::MAX_SIZE];
+    let native = EncryptedSecret::try_from(expected_secret.as_slice().to_vec()).expect(
+        "It should be possible to convert an array of MAX size into a EncryptedSecret object.",
+    );
+    let tss = TPM2B_ENCRYPTED_SECRET::from(native);
+    assert_eq!(EncryptedSecret::MAX_SIZE, tss.size as usize);
+    // This will be a compiler error if the max size does not match the TSS buffer size.
+    assert_eq!(expected_secret, tss.secret);
+}
```
ionut-arm, line 15: I guess in the future we can make a macro out of this sort of test.

ionut-arm (approved):
Thanks for the effort, that was very quick!

wiktor-k (approved):
Looks pretty straightforward, thanks for the work! :clap:

PR #551, opened 2024-10-17
Add ctx.get_time

nilsb99:
For a time-based remote attestation project in a uni class, we needed this function and wanted to contribute it to upstream.

We followed the existing implementation of `ctx.quote()` to write `ctx.get_time()` and its test.

### DIFF
--- a/tss-esapi/src/context/tpm_commands/attestation_commands.rs
+++ b/tss-esapi/src/context/tpm_commands/attestation_commands.rs
@@ -6,13 +6,12 @@ use crate::{
         Attest, AttestBuffer, CreationTicket, Data, Digest, PcrSelectionList, Signature,
         SignatureScheme,
     },
-    tss2_esys::{Esys_Certify, Esys_Quote},
+    tss2_esys::{Esys_Certify, Esys_CertifyCreation, Esys_GetTime, Esys_Quote},
     Context, Result, ReturnCode,
 };
 use log::error;
 use std::convert::TryFrom;
 use std::ptr::null_mut;
-use tss_esapi_sys::Esys_CertifyCreation;
 
 impl Context {
     /// Prove that an object is loaded in the TPM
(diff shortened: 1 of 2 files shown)


tss-esapi/src/context/tpm_commands/attestation_commands.rs
```diff
@@ -6,13 +6,13 @@ use crate::{
...
     },
-    tss2_esys::{Esys_Certify, Esys_Quote},
+    tss2_esys::{Esys_Certify, Esys_GetTime, Esys_Quote},
     Context, Result, ReturnCode,
 };
 use log::error;
 use std::convert::TryFrom;
 use std::ptr::null_mut;
-use tss_esapi_sys::Esys_CertifyCreation;
+use tss_esapi_sys::{Esys_CertifyCreation, ESYS_TR_RH_ENDORSEMENT};
```
Superhepper, line 15: I know it is not you who added this but. I do not think we should import using the ```tss_esapi_sys``` but instead use the ```tss_esys``` alias. So the ```Esys_CertifyCreation``` should be moved.

tss-esapi/src/context/tpm_commands/attestation_commands.rs
```diff
@@ -322,8 +322,47 @@ impl Context {
...
+        qualifying_data: Data,
+        signing_scheme: SignatureScheme,
+    ) -> Result<(Attest, Signature)> {
+        let mut timeinfo_ptr = null_mut();
+        let mut signature_ptr = null_mut();
+        ReturnCode::ensure_success(
+            unsafe {
+                Esys_GetTime(
+                    self.mut_context(),
+                    ESYS_TR_RH_ENDORSEMENT,
```
Superhepper, line 341: In general I think one should avoid using the TSS constants directly. So I would suggest writing this as ```ObjectHandle::Endorsement.into()``` or something similar.

tss-esapi/src/context/tpm_commands/attestation_commands.rs
```diff
@@ -322,8 +322,47 @@ impl Context {
         ))
     }
 
+    /// Get the current time and clock from the TPM
+    ///
```
Superhepper, line 326: It would be excellent if you could add an runable example and documentation of the parameters.

Superhepper (changes requested):
Thank you so much for the PR. I had some comments.

Superhepper (approved):

tss-esapi/src/context/tpm_commands/attestation_commands.rs
```diff
@@ -322,8 +321,120 @@ impl Context {
...
+        let mut timeinfo_ptr = null_mut();
+        let mut signature_ptr = null_mut();
+        ReturnCode::ensure_success(
+            unsafe {
+                Esys_GetTime(
+                    self.mut_context(),
+                    ObjectHandle::Endorsement.into(),
+                    signing_key_handle.into(),
+                    self.optional_session_1(),
+                    self.optional_session_2(),
```
ionut-arm, line 416: I think these two are actually required, the first one for the endorsement, the second for the signing key. Looking at `quote` above I wonder if that's a bug, AFAICT the spec asks for one authorization session at least there.

ionut-arm (commented):
Awesome, thanks for the patch, it looks pretty good! Amazing that you've worked with something like this in Uni :)

tss-esapi/src/context/tpm_commands/attestation_commands.rs
Superhepper: *TCG TSS 2.0 Enhanced System API (ESAPI) Specification, Version 1.00, Revision 14, October 1, 2021*
> 11.3.45 Esys_GetTime Functions
> TSS2_DLL_EXPORT TSS2_RC Esys_GetTime_Async(
>    ESYS_CONTEXT *esysContext,
>    ESYS_TR privacyAdminHandle,
>    ESYS_TR signHandle,
>    ESYS_TR privacyAdminHandleSession1,
>    ESYS_TR signHandleSession2,
>    ESYS_TR optionalSession3,
>    TPM2B_DATA const *qualifyingData,
>    TPMT_SIG_SCHEME const *inScheme);

I think you are right only the third session is optional.

Superhepper: Regarding `Esys_Quote`.

*TCG TSS 2.0 Enhanced System API (ESAPI) Specification, Version 1.00, Revision 14, October 1, 2021*
> 11.3.42     Esys_Quote Functions
> TSS2_DLL_EXPORT TSS2_RC Esys_Quote_Async(
>     ESYS_CONTEXT *esysContext,
>     ESYS_TR signHandle,
>     ESYS_TR signHandleSession1,
>     ESYS_TR optionalSession2,
>     ESYS_TR optionalSession3,
>     TPM2B_DATA const *qualifyingData,
>     TPMT_SIG_SCHEME const *inScheme,
>    TPML_PCR_SELECTION const *PCRselect);

So it seems Quote is wrong as well because the first session should be required.

Superhepper (approved):

Superhepper:
@nilsb99 I did not ask this before. But do you need this in an official release? Because if you do I suggest you cherry-pick this to the 7.x.y branch in a new PR.

ionut-arm (approved):
🚀

PR #552, opened 2024-10-18
Align ek abstraction with TCG EK Credential profile

THS-on:
The used hashing and symmetric algorithm and authPolicy are different based on the key size/curve.

While the old generated EKs still work, they are not that useful as they don't match the EK found in provisioned EK certificates according to the TCG EK Credential profile. This fixes that.

### DIFF
--- a/tss-esapi/src/abstraction/ek.rs
+++ b/tss-esapi/src/abstraction/ek.rs
@@ -12,9 +12,9 @@ use crate::{
         reserved_handles::{Hierarchy, NvAuth},
     },
     structures::{
-        Digest, EccParameter, EccPoint, EccScheme, KeyDerivationFunctionScheme, Public,
-        PublicBuilder, PublicEccParametersBuilder, PublicKeyRsa, PublicRsaParametersBuilder,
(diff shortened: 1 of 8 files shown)


THS-on:
What is currently missing is reading the template and potential nonce from NV. Do we need that feature?

Superhepper (approved):
Thank you!
LGTM.

tss-esapi/src/abstraction/ek.rs
```diff
@@ -42,12 +66,19 @@ pub fn create_ek_public_from_default_template<IKC: IntoKeyCustomization>(
 ) -> Result<Public> {
     let key_customization = key_customization.into_key_customization();
 
+    // user_with_auth is not set for the lower profiles (RSA 20248 and ECC P256)
```
ionut-arm, line 69: Nit:
```suggestion
    // user_with_auth is not set for the lower profiles (RSA 2048 and ECC P256)
```

ionut-arm (approved):
Thank you for the fix, much needed that someone with more XP on this refreshed it :)

ionut-arm:
Do we need to backport this to 7.x.y btw?

PR #557, opened 2024-11-30
Fixes API breaking changes.

Superhepper:
This is an attempt to avoid breaking the API in major version 7.

### DIFF
--- a/tss-esapi/src/abstraction/ak.rs
+++ b/tss-esapi/src/abstraction/ak.rs
@@ -8,8 +8,8 @@ use crate::{
     handles::{AuthHandle, KeyHandle, SessionHandle},
     interface_types::{
         algorithm::{
-            EccSchemeAlgorithm, HashingAlgorithm, PublicAlgorithm, RsaSchemeAlgorithm,
-            SignatureSchemeAlgorithm,
+            AsymmetricAlgorithm, EccSchemeAlgorithm, HashingAlgorithm, PublicAlgorithm,
+            RsaSchemeAlgorithm, SignatureSchemeAlgorithm,
         },
         session_handles::PolicySession,
     },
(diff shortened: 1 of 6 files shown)


ionut-arm (approved):
Looks good!

Superhepper:
@THS-on Did you have the time to check if changes would have any effect on the RSA 2048 keys?

tss-esapi/src/abstraction/ak.rs
```diff
@@ -228,8 +230,36 @@ pub fn load_ak(
     Ok(key_handle)
 }
 
-/// This creates an Attestation Key in the Endorsement hierarchy
+/// This creates an Attestation Key in the Endorsement hierarchy.
 pub fn create_ak<IKC: IntoKeyCustomization>(
```
wiktor-k, line 234: I think it'd be good to mark one of them as [`#[deprecated]`](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute)...

wiktor-k (dismissed):

tss-esapi/src/abstraction/ak.rs
Superhepper: I was thinking about that. The problem is that the function will not be removed but it will have its arguments changed in the next major version.

So xx will have its arguments changed.
and xx_2 will be removed. So xx_2 is the function that we want people to use right now. But it will be removed so marking it deprecated here could cause a lot of confusion.

wiktor-k: Good point, so let's not do that. But a doc-comment WARNING note or something may still be a good idea? :thinking:

Superhepper: That is an excellent idea. I need to rephrase the comments any way because I could barely understand them my self when I tried to look at the code in the browser.

Superhepper: I think I have managed to fix it now. Please have a last look at the comments so I have not made some embarrassing mistakes @wiktor-k @ionut-arm

wiktor-k: I think it looks really nice :ok_hand:

wiktor-k (approved):

ionut-arm (approved):
LGTM, thanks!!

PR #563, opened 2025-01-29
support for pure-rust make credentials

### LINKED ISSUE #160: Calling MakeCredential without a TPM
In tpm2-tools, `tpm2_makecredential` [can be called without a TPM](https://github.com/tpm2-software/tpm2-tools/blob/master/man/tpm2_makecredential.1.md).
Is there any way to call `context.makecredential(...)` without initializing a TCTI?

baloo:
This brings support for a pure rust implementation of make credentials which will not involve the TPM or `tpm2-tss`.

Fixes #160

### DIFF
--- a/tss-esapi/src/utils/credential.rs
+++ b/tss-esapi/src/utils/credential.rs
@@ -1,0 +1,430 @@
+// Copyright 2025 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+
+use core::{
+    fmt,
+    ops::{Add, Mul},
+};
+
+use cfb_mode::cipher::BlockCipherEncrypt;
+use digest::{
+    Digest, FixedOutputReset, Key, KeyInit, Mac, OutputSizeUser,
+    array::ArraySize,
+    common::{BlockSizeUser, Iv, KeyIvInit, KeySizeUser},
+    consts::{B1, U8},
+    typenum::{
+        Unsigned,
+        operator_aliases::{Add1, Sum},
+    },
+};
+use ecdsa::elliptic_curve::{
+    AffinePoint, Curve, CurveArithmetic, FieldBytesSize, PublicKey,
+    sec1::{FromSec1Point, ModulusSize, ToSec1Point},
+};
+use hmac::SimpleHmac;
+use log::error;
+use rand::rng;
+use zeroize::Zeroizing;
+
+#[cfg(feature = "rsa")]
+use rsa::RsaPublicKey;
+
+use crate::{
+    error::{Error, Result, WrapperErrorKind},
+    structures::{EncryptedSecret, IdObject, Name},
+    utils::{TpmHmac, kdf, secret_sharing},
+};
+
+/// Test if a key is considered weak according to TCG.
+///
+/// TCG will require weak keys to be re-generated,  
+/// See:
+/// ```text
+/// 11.4.10.4 Rejection of weak keys
+/// https://trustedcomputinggroup.org/wp-content/uploads/TPM-2.0-1.83-Part-1-Architecture.pdf#page=82
+/// The Key was considered weak, and we should re-run the creation of the encrypted
+/// secret.
+/// ```
+pub trait TcgKeyTest: KeyInit {
+    fn tcg_weak_key_test(key: &Key<Self>) -> core::result::Result<(), WeakKeyError>;
+}
+
+/// The error type returned when a key is found to be weak.
+#[derive(Copy, Clone, Eq, PartialEq, Debug)]
+pub struct WeakKeyError;
+
+impl fmt::Display for WeakKeyError {
+    #[inline]
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str("WeakKey")
+    }
+}
+
+impl core::error::Error for WeakKeyError {}
+
+type WeakResult<T> = core::result::Result<T, WeakKeyError>;
+
+#[cfg(feature = "aes")]
+mod key_test_aes {
+    use cipher::{Key, KeySizeUser, typenum::Unsigned};
+    use elliptic_curve::subtle::{Choice, ConstantTimeGreater};
+
+    use super::{TcgKeyTest, WeakKeyError};
+
+    macro_rules! weak_key_test {
+        ($k: ty) => {
+            impl TcgKeyTest for $k {
+                fn tcg_weak_key_test(key: &Key<Self>) -> Result<(), WeakKeyError> {
+                    // Check if any bit of the upper half of the key is set
+                    //
+                    // This follows the in terpretation laid out in section `11.4.10.4 Reject of weak keys`
+                    // from the TPM specification:
+                    // ```
+                    // In the case of AES, at least one bit in the upper half of the key must be set
+                    // ```
+                    // See: https://trustedcomputinggroup.org/wp-content/uploads/TPM-2.0-1.83-Part-1-Architecture.pdf#page=82
+                    let mut weak = Choice::from(0);
+
+                    for v in &key[..(<<$k as KeySizeUser>::KeySize as Unsigned>::USIZE / 2)] {
+                        weak |= <_ as ConstantTimeGreater>::ct_gt(v, &0);
+                    }
+
+                    if weak.unwrap_u8() == 0 {
+                        Err(WeakKeyError)
+                    } else {
+                        Ok(())
+                    }
+                }
+            }
+        };
+    }
+
+    weak_key_test!(aes::Aes128);
+    weak_key_test!(aes::Aes192);
+    weak_key_test!(aes::Aes256);
+}
+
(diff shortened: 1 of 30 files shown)


tss-esapi/tests/integration_tests/abstraction_tests/credential_tests.rs
```diff
@@ -0,0 +1,121 @@
...
+    let (credential_blob, secret) = utils::make_credential_ecc(
+        PublicKey::<p256::NistP256>::try_from(&ek_pub).unwrap(),
+        &cred,
+        key_name,
+    );
+
+    println!("{credential_blob:?}");
+    println!("{secret:?}");
+
+    let (session_aastributes, session_attributes_mask) = SessionAttributesBuilder::new().build();
```
wiktor-k, line 67: `session_aastributes` -> `session_attributes`

tss-esapi/src/utils/credential.rs
```diff
@@ -0,0 +1,104 @@
...
+
+    let mut sensitive_data = {
+        let mut out = vec![];
+        out.extend_from_slice(&u16::try_from(secret.len()).unwrap().to_be_bytes()[..]);
+        out.extend_from_slice(secret);
+        out
+    };
+
+    let sym_key = kdf::kdfa::<Sha256, kdf::Storage, aes::Aes128>(&seed, key_name.value(), &[]);
+    println!("----");
```
wiktor-k, line 79: This can be removed.

wiktor-k (commented):
Looks really cool. Of course error handling and other niceties are missing but that's not the point. Happy to see tests, I've skimmed them.

Thanks! :+1:

tss-esapi/tests/integration_tests/abstraction_tests/credential_tests.rs
baloo: This one was a copy paste from the other credentials tests. I wasn't sure if there was a special meaning or it was a typo.

baloo: https://github.com/parallaxsecond/rust-tss-esapi/pull/564

tss-esapi/tests/integration_tests/abstraction_tests/credential_tests.rs
```diff
@@ -0,0 +1,121 @@
...
+        ak_res.out_private,
+        ak_res.out_public,
+    )
+    .unwrap();
+
+    let (_, key_name, _) = context.read_public(ak_ecc).unwrap();
+    let cred = vec![1, 2, 3, 4, 5];
+    let expected = Digest::try_from(vec![1, 2, 3, 4, 5]).unwrap();
+
+    let (credential_blob, secret) = utils::make_credential_ecc::<_, sha2::Sha256, aes::Aes128>(
```
baloo, line 58: `sha2::Sha256, aes::Aes128` here codes for `EKHash` and `EkCipher`.
Those should be read from the template of the EK ideally.

Although in realily, the template would have been dropped already and we're only working with a PEM encoded public key, and there should be some kind of default value.
https://github.com/tpm2-software/tpm2-tools/blob/master/tools/tpm2_makecredential.c#L340

baloo:
I've finished support for both RSA and ECC, and there is now error management.
There isn't much else I can do.

Cargo.toml
```diff
@@ -1,3 +1,23 @@
...
+
+# https://github.com/RustCrypto/RSA/pull/467
+rsa = { git = "https://github.com/baloo/RSA.git", branch = "baloo/oaep/non-string-label" }
+
+# https://github.com/RustCrypto/traits/issues/1738
+# https://github.com/RustCrypto/traits/pull/1742
+# Pending release of crypto-common 0.2.0-rc.2
+# Pending release of digest 0.11.0-rc.0
+crypto-common = { git = "https://github.com/RustCrypto/traits.git" }
+digest = { git = "https://github.com/RustCrypto/traits.git" }
```
ionut-arm, line 23: Are we waiting for these to be released on crates.io?

tss-esapi/src/utils/mod.rs
```diff
@@ -23,6 +23,11 @@ use crate::{Context, Error, Result, WrapperErrorKind};
 use std::convert::TryFrom;
 use zeroize::Zeroize;
 
+mod credential;
+pub mod kdf;
```
ionut-arm, line 27: Since these make use of the Rust Crypto dependencies, would it make sense to gate them on that feature too?

tss-esapi/src/utils/credential.rs
```diff
@@ -0,0 +1,287 @@
...
+    <<EkHash as EagerHash>::Core as OutputSizeUser>::OutputSize: ArraySize + Mul<U8>,
+    <<<EkHash as EagerHash>::Core as OutputSizeUser>::OutputSize as Mul<U8>>::Output: Unsigned,
+
+    EkCipher: KeySizeUser + BlockCipherEncrypt + KeyInit,
+    <EkCipher as KeySizeUser>::KeySize: Mul<U8>,
+    <<EkCipher as KeySizeUser>::KeySize as Mul<U8>>::Output: ArraySize,
+{
+    // Prepare the sensitive data
+    // this will be then encrypted using AES-CFB (size of the symmetric key depends on the EK).
+    let mut sensitive_data = {
```
ionut-arm, line 230: Should this be zeroized once we're done with it?

tss-esapi/tests/integration_tests/abstraction_tests/credential_tests.rs
ionut-arm: Is your question about this test in particular, or about the interface of `make_credential_ecc`, and whether we can deduce the type params from the inputs?

I think generally you should be able to deduce the hash and the cipher for the EK if you know the nature of the public key, for example by doing the reverse of the mapping done here: https://github.com/parallaxsecond/rust-tss-esapi/blob/main/tss-esapi/src/abstraction/ek.rs

ionut-arm (commented):
Only had a brief look, will come back for more :)

Cargo.toml
baloo: Yeah, I'm relying on a bunch of pre-releases crates here. The kbkdf we just finished and released the first version of, but that relies on a bunch of pre-release or rc crates.

I've requested a new feature in the trait system for like detecting weak keys and be in compliance with the spec, probably safe to ignore that for now (and not merge this PR) until we have stable release.

The last huge pain point in the release cycle is a major rewrite of RSA, but I expect it to merge soon, and for the whole ecosystem to bump up.

tss-esapi/tests/integration_tests/abstraction_tests/credential_tests.rs
baloo: No that was an API question.

I know how to get the parameters from a Public, but I don't expect the public or its template to always available.

tss-esapi/src/utils/credential.rs
baloo: it's rewritten in-place

baloo: added a comment to make it clear

Firstyear (commented):
Outside of a skim-review, most of this looks like a really good change, especially going in the direction of rustcrypto.

The biggest issue with the rustcrypto ecosystem is the use of tiny crates, so finding a way to keep all the versions in lock-step will be a thing to consider in future.

baloo:
I'm fairly involved in RustCrypto, would happily handle the versions.

tss-esapi/src/utils/credential.rs
```diff
@@ -0,0 +1,287 @@
...
+struct TpmHmac<H>(PhantomData<H>);
+
+impl<H> KeySizeUser for TpmHmac<H>
+where
+    H: OutputSizeUser,
+{
+    type KeySize = H::OutputSize;
+}
+
+pub fn make_credential_ecc<C, EkHash, EkCipher>(
```
Firstyear, line 58: Can we get some documentation about the input parameters to this function please?

Firstyear (changes requested):

Cargo.toml
```diff
@@ -1,3 +1,29 @@
 [workspace]
 resolver = "2"
 members = ["tss-esapi", "tss-esapi-sys"]
+
+[patch.crates-io]
```
wiktor-k, line 5: I wonder what's the state of the RustCrypto ecosystem now... do we still need this patch section? :thinking:

Firstyear: Do we have a schedule for when they might release and update here? It'd be amazing to have this merged.

baloo: Next couple of months I would say. I think all the new things we wanted to have in crypto-bigint are in now.

tss-esapi/src/utils/credential.rs
baloo: Actually no, you're right, if an error occurs then it's not zeroized anymore.

PR #569, opened 2025-03-31
integration_tests: don't swallow error while creating ubuntu docker image

brandsimon:
When there is an error fetching the script and the result is empty, bash does not complain.
The result is a broken image.


brandsimon:
I think we use the `tpm2-software/ubuntu-20.04` image, because it is more difficult to install the tpm2 stuff then it is to install the rust stuff. (Also there are a lot of manual tss2 setup commands)

PR #570, opened 2025-03-31
Add checkquote function without using the tpm2

brandsimon:


baloo:
Not exactly a review but a couple of improvements on this PR: https://github.com/brandsimon/rust-tss-esapi/pull/6

baloo:
> I did not update any packages, but got the following [error](https://github.com/parallaxsecond/rust-tss-esapi/actions/runs/14178635892/job/39719429539?pr=570):
> 
> ```
> error: rustc 1.78.0 is not supported by the following package:
>   base64ct@1.7.3 requires rustc 1.81
> ```
> 
> Is this expected?

We released `base64ct 1.7` bumping MSRV to 1.81 a couple of weeks ago.

The [best practice](https://doc.rust-lang.org/cargo/faq.html#why-have-cargolock-in-version-control) is to commit the `Cargo.lock`, but this is not done here. Maybe a decision to revisit?

baloo:
> I did not update any packages, but got the following [error](https://github.com/parallaxsecond/rust-tss-esapi/actions/runs/14178635892/job/39719429539?pr=570):
> 
> ```
> error: rustc 1.78.0 is not supported by the following package:
>   base64ct@1.7.3 requires rustc 1.81
> ```
> 
> Is this expected?

https://github.com/parallaxsecond/rust-tss-esapi/pull/572
Something forgotten in #537

baloo:
Here is a followup to fix the rsa pkcs1v15 signature check: https://github.com/brandsimon/rust-tss-esapi/pull/7

tss-esapi/src/utils/mod.rs
```diff
@@ -23,6 +23,17 @@ use crate::{Context, Error, Result, WrapperErrorKind};
...
 
+#[cfg(all(
+    any(feature = "p256", feature = "rsa",),
+    any(feature = "sha1", feature = "sha2",)
+))]
+mod quote;
+#[cfg(all(
+    any(feature = "p256", feature = "rsa",),
+    any(feature = "sha1", feature = "sha2",)
+))]
```
baloo, line 34:
```suggestion
mod quote;
```

Those cfg are not required anymore

tss-esapi/src/utils/quote.rs
```diff
@@ -0,0 +1,393 @@
...
+        return Ok(false);
+    };
+
+    let verifying_key = VerifyingKey::from(public);
+
+    match hashing_algorithm {
+        #[cfg(feature = "sha1")]
+        HashingAlgorithm::Sha1 => {
+            let hash = sha1::Sha1::digest(&message);
+            Ok(verifying_key.verify_prehash(&hash, &signature).is_ok())
```
baloo, line 57: This one, I'm actually surprised it works. I was convinced the hash size had to line up with the size of the curve.

tss-esapi/src/utils/quote.rs
```diff
@@ -0,0 +1,393 @@
...
+            //#[cfg(feature = "p192")]
+            //impl_check_ecdsa!(p192::NistP192);
+            #[cfg(feature = "p224")]
+            impl_check_ecdsa!(p224::NistP224);
+            #[cfg(feature = "p256")]
+            impl_check_ecdsa!(p256::NistP256);
+            #[cfg(feature = "p384")]
+            impl_check_ecdsa!(p384::NistP384);
+            //#[cfg(feature = "p521")]
+            //impl_check_ecdsa!(p521::NistP521);
```
baloo, line 347: This one we can bring once we upgrade to p521 v0.14

tss-esapi/src/utils/quote.rs
```diff
@@ -0,0 +1,393 @@
...
+            //#[cfg(feature = "sm2")]
+            //impl_check_ecdsa!(sm2::Sm2);
+        }
+        #[cfg(feature = "rsa")]
+        (Public::Rsa { .. }, sig @ Signature::RsaSsa(pkcs_sig)) => {
+            let Ok(sig) = pkcs1v15::Signature::try_from(sig.clone()) else {
```
baloo, line 353: note to self: change the TryFrom in `structures::Signature` to accept a ref, we don't need the owned structure.

tss-esapi/src/utils/quote.rs
baloo: https://datatracker.ietf.org/doc/html/rfc6979#section-2.3.2

```
   1.  The sequence is first truncated or expanded to length qlen:

       *  if qlen < blen, then the qlen leftmost bits are kept, and
          subsequent bits are discarded;

       *  otherwise, qlen-blen bits (of value zero) are added to the
          left of the sequence (i.e., before the input bits in the
          sequence order)
          
```

TIL. I guess you don't get to sign a sha1 with P521, but this should be taken care of by the verifying library, nothing we should care about here.

baloo:
You can rebase on main to fix the two tests.

tss-esapi/src/utils/quote.rs
baloo: https://github.com/parallaxsecond/rust-tss-esapi/pull/573

brandsimon:
@baloo Thank you very very much for your work.
The verification functions are a lot cleaner now!

I did some changes so `cargo fmt` and `cargo clippy` succeed. From my side, this is ready to merge.

tss-esapi/src/utils/mod.rs
brandsimon: The standard configuration has no sha1 and sha2 feature enabled.
Without both, the compiler complains about the match, which does not match anything but the wildcard.
Without rsa and without elliptic curve, the compiler also has some complaints.

baloo: I think I would rather take a couple `#[allow(unused)]` over those combination of cfg, but this is a matter of taste I guess.

<details>

```
diff --git a/tss-esapi/Cargo.toml b/tss-esapi/Cargo.toml
index 05375b055a..6db91d09c3 100644
--- a/tss-esapi/Cargo.toml
+++ b/tss-esapi/Cargo.toml
@@ -49,7 +49,7 @@
 sha3 = { version = "0.10.8", optional = true }
 sm2 = { version = "0.13.3", optional = true }
 sm3 = { version = "0.4.2", optional = true }
-digest = { version = "0.10.7", optional = true }
+digest = { version = "0.10.7", optional = true, features = ["alloc"] }
 signature = { version = "2.2.0", features = ["std"], optional = true}
 cfg-if = "1.0.0"
 strum = { version = "0.26.3", optional = true }
diff --git a/tss-esapi/src/utils/mod.rs b/tss-esapi/src/utils/mod.rs
index 6077182a6e..d4df9c5cae 100644
--- a/tss-esapi/src/utils/mod.rs
+++ b/tss-esapi/src/utils/mod.rs
@@ -23,15 +23,7 @@
 use std::convert::TryFrom;
 use zeroize::Zeroize;

-#[cfg(all(
-    any(feature = "p224", feature = "p256", feature = "p384", feature = "rsa"),
-    any(feature = "sha1", feature = "sha2",)
-))]
 mod quote;
-#[cfg(all(
-    any(feature = "p224", feature = "p256", feature = "p384", feature = "rsa"),
-    any(feature = "sha1", feature = "sha2",)
-))]
 pub use quote::checkquote;

 /// Create the [Public] structure for a restricted decryption key.
diff --git a/tss-esapi/src/utils/quote.rs b/tss-esapi/src/utils/quote.rs
index 4063b5ede3..388c3aa2b0 100644
--- a/tss-esapi/src/utils/quote.rs
+++ b/tss-esapi/src/utils/quote.rs
@@ -8,23 +8,22 @@
     structures::{Attest, AttestInfo, DigestList, PcrSelectionList, Public, QuoteInfo, Signature},
     traits::Marshall,
 };
-use digest::{Digest, DynDigest};
+#[allow(unused)]
+use digest::Digest;
+use digest::DynDigest;

-#[cfg(any(feature = "p224", feature = "p256", feature = "p384"))]
 use crate::{abstraction::public::AssociatedTpmCurve, structures::EccSignature};
-#[cfg(any(feature = "p224", feature = "p256", feature = "p384"))]
 use ecdsa::{
     hazmat::{DigestPrimitive, VerifyPrimitive},
     PrimeCurve, SignatureSize, VerifyingKey,
 };
-#[cfg(any(feature = "p224", feature = "p256", feature = "p384"))]
 use elliptic_curve::{
     generic_array::ArrayLength,
     point::AffinePoint,
     sec1::{FromEncodedPoint, ModulusSize, ToEncodedPoint},
     CurveArithmetic, FieldBytesSize,
 };
-#[cfg(any(feature = "p224", feature = "p256", feature = "p384"))]
+#[allow(unused)]
 use signature::hazmat::PrehashVerifier;

 #[cfg(feature = "rsa")]
@@ -32,7 +31,7 @@
 #[cfg(feature = "rsa")]
 use signature::Verifier;

-#[cfg(any(feature = "p224", feature = "p256", feature = "p384"))]
+#[allow(unused)]
 fn verify_ecdsa<C>(
     public: &Public,
     message: &[u8],
@@ -147,6 +146,7 @@
     }
 }

+#[allow(unused)]
 fn checkquote_pcr_digests(
     quote: &QuoteInfo,
     selections: &PcrSelectionList,
@@ -185,8 +185,8 @@
     if digest_pos != digests.len() {
         return Err(Error::WrapperError(WrapperErrorKind::WrongParamSize));
     }
-    let digest = hasher.finalize().to_vec();
-    if digest != quote.pcr_digest().as_bytes() {
+    let digest = hasher.finalize();
+    if digest.as_ref() != quote.pcr_digest().as_bytes() {
         return Ok(false);
     }
     Ok(true)
@@ -312,10 +312,11 @@
         }
     };

+    #[allow(unused)]
     let bytes = attest.marshall()?;

+    #[allow(unused)]
     let hash_alg = match (public, signature) {
-        #[cfg(any(feature = "p224", feature = "p256", feature = "p384"))]
         (Public::Ecc { parameters, .. }, _) => {
             let mut hash_alg = None;
             macro_rules! impl_check_ecdsa {
```

</details>

brandsimon: I totally understand that it looks ugly and I am totally up to a cleaner solution.
Allowing dead code and unused variables does not get rid of all compilation issues. E.g. if sha1 and sha2 are not enabled, the errors include unreachable statements:

```
    let mut hasher: Box<dyn DynDigest> = match hash_alg {
        #[cfg(feature = "sha1")]
        HashingAlgorithm::Sha1 => Box::new(sha1::Sha1::new()),
        #[cfg(feature = "sha2")]
        HashingAlgorithm::Sha256 => Box::new(sha2::Sha256::new()),
        #[cfg(feature = "sha2")]
        HashingAlgorithm::Sha384 => Box::new(sha2::Sha384::new()),
        #[cfg(feature = "sha2")]
        HashingAlgorithm::Sha512 => Box::new(sha2::Sha512::new()),
        _ => {
            return Err(Error::WrapperError(WrapperErrorKind::UnsupportedParam));
        }   
    };  
```

Results in
```
error: unreachable statement                                                                                                                                               
   --> tss-esapi/src/utils/quote.rs:175:5                                                                                                                                  
    |                                                                                                                                                                      
161 |       let mut hasher: Box<dyn DynDigest> = match hash_alg {                                                                                                          
    |  __________________________________________-                                                                                                                         
162 | |         #[cfg(feature = "sha1")]                                                                                                                                   
163 | |         HashingAlgorithm::Sha1 => Box::new(sha1::Sha1::new()),                                                                                                     
164 | |         #[cfg(feature = "sha2")]                                                                                                                                   
...   |                                                                                                                                                                    
173 | |     };                                                                                                                                                             
    | |_____- any code following this `match` expression is unreachable, as all arms diverge                                                                               
174 |                                                                                                                                                                      
175 | /     for selection in selections.get_selections() {                                                                                                                 
176 | |         let sel_count = selection.selected().len();                                                                                                                
177 | |         if digest_pos + sel_count > digests.len() {                                                                                                                
178 | |             return Err(Error::WrapperError(WrapperErrorKind::WrongParamSiz...                                                                                      
...   |                                                                                                                                                                    
184 | |     }                                                                                                                                                              
    | |_____^ unreachable statement
```

brandsimon: Also if the result of this discussion is to remove (some) ifdef and allow dead code and unused variables, ci-tests should be excluded, so we still catch unclean code. (I still think the current version is cleaner in total)

Maybe we should also add test compilation with different features enabled during ci-test.

baloo: This unreachable statement we can probably sort out by moving this to a function, but I get what you're saying.

baloo: https://github.com/brandsimon/rust-tss-esapi/pull/8

Superhepper: All the features should be tested automatically. I made an effort a while ago to make it happen by having the crate itself as `dev-dependency` with all features we enabled. I forgot to add some documentation though to specify that all new options needs to be added to the list.

tss-esapi/src/utils/quote.rs
```diff
@@ -0,0 +1,390 @@
...
+        }
+    }
+    if digest_pos != digests.len() {
+        return Err(Error::WrapperError(WrapperErrorKind::WrongParamSize));
+    }
+    let digest = hasher.finalize().to_vec();
+    if digest != quote.pcr_digest().as_bytes() {
+        return Ok(false);
+    }
+    Ok(true)
```
wiktor-k, line 192: I think this can be simplified as:

```suggestion
    Ok(digest == quote.pcr_digest().as_bytes())
```

I wonder if we need a constant-time equality there? I think RustCrypto had something for this :thinking: 

I think this can be optimized a bit further by using `as_ref` instead of `to_vec` in the result of `hasher.finalize()`. See [the example on rust playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=b59ea47f77a208f92576337f49ce25a1).

(IIRC `pcr_digest()` also has `as_ref()` implementation :thinking: )

tss-esapi/src/utils/quote.rs
```diff
@@ -0,0 +1,390 @@
...
+///
+/// # Arguments
+/// * `attest` - Attestation data containing a quote
+/// * `signature` - Signature for the attestation data
+/// * `public` - TPM2 public struct which contains the public key for verification
+/// * `pcr_data` - Optional pcr values to verify
+/// * `qualifying_data` - qualifying_data to verify
+///
+/// # Returns
+/// The command returns a boolean
```
wiktor-k, line 205: I think this needs a bit more elaborate description:

```suggestion
/// The command returns `true` if the quote is valid or `false` otherwise.
```

tss-esapi/src/utils/quote.rs
```diff
@@ -0,0 +1,390 @@
...
+}
+
+/// Verify a quote
+///
+/// # Arguments
+/// * `attest` - Attestation data containing a quote
+/// * `signature` - Signature for the attestation data
+/// * `public` - TPM2 public struct which contains the public key for verification
+/// * `pcr_data` - Optional pcr values to verify
+/// * `qualifying_data` - qualifying_data to verify
```
wiktor-k, line 202:
```suggestion
/// * `qualifying_data` - qualifying data to verify
```

tss-esapi/src/utils/quote.rs
```diff
@@ -0,0 +1,390 @@
...
+                        };
+                        if !verify_ecdsa::<$curve>(&public, &bytes, &sig, sig.hashing_algorithm())?
+                        {
+                            return Ok(false);
+                        }
+                        hash_alg = Some(sig.hashing_algorithm());
+                    }
+                };
+            }
+            //#[cfg(feature = "p192")]
```
wiktor-k, line 335: If this doesn't work I think it can be removed.

tss-esapi/src/utils/quote.rs
```diff
@@ -0,0 +1,390 @@
...
+/// ```
+pub fn checkquote(
+    attest: &Attest,
+    signature: &Signature,
+    public: &Public,
+    pcr_data: &Option<(PcrSelectionList, DigestList)>,
+    qualifying_data: &Vec<u8>,
+) -> Result<bool> {
+    let quote = match attest.attested() {
+        AttestInfo::Quote { info } => info,
```
wiktor-k, line 309: Do we expect this list to grow? If not I guess `if let` would be a bit more concise :thinking:

wiktor-k (commented):
I think this looks very nice :ok_hand:  just a couple of nits/questions.

tss-esapi/src/utils/quote.rs
baloo: I don't think there is a reason to have that be a constant-time check. This would not leak a private key or anything.

tss-esapi/src/utils/mod.rs
brandsimon: @baloo 
Your branch fails to compile with default features.
```
error: unreachable statement                                                                                                                                               
   --> tss-esapi/src/utils/quote.rs:398:5                                                                                                                                  
    |                                                                                                                                                                      
334 |       let hash_alg = match (public, signature) {                                                                                                                     
    |  ____________________-                                                                                                                                               
335 | |         #[cfg(any(feature = "p224", feature = "p256", feature = "p384"))]                                                                                          
336 | |         (Public::Ecc { parameters, .. }, _) => {                                                                                                                   
337 | |             let mut hash_alg = None;                                                                                                                               
...   |                                                                                                                                                                    
396 | |     };                                                                                                                                                             
    | |_____- any code following this `match` expression is unreachable, as all arms diverge                                                                               
397 |                                                                                                                                                                      
398 | /     if qualifying_data != attest.extra_data().as_bytes() {                                                                                                         
399 | |         return Ok(false);                                                                                                                                          
400 | |     }                                                                                                                                                              
    | |_____^ unreachable statement
```

brandsimon: If no one has hard feelings about this, I'd prefer to keep the cfgs for now.

brandsimon: @Superhepper Do you mean all features enabled or for every feature combination? (I meant all combinations, I created a separate issue for this

brandsimon:
@wiktor-k 
Thank you, all done!
The clippy failures are also on main and I think because of a new version of clippy.

brandsimon:
I fixed the clippy issues in https://github.com/parallaxsecond/rust-tss-esapi/pull/575
Once merged, I rebase onto main.

Superhepper (commented):
It is my personal preference to try to avoid names like common, utils, e.t.c because they tend to end up becoming a garbage can full of "nice to have things". And it has been one of my personal goals to try to get rid of the utils module once and for all.

With that said. I would much rather have this put under `abstraction`. And if you want to be really specific about it you could even create a sub module in there called `no_tpm`.

wiktor-k:
> It is my personal preference to try to avoid names like common, utils, e.t.c because they tend to end up becoming a garbage can full of "nice to have things". And it has been one of my personal goals to try to get rid of the utils module once and for all.

FWIW I agree with this perspective and also don't like utils, helpers, etc.

wiktor-k (approved):
LGTM. FWIW you can fix the sign-offs with `git rebase --signoff main && git push --force-with-lease`... or... well... since it's just last commit I guess `git commit --amend --signoff --no-edit` would work too :)

brandsimon:
@wiktor-k 
Thank you, signed-off the last commit

Superhepper (approved):

tss-esapi/src/abstraction/no_tpm/quote.rs
```diff
@@ -0,0 +1,379 @@
...
+        return Ok(false);
+    };
+
+    let verifying_key = VerifyingKey::from(public);
+
+    match hashing_algorithm {
+        #[cfg(feature = "sha1")]
+        HashingAlgorithm::Sha1 => {
+            let hash = sha1::Sha1::digest(message);
+            Ok(verifying_key.verify_prehash(&hash, &signature).is_ok())
```
baloo, line 61: The use of `PrehashVerifier` should be swapped with `DigestVerifier` just so we stay off the hazmat.

https://github.com/RustCrypto/elliptic-curves/pull/1146

PR #572, opened 2025-04-01
tests: update `Cargo.lock`

baloo:
A fresh `Cargo.lock`, base64ct is expressively downgraded to 1.6.0 to preserve MSRV.

That should have been done in https://github.com/parallaxsecond/rust-tss-esapi/pull/537, but I didn't realized that was required then.


wiktor-k (approved):
Looks innocent :+1:

Superhepper (approved):

PR #575, opened 2025-04-10
Fix clippy issues

brandsimon:


wiktor-k (approved):
Too bad that the manual indent is now broken but there's no way to discuss this fine details with clippy I guess :man_shrugging:

Superhepper (approved):

PR #578, opened 2025-05-30
feat: ignore fake/nonsense prerelease info from tss version

mcr:
when building against a libtss from source code, the git tree is sometimes dirty, and so libtss puts that info into the VERSION file.
But, it's not exactly a correct prerelease SemVer string.... so just blank that part before comparing to required version.

### DIFF
--- a/tss-esapi/build.rs
+++ b/tss-esapi/build.rs
@@ -1,6 +1,6 @@
 // Copyright 2021 Contributors to the Parsec project.
 // SPDX-License-Identifier: Apache-2.0
-use semver::{Version, VersionReq};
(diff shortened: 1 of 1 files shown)


Superhepper (approved):
Thank you for the PR.
Yes this seems reasonable. I wonder if it would be possible to issue a warning or an info message that the version seems to indicate a non officiell release.

wiktor-k:
Aaah, so without this fix the version parsing fails? I'd add that info to the commit message (so that it's clear it's a bug fix when using prerelease sources).

I'd be good to drop the commented out eprintln.

Thanks for your change!

PR #584, opened 2025-09-24
Add definition for NV_Extend

nathanrossi:
Add a implementation for NV_Extend, which is used with a NV Index defined with the NvIndexType::Extend to provide a PCR equivalent that uses NV memory. The hashing algorithm used is defined by the index name algorithm.

An example and test are included for this function, setting up the NV index with the required type and typical attributes and performing an extend of the index and confirming the resulting hash value.

### DIFF
--- a/tss-esapi/src/context/tpm_commands/non_volatile_storage.rs
+++ b/tss-esapi/src/context/tpm_commands/non_volatile_storage.rs
@@ -6,7 +6,7 @@ use crate::{
     interface_types::reserved_handles::{NvAuth, Provision},
     structures::{Auth, MaxNvBuffer, Name, NvPublic},
     tss2_esys::{
-        Esys_NV_DefineSpace, Esys_NV_Increment, Esys_NV_Read, Esys_NV_ReadPublic,
+        Esys_NV_DefineSpace, Esys_NV_Extend, Esys_NV_Increment, Esys_NV_Read, Esys_NV_ReadPublic,
         Esys_NV_UndefineSpace, Esys_NV_UndefineSpaceSpecial, Esys_NV_Write,
     },
     Context, Result, ReturnCode,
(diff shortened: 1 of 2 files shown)


Superhepper (dismissed):
Thank you for the PR. This looks good to me.

tss-esapi/src/context/tpm_commands/non_volatile_storage.rs
```diff
@@ -698,7 +698,123 @@ impl Context {
...
+    /// #     .expect("Received invalid handle");
+    /// # let (session_attributes, session_attributes_mask) = SessionAttributes::builder()
+    /// #     .with_decrypt(true)
+    /// #     .with_encrypt(true)
+    /// #     .build();
+    /// # context.tr_sess_set_attributes(session, session_attributes, session_attributes_mask)
+    /// #     .expect("Failed to set attributes on session");
+    /// # context.set_sessions((Some(session), None, None));
+    /// #
+    /// # let nv_index = NvIndexTpmHandle::new(0x01500028)
```
Firstyear, line 753: Where does this magic number come from?

Firstyear (changes requested):
I think my major thought here is that while the example is really good, it has a few caveats.

First, examples are run during "cargo test". I don't think this will be a problem, but this does differ from the normal tests that exist in the project.

Second, a lot of the example is commented out - the issue with examples sometimes is that while it's great to show usage of the API, it's also important to show the supporting elements. For example I would uncomment everything from `let session` and below.

Finally as commented, you use a magic number and it's unclear where you derived that from, so it'll be great to explain the origin of that.

Superhepper:
> Second, a lot of the example is commented out - the issue with examples sometimes is that while it's great to show usage of the > API, it's also important to show the supporting elements. For example I would uncomment everything from let session and
> below.

Previously we have either just copied the existing integration test into the documentation examples and commented out all the stuff that is not directly related to the function call (such as but not limited to, TCTI config loading, Context creation and Session creation) in order to make the examples smaller and easier to read. Some of the APIs are quite complex and requires a lot of things before they can be used which would create examples that would be crazy long. I can agree though that it can be quite deceiving for some one who looks at the documentation and thinks that it looks simple (when in my opinion nothing about the TPM is simple). We should probably add to the library documentation that code in the examples is actually not all the code that is needed in order to actually call the API successfully and that the full code can be found in the integration tests.

I wouldn't mind getting the test duplicated into the integration tests and if it could be done with some variation that would be even better!

> Finally as commented, you use a magic number and it's unclear where you derived that from, so it'll be great to explain the 
> origin of that.

Yeah, guess they can seem magical but it is actually exactly as described by @nathanrossi. Every NV handle used in the tests uses a unique value in the NV handle range. So the way to get a new handle is actually to search for the highest number in the code base and a +1. This is done this way in order to make it easier to determine where a problem regarding NV handles in test originated from.


We should probably also add some kind of Repo Development Guide where this things can be described in order to make it easier for new people to contribute.

tss-esapi/src/context/tpm_commands/non_volatile_storage.rs
Firstyear: More that if I was reading this, I would think "NvIndexTpmHandle::new" wants an integer but I would not know where to pull that value from. So at least documenting "the value should be an int from X to Y" or "you can get the address by doing Z" would help, because else this is just a magic number.

Superhepper:
We have specified which specifications that are going to be referenced in the documentation. But in the documentation for the context methods we seldom refer to them. Maybe just add a "for valid ranges for NV handles please see the specifcation".

Firstyear:
Sounds reasonable to me.

ionut-arm:
> Maybe just add a "for valid ranges for NV handles please see the specifcation".

And ideally mention which specification / section (e.g., Commands spec, section x.y.z). Leaving it as "see the spec" works, but it's a bit more helpful like this for anyone who's not familiar with the beast that is the TPM2 spec.

nathanrossi:
I have updated this pull request to include more details about the nv_index_handle, providing some references to useful specifications for both general NV index use and specific use that is relevant for nv_extend.

The example is kept, with the session setup commented/hidden, as I was not sure if there was agreement on the verbosity it introduced. However I have split the example and created a separate integration test that includes the example and performs the assertions, as well as including a second usage test which has differing attribute parameters and authorization.

Superhepper (approved):

PR #591, opened 2025-10-28
Windows now needs msbuild dep for generate-bindings

NuSkooler:
Need the dep as it's referenced in build.rs now.


Superhepper:
You need to sign the commit. That is why the DCO does not pass.

NuSkooler:
> You need to sign the commit. That is why the DCO does not pass.

Singing updated!

PR #595, opened 2025-11-08
Updates the documentation regarding the bundled feature.

Superhepper:


wiktor-k (dismissed):
I can't test it right now on Windows but it looks good on the surface.

Firstyear:
Yep, seems simple enough to me :)

PR #616, opened 2026-02-04
chore(deps): Bump rustcrypto dependencies to new releases

baloo:


ionut-arm (dismissed):
LGTM so far

Superhepper:
Nice to see this coming along as well. Only waiting for a few more dependencies to be properly released.

ionut-arm:
Looking forward to this moving along! What crates are we waiting for to be released?

baloo:
> Looking forward to this moving along! What crates are we waiting for to be released?

Last one is rsa. I don't know when we can expect a release.

Issue #633, opened 2026-04-22
bindgen dependency needs bump to >= 0.72.1 for LLVM/Clang 22 compatibility


sarroutbi:
### Summary

`tss-esapi-sys` currently depends on `bindgen ^0.66.1`. LLVM 22 introduces a breaking change in how `TagType` declarations are represented in the AST, which causes bindgen versions prior to 0.72.1 to generate incorrect bindings
(incomplete structs with missing fields, replaced by a 1-byte `_address` placeholder).

This results in a build failure (FTBFS) for any project using `tss-esapi` with the `generate-bindings` feature when building against Clang 22.

### Upstream Reference

- rust-lang/rust-bindgen#3264
- Fix landed in bindgen 0.72.1 via rust-lang/rust-bindgen#3278

### Request

Would it be possible to release a 7.x patch (e.g. 7.6.1 or 7.7.0) of `tss-esapi` / `tss-esapi-sys` that bumps the `bindgen` dependency to `>= 0.72.1`?

I see that `tss-esapi-sys 0.6.0-alpha.2` (used by `tss-esapi 8.0.0-alpha.2`) already has `bindgen ^0.72.0`, but upgrading to the 8.x pre-release is not feasible for many consumers due to API breakage and downstream dependency
constraints (e.g. `tpm2-policy ^0.6.0` requires `tss-esapi ^7.0.0`).

### Impact

Distributions shipping LLVM 22 (e.g. Fedora) will be unable to build packages that depend on `tss-esapi` with `generate-bindings` enabled unless this is addressed. This affects at least:

- [clevis-pin-tpm2](https://github.com/fedora-iot/clevis-pin-tpm2)

Superhepper:
I mean it could be possible as long as that version is compatible with the MSRV and edition of the 7-branch.

anta5010:
tss-esapi 7.7.0 with a fix included (uses tss-esapi-sys 0.6.0) has been released

PR #634, opened 2026-04-23
Bump bindgen to 0.72.1 (fix LLVM 22 compatibility)

sarroutbi:
Resolves: #633

# Summary

**Bump `bindgen` dependency from `^0.66.1` to `^0.72.1` in `tss-esapi-sys` to fix build failures with LLVM/Clang 22.**:

LLVM 22 introduces a breaking change in how `TagType` declarations are represented in the AST, causing bindgen < 0.72.1 to generate incorrect bindings (incomplete structs with missing fields, replaced by a 1-byte `_address` placeholder). This results in FTBFS for any project using `tss-esapi-sys` with the `generate-bindings` feature when building against Clang 22.

### Upstream references

- [Missing fields in nested class with LLVM 22](https://github.com/rust-lang/rust-bindgen/issues/3264)
- [Fix landed in bindgen 0.72.1](https://github.com/rust-lang/rust-bindgen/pull/3278)

### MSRV compatibility

bindgen 0.72.1 has MSRV 1.70.0, while the 7.x branch has MSRV 1.66.0. Since `bindgen` is an **optional build dependency** (only pulled in with the `generate-bindings` feature), users relying on pre-generated bindings are completely unaffected and can continue using Rust 1.66.0. Anyone hitting the LLVM 22 bug necessarily has `generate-bindings` enabled on a modern distro that ships Rust well above 1.70.0.

### Impact

Distributions shipping LLVM 22 (e.g. Fedora) are unable to build packages that depend on `tss-esapi` with `generate-bindings` enabled. This affects at least `clevis-pin-tpm2`.


tss-esapi-sys/Cargo.toml
```diff
@@ -14,7 +14,7 @@ links = "tss2-esys"
 rust-version = "1.66.0"
 
 [build-dependencies]
-bindgen = { version = "0.66.1", optional = true }
+bindgen = { version = "0.72.1", optional = true }
```
anta5010, line 17: bindgen 0.72.1 MSRV is 1.70.
I think you should update the "rust-version" above

tss-esapi/tests/lint-checks.sh
```diff
@@ -20,4 +20,9 @@ fi
 ##################
 # Execute clippy #
 ##################
-cargo clippy --all-targets --all-features -- -D clippy::all -D clippy::cargo
\ No newline at end of file
+if [[ ! -z ${RUST_TOOLCHAIN_VERSION:+x} ]]; then
```
anta5010: why don't we bump clippy MSRV?

ionut-arm (dismissed):
Looks good to me, let's hope the CI agrees

wiktor-k (dismissed):
LGTM 👍

anta5010:
You missed one 1.66 in ci.yml line 35

anta5010:
I think you can set MSRV to 1.85 in all places (twice in ci.yaml) and in Cargo.toml.
We are also going to bump picky-asn1-* crates which have MSRV 1.85 set.

anta5010:
:( there are three places in ci.yaml.
We'd better define a variable.

anta5010 (dismissed):
LGTM. Thank you for run-ci-locally.sh

ionut-arm (dismissed):

tss-esapi-sys/Cargo.toml
```diff
@@ -11,10 +11,10 @@ license = "Apache-2.0"
 repository = "https://github.com/parallaxsecond/rust-tss-esapi"
 documentation = "https://docs.rs/crate/tss-esapi-sys"
 links = "tss2-esys"
-rust-version = "1.66.0"
+rust-version = "1.85.0"
```
Superhepper, line 14: Changing the MSRV or the edition has the potential of braking stuff for consumers of the crate even though there are no API changes. And I am little unsure of how to handle this from a versioning perspective.

tss-esapi/tests/run-ci-locally.sh
```diff
@@ -0,0 +1,342 @@
+#!/usr/bin/env bash
```
Superhepper: This is excellent!

tss-esapi-sys/Cargo.toml
sarroutbi: > Changing the MSRV or the edition has the potential of braking stuff for consumers of the crate even though there are no API changes. And I am little unsure of how to handle this from a versioning perspective.

Is it feasible to bring this change to 7.7.0 and keep 7.6.1 onward for the "stable" unbroken version?

anta5010: I think bumping MSRV is a necessary step even if it adds some inconvenience. 
Building tss-esapi v7.6.0 with resent version of rust would produce, for example, a warning:
``` 
warning: the following packages contain code that will be rejected by a future version of Rust: picky-asn1-x509 v0.12.0
```
I will raise a PR to fix it  as soon as this PR is merged. The picky-asn1-* crates were the warning is fixed require rust 1.85.

I'm not sure we can do any version except 7.7.0 (not before my PR reviewed and merged).
It's still possible to include tss-esapi with `version="=7.6.0" in cases where the new MSRV is an issue.
I saw crates with MSRV bumped in the same major version, so it's not an exception I think

tss-esapi/tests/run-ci-locally.sh
sarroutbi: > This is excellent!

Thanks. I removed it as it should go on its own PR: #636

tss-esapi-sys/Cargo.toml
ionut-arm: Indeed, I guess we have to weigh the inconvenience of breaking people who are on Rust <1.85, against keeping up with the distros and everyone else who's using the crate. We can't use 8.x.y for this crate either, since we've moved on with other stuff for 8. And asking people to move from 7 to 8 of this crate is a huge step.

As Anton says - for those who absolutely need an older compiler, they can pin the older version. Even the [Rust documentation](https://doc.rust-lang.org/cargo/reference/rust-version.html#setting-and-updating-rust-version) says this is ok for a minor increase:

> Note: [Changing rust-version](https://doc.rust-lang.org/cargo/reference/semver.html#env-new-rust) is assumed to be a minor incompatibility

sarroutbi: > I think bumping MSRV is a necessary step even if it adds some inconvenience. Building tss-esapi v7.6.0 with resent version of rust would produce, for example, a warning:
> 
> ```
> warning: the following packages contain code that will be rejected by a future version of Rust: picky-asn1-x509 v0.12.0
> ```
> 
> I will raise a PR to fix it as soon as this PR is merged. The picky-asn1-* crates were the warning is fixed require rust 1.85.
> 
> I'm not sure we can do any version except 7.7.0 (not before my PR reviewed and merged). It's still possible to include tss-esapi with `version="=7.6.0" in cases where the new MSRV is an issue. I saw crates with MSRV bumped in the same major version, so it's not an exception I think



> I think bumping MSRV is a necessary step even if it adds some inconvenience. Building tss-esapi v7.6.0 with resent version of rust would produce, for example, a warning:
> 
> ```
> warning: the following packages contain code that will be rejected by a future version of Rust: picky-asn1-x509 v0.12.0
> ```

Not sure what you mean with this. I would expect this to land in v7.7.0

> 
> I will raise a PR to fix it as soon as this PR is merged. The picky-asn1-* crates were the warning is fixed require rust 1.85.

Feel free to add it to this PR directly

> 
> I'm not sure we can do any version except 7.7.0 (not before my PR reviewed and merged). It's still possible to include tss-esapi with `version="=7.6.0" in cases where the new MSRV is an issue. I saw crates with MSRV bumped in the same major version, so it's not an exception I think

OK. Sorry, newbie in the project. It makes sense.

anta5010: > Not sure what you mean with this. I would expect this to land in v7.7.0

Yes, it will be included into 7.7.0
@ionut-arm is going to bump dependency crates and bump tss-esapi and tss-esapi-sys versions in separate PRs

ionut-arm (approved):

PR #638, opened 2026-05-01
Bump sys crate version on main after 0.6.0 release.

Superhepper:
After the release of tss-esapi-sys 0.6.0 it is
no longer a good idea to continue to release
0.6.0-alpha releases. So this commit bumps the
version of the tss-espi-sys crate to enable further development.


ionut-arm (approved):
Ah, shit. I had probably missed checking the version on that side.

wiktor-k (approved):

PR #639, opened 2026-05-01
Update tss-espi dependency versions.

Superhepper:
Updates the versions of bitfield, strum
and strum_macro dependencies.

There were also a build warning in the tests
that got fixed by not including TPM2_PT_FIXED
and TPM2_PT_VAR in the 'property_tag_tests.rs'
because those or not really part of the enum.


ionut-arm (commented):
Looks good, but is there a need to update the frozen Cargo.lock?

Superhepper:
> Looks good, but is there a need to update the frozen Cargo.lock?

Fixed that.

ionut-arm (approved):
Thanks!

wiktor-k (approved):
Thanks!

PR #653, opened 2026-06-10
Add an initializer for static binaries to use

indirect:
I'm working on a project that uses this crate, but that would ideally be built completely statically for reproducibility reasons. With a fully static binary, `dlopen` is an empty stub, which prevented this crate from working.

I'm extremely unsure if this is the best approach for the crate as a whole, but I can confirm this change does in fact solve my specific problem, making it possible to use the rest of the crate even when `dlopen` isn't available.

You can see it in practice (and try out using this code) over at https://github.com/lunal-dev/kettle.

### DIFF
--- a/tss-esapi/src/tcti_ldr.rs
+++ b/tss-esapi/src/tcti_ldr.rs
@@ -24,12 +24,34 @@ const TABRMD: &str = "tabrmd";
 const LIBTPMS: &str = "libtpms";
 const TBS: &str = "tbs";
 
+// libc malloc/free, declared inline to avoid a new dependency. Required for
+// Tss2_Tcti_Device_Init, which initialises a caller-
+// provided buffer in place, and on Drop we libc-free it ourselves.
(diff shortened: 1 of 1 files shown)


Superhepper:
I mean it is not bad as first draft. One thing to think of would to be to make the `unsafe` blocks as small as possible. From experience I can tell you that a lot of bad things can happen when you try to be clever inside of unsafe.

Another thing I wonder about is if it would be interesting in the bundled feature to have the option to link everything statically.

indirect:
I would love to have a feature that makes everything static! That would straight up solve my use-case entirely.

I will give this another pass soon to see if I can shrink the unsafes and get all the tests passing again now that I have rebased.

Superhepper:
Try to rebase this PR on main now. To see if the checks goes green.

I have also added this Issue #670 and I will have to see if I or some one else(you?) picks it up.

PR #658, opened 2026-07-11
HashAgile: expose the algorithm and the content

baloo:
This allows to verify an hmac signature emitted by a TPM.

### DIFF
--- a/tss-esapi/src/abstraction/hashing.rs
+++ b/tss-esapi/src/abstraction/hashing.rs
@@ -3,6 +3,12 @@
 
 use crate::interface_types::algorithm::HashingAlgorithm;
 
+#[cfg(feature = "rustcrypto")]
+use {
+    crate::{Error, Result, WrapperErrorKind, structures::HashAgile},
+    digest::{Output, OutputSizeUser},
+};
+
 /// Provides the value of the digest used in this crate for the digest.
 pub trait AssociatedHashingAlgorithm {
     /// Value of the digest when interacting with the TPM.
@@ -47,4 +53,27 @@ impl AssociatedHashingAlgorithm for sha3::Sha3_384 {
 #[cfg(feature = "sha3")]
 impl AssociatedHashingAlgorithm for sha3::Sha3_512 {
     const TPM_DIGEST: HashingAlgorithm = HashingAlgorithm::Sha3_512;
+}
+
+impl HashAgile {
+    #[cfg(feature = "rustcrypto")]
+    #[expect(
+        clippy::unnecessary_fallible_conversions,
+        reason = "GenericArray::From<&[T]> will panic if the payload is not the same length"
+    )]
+    /// Return the hash content of the [`HashAgile`]
+    ///
+    /// # Errors
+    ///
+    /// Return an error if the specified [`AssociatedHashingAlgorithm`] is not the
+    /// one in the [`HashAgile`].
+    pub fn to_output<H: AssociatedHashingAlgorithm + OutputSizeUser>(&self) -> Result<Output<H>> {
+        if self.algorithm == H::TPM_DIGEST {
+            <&Output<H>>::try_from(self.digest.as_bytes())
+                .map_err(|_| Error::local_error(WrapperErrorKind::WrongValueFromTpm))
+                .cloned()
+        } else {
+            Err(Error::local_error(WrapperErrorKind::InvalidParam))
+        }
+    }
 }
--- a/tss-esapi/src/structures/hash/agile.rs
+++ b/tss-esapi/src/structures/hash/agile.rs
@@ -8,8 +8,8 @@ use std::convert::{TryFrom, TryInto};
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct HashAgile {
-    algorithm: HashingAlgorithm,
-    digest: Digest,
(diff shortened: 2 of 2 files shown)


baloo:
missing clippy lints are fixed in https://github.com/parallaxsecond/rust-tss-esapi/pull/657, i'll rebase once that merges.

Superhepper:
Question: Is there any way to reverse the dependency, so that the abstraction stuff remains in the abstraction module, without making things very awkward?

baloo:
I've moved the implementation in `tss-esapi/src/abstraction/hashing.rs`, I had to massage the visibility of the fields a bit, but let me know if that's an acceptable alternative.

Superhepper:
Yeah, I think its fine. How would the `use` statements end up looking if I would want to use the functionality?
```rust
use tss_esapi::{structures::HashAgile, abstractions::*};
```
?

Or can it be narrowed down? I mean if this is the case I still think it is reasonable.

baloo:
You don't need any `use` imports.

```
    let tss_esapi::structures::Signature::Hmac(hmac_sig) = signature else {
        return Err(io::Error::other("expected HMAC signature").into());
    };

    let sig_slice = hmac_sig
        .to_output::<sha2::Sha256>()
        .map_err(|e| io::Error::other(format!("failed to convert HMAC signature: {e}")))?;
```

You will need to enable `abstractions`, and `rustcrypto` features though.

tss-esapi/src/abstraction/hashing.rs
```diff
@@ -48,3 +54,26 @@ impl AssociatedHashingAlgorithm for sha3::Sha3_384 {
 impl AssociatedHashingAlgorithm for sha3::Sha3_512 {
     const TPM_DIGEST: HashingAlgorithm = HashingAlgorithm::Sha3_512;
 }
+
+impl HashAgile {
+    #[cfg(feature = "rustcrypto")]
+    #[allow(
+        clippy::unnecessary_fallible_conversions,
+        reason = "GenericArray::From<&[T]> will panic if the payload is not the same length"
```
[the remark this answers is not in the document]
baloo: I mean I don't know, that was the previous version, @Superhepper an opinion?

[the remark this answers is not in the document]
Superhepper: No, it is wrong. The original was gated it with both "abstraction" and "rustcrypto". So not receiving it when `default-features = False` is BY DESIGN. So I would consider this a false positive.

Superhepper (dismissed):

tss-esapi/src/abstraction/hashing.rs
```diff
@@ -48,3 +54,26 @@ impl AssociatedHashingAlgorithm for sha3::Sha3_384 {
 impl AssociatedHashingAlgorithm for sha3::Sha3_512 {
     const TPM_DIGEST: HashingAlgorithm = HashingAlgorithm::Sha3_512;
 }
+
+impl HashAgile {
+    #[cfg(feature = "rustcrypto")]
+    #[allow(
```
wiktor-k: Have you considered using `expect` instead of `allow`?

This way if the lint stops being reported (e.g. when `::from` gets removed) this instruction can be removed...

wiktor-k (dismissed):
LGTM 👍 thanks!

ionut-arm (dismissed):

tss-esapi/src/abstraction/hashing.rs
baloo: You got it reversed.

What clippy is saying is that you should use `Output::from`because it's available. (`Output` is an alias on `generic_array::GenericArray`). Sadly, `generic_array::GenericArray::from` WILL panic if the slice doesn't have the correct size and one should really use `try_from` (this is what I'm trying to point at with the reason).

Clippy complains about using `try_from` when it sees `from` is available.

This goes away with https://github.com/parallaxsecond/rust-tss-esapi/pull/616 which migrates to `hybrid_array::Array` which doesn't have a `From` implementation, and where I'm removing the `clippy::allow` (Don't hunt for it in 616, it's not there yet. This is in a followup commit that will go in when I rebase on top of here, there is no HashAgile::to_output there yet.)

The alternative might be to use https://docs.rs/generic-array/latest/generic_array/struct.GenericArray.html#method.try_from_slice which wouldn't trigger clippy.

wiktor-k: I think there's a slight misunderstanding here. I was talking about the *attibute* called `expect` like so: https://doc.rust-lang.org/reference/attributes/diagnostics.html#r-attributes.diagnostics.lint.expect

IMHO it's generally better than `allow` as it triggers when the lint is no longer needed.

baloo: I've never seen clippy's expect! I confused your suggestion with [Result::expect](https://doc.rust-lang.org/std/result/enum.Result.html#method.expect)!

wiktor-k: Yep, sorry I should've worded it better or just used code suggestion :sweat_smile:

wiktor-k (approved):
LGTM 👍 thanks!

Superhepper (approved):

ionut-arm (approved):

ionut-arm:
Given the approvals I'll merge and fix any new CI issues afterwards.

PR #662, opened 2026-07-22
workspace: keep MSRV at 1.85

baloo:
Context:
`ignore 0.4.30` started using let-chains which were stabilized in 1.88 while retaining what was effectively a declared MSRV of 1.85.

This introduces a direct dependency on the ignore crate to ensure we don't break MSRV 1.85.

This is a workaround an [incorrect release of ignore].

[incorrect release of ignore]: <https://github.com/BurntSushi/ripgrep/issues/3473>


baloo:
cc @hyperfinitism

Superhepper:
Because it has not been released yet. I see no problem with bumping MSRV further if necessary.

baloo:
Closing in favor of https://github.com/parallaxsecond/rust-tss-esapi/pull/663

PR #663, opened 2026-07-22
bump MSRV to 1.88

baloo:
One of our indirect dependency (ignore) now has a minimum MSRV of 1.88.

To smooth the transition and avoid workaround, we'll just bump MSRV to 1.88 as well.


ionut-arm (dismissed):
This is needed to get the CI green again. It doesn't fully get there, but I say we merge it regardless.

Superhepper:
The CI is green again so just rebase this so we can have it merged.

ionut-arm:
> The CI is green again so just rebase this so we can have it merged.

I'm tempted to just merge as it is, and if something breaks we can fix afterwards, I think we've fixed that had failed here.

PR #666, opened 2026-08-17
Pin tpm2-tss in Rawhide

ionut-arm:
Pins the tpm2-tss libraries for Fedora rawhide. Removes the no_mangle_generic_items clippy rule.

Together with #663 this should turn the CI green.

### DIFF
--- a/tss-esapi/tests/Dockerfile-fedora-rawhide
+++ b/tss-esapi/tests/Dockerfile-fedora-rawhide
@@ -1,8 +1,37 @@
 FROM fedora:rawhide
 
+# TODO: revert when mainline is fixed
+# RUN dnf install -y \
+# 	tpm2-tss-devel tpm2-tools \
+# 	swtpm swtpm-tools swtpm-selinux\
+# 	rust clippy cargo \
+# 	llvm llvm-devel clang pkg-config \
+# 	dbus-daemon rust-gobject-sys-devel
+
 RUN dnf install -y \
-	tpm2-tss-devel tpm2-tools \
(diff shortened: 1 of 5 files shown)


tss-esapi/tests/Dockerfile-fedora-rawhide
```diff
@@ -1,8 +1,37 @@
 FROM fedora:rawhide
 
+# TODO: revert when mainline is fixed
```
Superhepper, line 3: Excellent with the comment. I was just thinking about it and how we should make it clear it is temporary. 

Maybe we could even make it as flag to the CI script. So that we set the flag in CI when we build the image `USE_TEMPORARILY_TPM2_TSS_LOCKED_VERSION` or something liken that. Not really sure if its a good idea or if it just will hide things that is then later forgotten.

Superhepper:
Just need to fix the clippy lints and this is done. The `assert_fs` change.

tss-esapi/tests/Dockerfile-fedora-rawhide
ionut-arm: I think maybe the best way is to create an issue here for tpm2-tss v5 support, and mention this there. Otherwise we're likely to forget anyway.

ionut-arm:
> Just need to fix the clippy lints and this is done. The `assert_fs` change.

I've cherry-picked the commit in #664 , cc @hyperfinitism

Superhepper (approved):

wiktor-k (approved):
LGTM 👍 thanks!

PR #667, opened 2026-08-20
Add Context::policy_nv (TPM2_PolicyNV)

joaopinto:
`TPM2_PolicyNV` is currently unimplemented — the source carries `// Missing function: PolicyNV`
and only `policy_nv_written` is available. This adds it.

`TPM2_PolicyNV` gates a policy on the contents of an NV index, with the comparison performed by
the TPM. That is why a software check is not a substitute: code can be given a false value for
an NV counter, the TPM cannot. The concrete use is anti-rollback — sealing a secret under a
policy asserting *counter <= version*, so an old sealed blob stops opening once the counter
advances.

The change is additive and follows existing patterns:

- `constants::ComparisonOperation` — the `TPM2_EO_*` comparison operators as an enum, with
  `From<ComparisonOperation> for TPM2_EO` and `TryFrom<TPM2_EO>`, matching how other TPM
  constants are wrapped in this crate. `TPM2_EO` was previously only a bare `UINT16`.
- `Context::policy_nv` — modelled on `Context::policy_nv_written` in the same file: same
  auth-handle handling as `nv_read` (`NvAuth` -> `AuthHandle`), same `ReturnCode::ensure_success`
  error handling, same optional-session pattern.

A test is included, following `test_policy_nv_written`. It defines an NV index first, because
`TPM2_PolicyNV` reads the index's Name to compute the policy digest and so requires the index to
exist even in a trial session, and it undefines the index before asserting so a failure leaves
no state behind.

Tested against `swtpm` and against an Infineon SLB 9672 on a Raspberry Pi 5.


tss-esapi/src/constants/comparison_operation.rs
```diff
@@ -0,0 +1,70 @@
...
+
+use crate::{Error, Result, WrapperErrorKind, tss2_esys::TPM2_EO};
+use log::error;
+use std::convert::TryFrom;
+
+/// Enum representing the comparison operations (TPM2_EO) of the TPM 2.0
+/// specification, used to compare an NV Index's contents with an operand.
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
+#[repr(u16)]
+pub enum ComparisonOperation {
```
ionut-arm, line 14: Please note that this type is already present! You can find it as `ArithmeticComparison` in `tss-esapi/src/interface_types/arithmetic_comparison.rs`. It's currently not used in any operations, though

ionut-arm (commented):
Thank you for the patch! This seems to implement an op that's covered by #668 as well. Thus, we'll likely merge it under that PR.

Did you have a need / interest in TPM2_PolicyNV, or are you looking to contribute more generally?

joaopinto:
> Thank you for the patch! This seems to implement an op that's covered by #668 as well. Thus, we'll likely merge it under that PR.
> 
> Did you have a need / interest in TPM2_PolicyNV, or are you looking to contribute more generally?

Thanks for the review. I had a concrete need for TPM2_PolicyNV while working with TPM-backed anti-rollback/sealed state on a Raspberry Pi 5 with an Infineon SLB 9672. I found the wrapper was missing, implemented it for my use case, and thought it made sense to contribute it upstream.

I hadn't noticed the existing ArithmeticComparison type. Thanks for pointing that out. I also see that #668 now covers PolicyNV, so I'm happy for that implementation to take precedence.

I'm not specifically looking for things to contribute to, but if I run into other gaps while using the library, I'm happy to contribute fixes where useful.

ionut-arm:
The effort is appreciated! Do let us know if you encounter any issues as well.

